PR #283, opened 2016-03-02
Factor out PackageCacheSynchronizer from UniversePackageCache

### LINKED ISSUE #275: Split up UniversePackageCache
While fixing #270, I discovered that `UniversePackageCache` is quite difficult to write unit tests for due to its tight coupling with the file system. To make writing tests easier, it should be split up into at least two traits/classes:

1. A trait (and corresponding implementation class) to manage the file system state. It would own the package cache directory and take care of synchronizing repository state. It would define public methods that provide access to the parsed and validated package files. This trait could then be mocked in unit tests.
1. A class that would implement the current public methods of `UniversePackageCache`, using the above trait to obtain package file data.

cruhland:
The new class is responsible for keeping the repository files
synchronized on the local file system. UniversePackageCache is now
only responsible for performing repository operations within a
static directory.

Fixes #275.


cruhland:
Closing this, it's been sitting around too long and has gotten stale. I might try to split it up into smaller PRs at some point.

PR #310, opened 2016-04-25
add cli property to universe resource model

tamarrow:
WIP for universe changes for specifying binary clis

### DIFF
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/Resource.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/Resource.scala
@@ -3,7 +3,13 @@ package com.mesosphere.universe.v3.model
 /**
   * Conforms to: https://github.com/mesosphere/universe/blob/version-3.x/repo/meta/schema/v3-repo-schema.json#/definitions/resource
   */
-case class Resource(
+case class V2Resource(
   assets: Option[Assets] = None,
   images: Option[Images] = None
+)
+
+case class V3Resource(
+  assets: Option[Assets] = None,
+  images: Option[Images] = None,
+  cli: Option[Cli] = None
 )
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
@@ -13,6 +13,7 @@ import com.twitter.util.Future
 import io.circe.parse._
 
 import scala.io.Source
+import scala.language.implicitConversions
 import java.io.InputStream
 import java.util.Properties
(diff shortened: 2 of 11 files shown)


BenWhitehead:
I forgot two things.

1. We need to create the `V2Resource` and `V3Resource` objects for each of the respective `V*Package` objects.
2. Please also copy the updated universe schema into `cosmos-model/src/resources/com/mesosphere/universe/v3/repo/meta/schema` so we have it in the code.

tamarrow:
@BenWhitehead updated

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -117,6 +118,12 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
         }
     }
   }
+  implicit def v2ResourceToV3Resource(r2: Option[V2Resource]): Option[V3Resource] = {
+    r2 match {
+      case None => None
+      case Some(r) => Some(V3Resource(r.assets, r.images))
+    }
+  }
```
tamarrow: this is temporary, @cruhland will fix with his changes.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/Resource.scala
```diff
@@ -3,7 +3,13 @@ package com.mesosphere.universe.v3.model
 /**
   * Conforms to: https://github.com/mesosphere/universe/blob/version-3.x/repo/meta/schema/v3-repo-schema.json#/definitions/resource
   */
-case class Resource(
+case class V2Resource(
   assets: Option[Assets] = None,
   images: Option[Images] = None
 )
+
+case class V3Resource(
```
cruhland: The `v3` model doesn't use the `V2Resource`, does it? If not, then these should be declared in different packages:

- `com.mesosphere.universe.v2.model.Resource`
- `com.mesosphere.universe.v3.model.Resource`

cruhland: Ah, I see. `V2Resource` is used in `V2Package`. Never mind!

cruhland: Hmm, actually, based on the other fields in `V2Package`, maybe we can have just one `Resource` type after all. It just means that whenever `Resource` appears in a `V2Package`, its `cli` field will be `None`.

cruhland: Never mind again. Discussed with Tamar, the current changes are correct.

cruhland:
LGTM :+1:

PR #315, opened 2016-04-30
Authorization header for integration tests

cruhland:
With the new security updates to adminrouter in DC/OS 1.7, running integration tests from outside the cluster has become more difficult. This change allows a pre-created authorization token to be given to Cosmos via an environment variable (for security). Cosmos will then use that token in the `Authorization` header of all requests to adminrouter.

I've moved the integration test build to the open-source section of TeamCity, and verified that it works with this PR.


jsancio:
LGTM. I think we should try to figure out a way to avoid package private in the code. I know that we need them for testing but that also means that any code in that package can access it.

cruhland:
[This StackOverflow question](http://stackoverflow.com/questions/21369093/how-can-a-private-class-method-be-tested-in-scala) has some discussion around unit tests for private methods. This is one of those areas where there are many competing opinions for how to do it.

I'm not that satisfied with the proposed answers. Maybe the one that comes closest to what I want is [using `PrivateMethodTester`](http://stackoverflow.com/a/24375762) from ScalaTest. In an ideal world you could access private methods from tests without exposing them to anything else.

PR #324, opened 2016-05-03
UpstreamUniverse trait and related data types

### LINKED ISSUE #316: Define universe data types
We need to define data types (case classes) for PackageId, Package, ResourceId and DockerImage.

### LINKED ISSUE #323: Implement proxy/stub for universe
We need a Scala trait and implementation that abstract the calls to Universe. This proxy or stub should defined all of the operations supported by Universe.

- [x] Refactor code to abstract how universe gets fetched
- [ ] Refactor existing code to translate old universe to common data model
- [ ] Implement ability to fetch from both old and new universe

cruhland:
Starting point for the Cosmos-Universe interface

Fixes #316.
Fixes #323.

If there's any egregious problems with this design, let's catch them now. Minor things like type and field names are also nice to discuss, but they will likely change in later PRs as we implement them and get a feel for what is most effective.

### DIFF
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/PackageId.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/PackageId.scala
@@ -1,0 +1,3 @@
+package com.mesosphere.universe
+
+final case class PackageId(packageName: String, releaseVersion: ReleaseVersion)
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/UniversePackage.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/UniversePackage.scala
@@ -1,0 +1,13 @@
+package com.mesosphere.universe
+
+import com.twitter.io.Buf.ByteBuffer
+import io.circe.JsonObject
+
+final case class UniversePackage(
+  packageId: PackageId,
+  packageJson: PackageDetails,
+  marathonJsonMustache: ByteBuffer,
+  commandJson: Option[Command],
+  configJson: Option[JsonObject],
+  resources: Seq[ResourceId]
+)
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/UpstreamUniverse.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/UpstreamUniverse.scala
@@ -1,0 +1,17 @@
+package com.mesosphere.universe
+
+import com.twitter.util.Future
+
+/** The interface to our parent Universe repository. */
+trait UpstreamUniverse {
+
+  /** Obtains all upstream package IDs that fall within the provided scope. */
+  def observablePackages(scope: PackageScope): Future[List[PackageId]]
+
+  /** Provides detailed information for the given package. */
+  def describePackageForSync(packageId: PackageId): Future[Option[UniversePackage]]
+
+  /** Translates the resource name of a Docker image into its official Docker image name */
+  def resolveDockerImage(imageResourceName: String): Future[Option[DockerImageName]]
+
+}
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/package.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/package.scala
@@ -1,0 +1,10 @@
+package com.mesosphere.universe
+
+object `package` {
+
+  type DockerImageName = String
+
+  /** Sequence of JSON object field names */
+  type ResourceId = Seq[String]
+
+}
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/PackageScope.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/PackageScope.scala
@@ -1,0 +1,4 @@
+package com.mesosphere.universe
+
+/** Parameters that restrict our view of the full Universe of packages. */
+final case class PackageScope(packagingVersion: PackagingVersion)


cruhland:
@BenWhitehead you'll be using these types in your Universe endpoint implementations, so let's make sure we both like them.

cosmos-model/src/main/scala/com/mesosphere/universe/PackageId.scala
```diff
@@ -0,0 +1,3 @@
+package com.mesosphere.universe
+
+final case class PackageId(packageName: String, releaseVersion: ReleaseVersion)
```
cruhland: As @jsancio and I discussed, I've put the `ReleaseVersion` in the package ID, not the `PackageDetailsVersion`. This is a simple way to ensure the IDs are ordered from least to most recent. But we might want to discuss alternatives.

BenWhitehead: what do we need ordering for? "newness" doesn't necessarily mean anything useful to the user.  I'd prefer we use the `PackageDetailsVersion`

cosmos-model/src/main/scala/com/mesosphere/universe/UniversePackage.scala
```diff
@@ -0,0 +1,12 @@
+package com.mesosphere.universe
+
+import io.circe.JsonObject
+
+final case class UniversePackage(
+  packageId: PackageId,
+  packageJson: PackageDetails,
+  marathonJsonMustache: Array[Byte],
```
BenWhitehead: I'm pretty sure we already have an encoder/decoder for `ByteBuffer` that make dealing with this easier (and `ByteBuffer` is nicer than raw `Array[Byte]`)

All of the objects that we're making here for new things should be namespaced to the "version" of the thing they apply to, we need to make sure that can express multiple versions of things if we're going to support multiple versions of repos with the same version of cosmos.

cosmos-model/src/main/scala/com/mesosphere/universe/package.scala
```diff
@@ -0,0 +1,10 @@
+package com.mesosphere.universe
+
+object `package` {
+
+  type DockerImageName = String
+
+  /** Sequence of JSON object field names */
+  type ResourceId = Seq[String]
```
BenWhitehead: Why are these type aliases instead of value objects?  `ResourceId` I suspect will have strict validation need for them and any old `Seq` of `String` probably won't be valid.

cosmos-model/src/main/scala/com/mesosphere/universe/UpstreamUniverse.scala
```diff
@@ -0,0 +1,17 @@
...
+trait UpstreamUniverse {
+
+  /** Obtains all upstream package IDs that fall within the provided scope. */
+  def observablePackages(scope: PackageScope): Future[List[PackageId]]
+
+  /** Provides detailed information for the given package. */
+  def describePackageForSync(packageId: PackageId): Future[Option[UniversePackage]]
+
+  /** Translates the resource name of a Docker image into its official Docker image name */
+  def resolveDockerImage(imageResourceName: String): Future[Option[DockerImageName]]
```
BenWhitehead: What's different with this compared to any other resource? Shouldn't we be able to resolve any resource?

cosmos-model/src/main/scala/com/mesosphere/universe/PackageId.scala
cruhland: We have to implement `PackageCollection.getPackageIndex` in terms of these data structures that we get back from Universe. That method returns `UniverseIndexEntry`, which contains a `currentVersion: PackageDetailsVersion` field. So at the very least we need a way to tell which `PackageId` in the set of IDs with the same package name is the "current" one.

cosmos-model/src/main/scala/com/mesosphere/universe/UniversePackage.scala
cruhland: > I'm pretty sure we already have an encoder/decoder for `ByteBuffer` that make dealing with this easier (and `ByteBuffer` is nicer than raw `Array[Byte]`)

Changed to `ByteBuffer`.

> All of the objects that we're making here for new things should be namespaced to the "version" of the thing they apply to

So I should put them in the package `com.mesosphere.universe.X` where `X` is a version string for the "new fancy Universe"?

cosmos-model/src/main/scala/com/mesosphere/universe/package.scala
cruhland: I was considering using much more strongly-typed representations but I wanted to see what others thought first. I can update them now.

cosmos-model/src/main/scala/com/mesosphere/universe/UpstreamUniverse.scala
cruhland: Cosmos doesn't strictly need to resolve other resources since the Mesos fetcher will be doing that. We'd also have to change the return type to be something like `Future[Option[Either[DockerImage, ByteBuffer]]]` and the parameter type to `ResourceId`.

jsancio: Yeah. That return type is not ideal for two reasons there is an `Either` and the `ByteBuffer` (some of those resources are unbounded, e.g. in the GB magnitude).

jsancio:
Not implementing this solution for this release!

PR #334, opened 2016-05-10
Update cosmos to pass authorization headers to downstream services

### LINKED ISSUE #333: Cosmos should pass through Authorization header to Marathon and Mesos

BenWhitehead:
When a request is made to cosmos any authorization header present in the request will be stored in a request session object and added as a header for any requests made to downstream services (marathon, mesos).

Fixes #333


jsancio:
LGTM. :shipit:

PR #336, opened 2016-05-12
update universe schema to have an optional marathon.json.mustache

tamarrow:
We will be supporting universe packages that are cli-only.

### DIFF
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/Binary.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/Binary.scala
@@ -1,0 +1,3 @@
+package com.mesosphere.universe
+
+case class Binary(kind: String, url: String, contentHash: List[HashInfo])
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/PackageFiles.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/PackageFiles.scala
@@ -7,7 +7,7 @@ case class PackageFiles(
   revision: String,
   sourceUri: Uri,
   packageJson: PackageDetails,
-  marathonJsonMustache: String,
+  marathonJsonMustache: Option[String] = None,
   commandJson: Option[Command] = None,
   configJson: Option[JsonObject] = None,
   resourceJson: Option[Resource] = None
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/Platforms.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/Platforms.scala
@@ -1,0 +1,7 @@
+package com.mesosphere.universe
+
+case class Platforms(
+  windows: Option[Architectures],
+  linux: Option[Architectures],
+  darwin: Option[Architectures]
+)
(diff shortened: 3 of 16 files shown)


cosmos-model/src/main/scala/com/mesosphere/universe/Binary.scala
```diff
@@ -0,0 +1,3 @@
+
+package com.mesosphere.universe
```
jsancio: newline formatting doesn't look right.

cosmos-model/src/main/scala/com/mesosphere/universe/PackageFiles.scala
```diff
@@ -1,13 +1,13 @@
-package com.mesosphere.universe
 
+package com.mesosphere.universe
```
jsancio: newline formatting doesn't look right.

cosmos-model/src/main/scala/com/mesosphere/universe/Platforms.scala
```diff
@@ -0,0 +1,7 @@
+package com.mesosphere.universe
+
+case class Platforms(
+  windows: Option[Map[String, Binary]],
+  linux: Option[Map[String, Binary]],
+  darwin: Option[Map[String, Binary]]
```
jsancio: Why are we using a `Map[String, Binary]`? We can instead add `case class Architectures(`x86-64`: Binary)`.

cosmos-model/src/main/scala/com/mesosphere/universe/Binary.scala
```diff
@@ -0,0 +1,3 @@
...
+case class Binary(kind: String, url: String, contentHash: List[HashInfo])
```
jsancio: Let's give this a better name. Maybe `SubcommandPackage`? or `CliPackage`?

jsancio: Also, `kind` is technically an enum. Do we serialize/deserialize enums anywhere else in Cosmos?

tamarrow: Re naming. I wanted to follow the naming from the universe-schema, so this all goes inside "cli". Maybe I can put all these files in a `CliPackage` directory or something to keep them together?

tamarrow: re: enums, not that I'm aware of, but this would allow cosmos to be more resilient if we add more types in the future.

cosmos-model/src/main/scala/com/mesosphere/universe/Platforms.scala
tamarrow: I wanted to make cosmos resilient of us added more options (ie more architectures) and having the cli handle that. But with versioning, making a change later would be fine.

tamarrow:
@jsancio updated

BenWhitehead:
Already merged with v3 universe model work: https://github.com/dcos/cosmos/blob/master/cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala#L74

PR #339, opened 2016-05-18
Support repo Zip files without directory entries

### LINKED ISSUE #250: NoSuchFileException when extracting repo bundles that do not contain directory entries
If a repository is added to Cosmos (via `POST /package/repository/add`) whose zipped bundle file contains only file entries and no directory entries, then when Cosmos attempts to extract the bundle to local disk (e.g. if the next request is `POST /package/search`), it attempts to copy the files into their (nonexistent) parent directories, resulting in a `NoSuchFileException`. Here's an example Zip file layout that triggers the exception:

```
  Length      Date    Time    Name
---------  ---------- -----   ----
     1170  2016-02-24 12:07   stub-universe-kafka/repo/packages/K/kafka/0/resource.json
     1017  2016-02-24 12:07   stub-universe-kafka/repo/packages/K/kafka/0/package.json
     3099  2016-02-24 12:07   stub-universe-kafka/repo/packages/K/kafka/0/marathon.json.mustache
      123  2016-02-24 12:07   stub-universe-kafka/repo/packages/K/kafka/0/command.json
     7200  2016-02-24 12:07   stub-universe-kafka/repo/packages/K/kafka/0/config.json
       27  2016-02-24 12:07   stub-universe-kafka/repo/meta/version.json
       76  2016-02-24 12:07   stub-universe-kafka/scripts/.stub_dir
---------                     -------
    12712                     7 files
```

~~This issue is blocked by #282.~~ No longer blocked: we will infer the parent directories from the file entry paths.

cruhland:
We ensure that each file in the Zip has all of its parent directories
created before copying it to the filesystem.

Fixes #250.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
@@ -176,30 +176,11 @@ object UniversePackageCache {
     universeDir: Path
   ): Path = {
 
-    val entries = Iterator.continually(Option(bundle.getNextEntry()))
-      .takeWhile(_.isDefined)
-      .flatten
-
     val tempDir = Files.createTempDirectory(universeDir, "repository")
     tempDir.toFile.deleteOnExit()
 
     try {
-      // Copy all of the entry to a temporary folder
-      entries.foreach { entry =>
-        val entryPath = Paths.get(entry.getName)
-
-        // Skip the root directory
-        if (entryPath.getNameCount > 1) {
-          val tempEntryPath = tempDir
-            .resolve(entryPath.subpath(1, entryPath.getNameCount))
-
-          if (entry.isDirectory) {
-            Files.createDirectory(tempEntryPath)
-          } else {
-            Files.copy(bundle, tempEntryPath)
-          }
-        }
-      }
+      extractBundleToDirectory(bundle, tempDir)
 
       // Move the symblic to the temp directory to the actual universe directory...
       val bundlePath = universeDir.resolve(base64(universeBundle))
@@ -228,6 +209,22 @@ object UniversePackageCache {
         // Only delete directory on failures because we want to keep it around on success.
         TwitterFiles.delete(tempDir.toFile)
         throw e
+    }
+  }
+
+  private[this] def extractBundleToDirectory(bundle: ZipInputStream, directory: Path): Unit = {
+    // getNextEntry() returns null when there are no more entries
+    Iterator.continually(Option(bundle.getNextEntry()))
+      .takeWhile(_.isDefined)
+      .flatten
+      .filter(!_.isDirectory)
+      .map(entry => Paths.get(entry.getName))
+      // Skip the root directory
+      .filter(entryPath => entryPath.getNameCount > 1)
+      .foreach { entryPath =>
+        val tempEntryPath = directory.resolve(entryPath.subpath(1, entryPath.getNameCount))
+        Option(tempEntryPath.getParent).foreach(Files.createDirectories(_))
+        Files.copy(bundle, tempEntryPath)
     }
   }
 
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/ZooKeeperStorage.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/ZooKeeperStorage.scala
@@ -69,7 +69,7 @@ private[cosmos] final class ZooKeeperStorage(
           Future(decodeData(bytes))
         case None =>
           readCacheStats.counter("miss").incr
-          create(DefaultRepos)
+          read()
       }
     }
   }
--- a/cosmos-server/src/test/scala/com/mesosphere/cosmos/UniversePackageCacheSpec.scala
+++ b/cosmos-server/src/test/scala/com/mesosphere/cosmos/UniversePackageCacheSpec.scala
@@ -1,8 +1,9 @@
 package com.mesosphere.cosmos
 
-import java.io.{ByteArrayInputStream, IOException, InputStream}
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, IOException, InputStream}
 import java.net.{MalformedURLException, URL}
 import java.nio.file.{Files, Path, Paths}
+import java.util.zip.{ZipEntry, ZipInputStream, ZipOutputStream}
 
 import com.mesosphere.cosmos.circe.Encoders._
 import com.mesosphere.cosmos.model.{PackageRepository, SearchResult}
@@ -12,9 +13,9 @@ import com.netaporter.uri.Uri
 import com.twitter.io.{Charsets, StreamIO}
 import com.twitter.util.{Await, Throw}
 import io.circe.syntax._
-import org.scalatest.FreeSpec
+import org.scalatest.{FreeSpec, PrivateMethodTester}
 
-final class UniversePackageCacheSpec extends FreeSpec {
+final class UniversePackageCacheSpec extends FreeSpec with PrivateMethodTester {
 
   import UniversePackageCacheSpec._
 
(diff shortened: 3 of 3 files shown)


cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -26,12 +27,14 @@ final class PackageDescribeSpec extends FreeSpec with TableDrivenPropertyChecks
   "The package describe endpoint" - {
     "returns an error when package w/ version not found" in {
       forAll (PackageDummyVersionsTable) { (packageName, packageVersion) =>
-        describeAndAssertError(
-          packageName = packageName,
-          status = Status.BadRequest,
-          expectedMessage = s"Version [$packageVersion] of package [$packageName] not found",
-          version = Some(packageVersion)
-        )
+        eventually { // Ignore failures due to concurrent ZK modifications
```
cruhland: This fixes a test failure I was getting locally. If it's worth the time investment I can try to find the root cause.

cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
```diff
@@ -231,6 +212,23 @@ object UniversePackageCache {
     }
   }
 
+  private[this] def extractBundleToDirectory(bundle: ZipInputStream, directory: Path): Unit = {
+    Iterator.continually(Option(bundle.getNextEntry()))
+      .takeWhile(_.isDefined)
+      .flatten
+      .filter(!_.isDirectory)
```
cruhland: This line is new.

cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
```diff
@@ -231,6 +212,23 @@ object UniversePackageCache {
...
+      .foreach { entry =>
+        val entryPath = Paths.get(entry.getName)
+
+        // Skip the root directory
+        if (entryPath.getNameCount > 1) {
+          val tempEntryPath = directory.resolve(entryPath.subpath(1, entryPath.getNameCount))
+          Option(tempEntryPath.getParent).foreach(Files.createDirectories(_))
```
cruhland: This line is new.

cosmos-server/src/test/scala/com/mesosphere/cosmos/UniversePackageCacheSpec.scala
```diff
@@ -133,6 +134,54 @@ final class UniversePackageCacheSpec extends FreeSpec {
...
 
+  "Issue #250: create parent directories for unzipped files" - {
+    "extractBundleToDirectory()" in {
+      val tempDir = Files.createTempDirectory(getClass.getSimpleName)
+
+      val extractBundleToDirectory = PrivateMethod[Unit]('extractBundleToDirectory)
+      val bundle = createBundle
+
+      // No exceptions will be thrown if this is successful
+      UniversePackageCache invokePrivate extractBundleToDirectory(bundle, tempDir)
```
cruhland: @jsancio what do you think of this for testing private methods? It uses reflection, but I don't think that's such a big deal in tests, since the test will fail with an exception if there's a problem calling the method.

cosmos-server/src/test/scala/com/mesosphere/cosmos/UniversePackageCacheSpec.scala
```diff
@@ -133,6 +134,54 @@ final class UniversePackageCacheSpec extends FreeSpec {
...
+    }
+
+    "extractBundle()" in {
+      val universeDir = Files.createTempDirectory(getClass.getSimpleName)
+
+      val extractBundle = PrivateMethod[Path]('extractBundle)
+      val bundle = createBundle
+
+      // No exceptions will be thrown if this is successful
+      UniversePackageCache invokePrivate extractBundle(bundle, Uri.parse("repo/uri"), universeDir)
```
BenWhitehead: I think it would be good to add some assertions for the file structure we expect on disk and that the files have the expected contents. This ensures that if the contents of the method are updated with an overly aggressive filter the test doesn't still pass. This will also ensure the filtering applied to the zip entries is doing what we want/expect.

cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
```diff
@@ -231,6 +212,21 @@ object UniversePackageCache {
     }
   }
 
+  private[this] def extractBundleToDirectory(bundle: ZipInputStream, directory: Path): Unit = {
+    Iterator.continually(Option(bundle.getNextEntry()))
```
BenWhitehead: Can you add a comment explaining how/why `getNextEntry` will result in `null`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
```diff
@@ -231,6 +212,21 @@ object UniversePackageCache {
...
+      .takeWhile(_.isDefined)
+      .flatten
+      .filter(!_.isDirectory)
+      .map(entry => Paths.get(entry.getName))
+      // Skip the root directory
```
BenWhitehead: Why do we skip the root directory?

cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
cruhland: Done.

cosmos-server/src/test/scala/com/mesosphere/cosmos/UniversePackageCacheSpec.scala
cruhland: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
cruhland: I believe this is because we require a top-level directory in the zip, but that directory is redundant when the files are extracted to disk, since we've already created an empty, uniquely-named destination directory to extract to.

cc @jsancio

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
cruhland: Looking into this now...

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -15,9 +15,10 @@ import io.circe.Json
 import io.circe.parse._
 import io.circe.syntax._
 import org.scalatest.FreeSpec
+import org.scalatest.concurrent.Eventually
 import org.scalatest.prop.TableDrivenPropertyChecks
 
-final class PackageDescribeSpec extends FreeSpec with TableDrivenPropertyChecks {
+final class PackageDescribeSpec extends FreeSpec with TableDrivenPropertyChecks with Eventually {
```
cruhland: Oops, will remove this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/ZooKeeperStorage.scala
```diff
@@ -69,7 +69,7 @@ private[cosmos] final class ZooKeeperStorage(
           Future(decodeData(bytes))
         case None =>
           readCacheStats.counter("miss").incr
-          create(DefaultRepos)
+          read()
```
cruhland: Here's the fix for the test failure I was getting in `PackageDescribeSpec`. The Curator framework's `NodeCache` will necessarily be slightly out of sync with the state of ZooKeeper itself. So when a cache read fails because the node doesn't exist, we should first try to read directly from ZK, in case the cache is stale. We don't want to immediately try to create the node, because that results in an error if the node exists. Plus, `read()` will call `create()` for us if the node actually does not exist.

BenWhitehead:
LGTM

cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
jsancio: That is correct. We cannot rely on the zip's root directory to be unique so Cosmos needs to do it.

PR #341, opened 2016-05-20
Include all CosmosError subclass arguments in "data" field of response

### LINKED ISSUE #312: improve error structuring with data fields
For all contextual information that we provide in error messages we should also provide in data fields so that the UI can change the message to whatever they see fit.

cc @mlunoe

cruhland:
Provides a default implementation of CosmosError.getData which uses
generated Circe Encoders to create a JSON representation of any
instance. Subclasses can still override the default if needed.

Fixes #312.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
@@ -18,12 +18,11 @@ sealed abstract class CosmosError(causedBy: Throwable = null /*java compatibilit
 
   def status: Status = Status.BadRequest
 
-  def getData: JsonObject = {
+  def getData: Option[JsonObject] = {
     this.asJson
       .asObject
       .flatMap(_.values.headOption)
       .flatMap(_.asObject)
-      .getOrElse(JsonObject.empty)
   }
 
   def getHeaders: Map[String, String] = Map.empty
@@ -38,8 +37,13 @@ case class PackageFileSchemaMismatch(fileName: String) extends CosmosError
 case class PackageAlreadyInstalled() extends CosmosError {
   override val status = Status.Conflict
 }
-case class MarathonBadResponse(marathonError: MarathonError) extends CosmosError
-
+case class MarathonBadResponse(marathonError: MarathonError) extends CosmosError {
+  val details = marathonError.details match {
+    case Some(details) => Some(JsonObject.singleton("errors", details.asJson))
+    case None => None
+  }
+  override def getData: Option[JsonObject] = details
+}
 case class MarathonGenericError(marathonStatus: Status) extends CosmosError {
   override val status = Status.InternalServerError
 }
@@ -70,7 +74,9 @@ case class PackageNotInstalled(packageName: String) extends CosmosError
 
 case class NelErrors(errs: NonEmptyList[CosmosError]) extends CosmosError //TODO: Cleanup
 
-case class JsonSchemaMismatch(errors: Iterable[Json]) extends CosmosError
+case class JsonSchemaMismatch(errors: Iterable[Json]) extends CosmosError {
+  override def getData: Option[JsonObject] = Some(JsonObject.singleton("errors", errors.asJson))
+}
 
 case class FileUploadError(message: String) extends CosmosError { override val status = Status.NotImplemented }
 
@@ -81,10 +87,12 @@ case class ServiceUnavailable(
   causedBy: Throwable
 ) extends CosmosError(causedBy) {
   override val status = Status.ServiceUnavailable
+  override val getData = Some(JsonObject.singleton("serviceName", serviceName.asJson))
 }
 case class Unauthorized(serviceName: String, realm: Option[String]) extends CosmosError with NoStackTrace {
   override val getMessage: String = s"Unable to complete request due to Unauthorized response from service [$serviceName]"
   override val status: Status = Status.Unauthorized
+  override val getData: Option[JsonObject] = Some(JsonObject.singleton("serviceName", serviceName.asJson))
   override val getHeaders: Map[String, String] = realm match {
     case Some(r) =>
       Map("WWW-Authenticate" -> r)
@@ -95,6 +103,7 @@ case class Unauthorized(serviceName: String, realm: Option[String]) extends Cosm
 case class Forbidden(serviceName: String) extends CosmosError with NoStackTrace {
   override val getMessage: String = s"Unable to complete request due to Forbidden response from service [$serviceName]"
   override val status: Status = Status.Forbidden
+  override val getData: Option[JsonObject] = Some(JsonObject.singleton("serviceName", serviceName.asJson))
 }
 
 case class IncompleteUninstall(packageName: String, causedBy: Throwable) extends CosmosError(causedBy)
@@ -105,13 +114,13 @@ case class ConcurrentAccess(causedBy: Throwable) extends CosmosError(causedBy)
 final case class RepoNameOrUriMissing() extends CosmosError
 
 case class RepositoryAlreadyPresent(nameOrUri: Ior[String, Uri]) extends CosmosError {
-  override def getData: JsonObject = {
+  override def getData: Option[JsonObject] = {
     val jsonMap = nameOrUri match {
       case Ior.Both(n, u) => Map("name" -> n.asJson, "uri" -> u.asJson)
       case Ior.Left(n) => Map("name" -> n.asJson)
       case Ior.Right(u) => Map("uri" -> u.asJson)
     }
-    JsonObject.fromMap(jsonMap)
+    Some(JsonObject.fromMap(jsonMap))
   }
 }
 case class RepositoryAddIndexOutOfBounds(attempted: Int, max: Int) extends CosmosError
@@ -122,20 +131,28 @@ case class UnsupportedRepositoryUri(uri: Uri) extends CosmosError
 case class RepositoryUriSyntax(
   repository: PackageRepository,
   causedBy: Throwable
-) extends CosmosError(causedBy)
+) extends CosmosError(causedBy) {
+  override def getData: Option[JsonObject] = {
+    Some(JsonObject.singleton("cause", causedBy.getMessage.asJson))
+  }
+}
 
 case class RepositoryUriConnection(
   repository: PackageRepository,
   causedBy: Throwable
-) extends CosmosError(causedBy)
+) extends CosmosError(causedBy) {
+  override def getData: Option[JsonObject] = {
+    Some(JsonObject.singleton("cause", causedBy.getMessage.asJson))
+  }
+}
 
 case class RepositoryNotPresent(nameOrUri: Ior[String, Uri]) extends CosmosError {
-  override def getData: JsonObject = {
+  override def getData: Option[JsonObject] = {
     val jsonMap = nameOrUri match {
       case Ior.Both(n, u) => Map("name" -> n.asJson, "uri" -> u.asJson)
       case Ior.Left(n) => Map("name" -> n.asJson)
       case Ior.Right(u) => Map("uri" -> u.asJson)
     }
-    JsonObject.fromMap(jsonMap)
+    Some(JsonObject.fromMap(jsonMap))
   }
 }
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
@@ -19,6 +19,16 @@ sealed abstract class CosmosError(causedBy: Throwable = null /*java compatibilit
   def status: Status = Status.BadRequest
 
   def getData: Option[JsonObject] = {
+    /* Circe encodes sum types like CosmosError into JSON as follows:
+     * {
+     *   "PackageNotFound": {
+     *     "packageName": "cassandra"
+     *   }
+     * }
+     *
+     * For this method, we just want the field values, so the code below extracts the nested object
+     * from the JSON that Circe generates.
+     */
     this.asJson
       .asObject
       .flatMap(_.values.headOption)
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
@@ -18,7 +18,13 @@ sealed abstract class CosmosError(causedBy: Throwable = null /*java compatibilit
 
   def status: Status = Status.BadRequest
 
-  def getData: Option[JsonObject] = None
+  def getData: JsonObject = {
+    this.asJson
+      .asObject
+      .flatMap(_.values.headOption)
+      .flatMap(_.asObject)
+      .getOrElse(JsonObject.empty)
+  }
 
   def getHeaders: Map[String, String] = Map.empty
 }
@@ -32,13 +38,8 @@ case class PackageFileSchemaMismatch(fileName: String) extends CosmosError
 case class PackageAlreadyInstalled() extends CosmosError {
   override val status = Status.Conflict
 }
-case class MarathonBadResponse(marathonError: MarathonError) extends CosmosError {
-  val details = marathonError.details match {
-    case Some(details) => Some(JsonObject.singleton("errors", details.asJson))
-    case None => None
-  }
-  override def getData: Option[JsonObject] = details
-}
+case class MarathonBadResponse(marathonError: MarathonError) extends CosmosError
+
 case class MarathonGenericError(marathonStatus: Status) extends CosmosError {
   override val status = Status.InternalServerError
 }
@@ -69,9 +70,7 @@ case class PackageNotInstalled(packageName: String) extends CosmosError
 
 case class NelErrors(errs: NonEmptyList[CosmosError]) extends CosmosError //TODO: Cleanup
 
-case class JsonSchemaMismatch(errors: Iterable[Json]) extends CosmosError {
-  override def getData: Option[JsonObject] = Some(JsonObject.singleton("errors", errors.asJson))
-}
+case class JsonSchemaMismatch(errors: Iterable[Json]) extends CosmosError
 
 case class FileUploadError(message: String) extends CosmosError { override val status = Status.NotImplemented }
 
@@ -82,12 +81,10 @@ case class ServiceUnavailable(
   causedBy: Throwable
 ) extends CosmosError(causedBy) {
   override val status = Status.ServiceUnavailable
-  override val getData = Some(JsonObject.singleton("serviceName", serviceName.asJson))
 }
 case class Unauthorized(serviceName: String, realm: Option[String]) extends CosmosError with NoStackTrace {
   override val getMessage: String = s"Unable to complete request due to Unauthorized response from service [$serviceName]"
   override val status: Status = Status.Unauthorized
-  override val getData: Option[JsonObject] = Some(JsonObject.singleton("serviceName", serviceName.asJson))
   override val getHeaders: Map[String, String] = realm match {
     case Some(r) =>
       Map("WWW-Authenticate" -> r)
@@ -98,7 +95,6 @@ case class Unauthorized(serviceName: String, realm: Option[String]) extends Cosm
 case class Forbidden(serviceName: String) extends CosmosError with NoStackTrace {
   override val getMessage: String = s"Unable to complete request due to Forbidden response from service [$serviceName]"
   override val status: Status = Status.Forbidden
-  override val getData: Option[JsonObject] = Some(JsonObject.singleton("serviceName", serviceName.asJson))
 }
 
 case class IncompleteUninstall(packageName: String, causedBy: Throwable) extends CosmosError(causedBy)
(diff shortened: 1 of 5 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
```diff
@@ -18,7 +18,13 @@ sealed abstract class CosmosError(causedBy: Throwable = null /*java compatibilit
...
   def status: Status = Status.BadRequest
 
-  def getData: Option[JsonObject] = None
+  def getData: JsonObject = {
+    this.asJson
+      .asObject
+      .flatMap(_.values.headOption)
+      .flatMap(_.asObject)
+      .getOrElse(JsonObject.empty)
+  }
```
cruhland: Here is the generic implementation. Circe by default will generate JSON for a subclass in a sealed hierarchy as follows:

```json
{
  "SubclassName": {
    "subclassField1": "field value",
    "subclassField2": "another value"
  }
}
```

This implementation simply extracts the inner object (with the fields) from the generated JSON.

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
```diff
@@ -32,13 +38,8 @@ case class PackageFileSchemaMismatch(fileName: String) extends CosmosError
 case class PackageAlreadyInstalled() extends CosmosError {
   override val status = Status.Conflict
 }
-case class MarathonBadResponse(marathonError: MarathonError) extends CosmosError {
-  val details = marathonError.details match {
-    case Some(details) => Some(JsonObject.singleton("errors", details.asJson))
-    case None => None
-  }
-  override def getData: Option[JsonObject] = details
-}
```
cruhland: Deleted this custom `getData` because the generic version should be good enough. I've done that for most `getData` implementations in this file; please comment if you think some should stay.

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
```diff
@@ -108,7 +104,16 @@ case class ConcurrentAccess(causedBy: Throwable) extends CosmosError(causedBy)
...
+case class RepositoryAlreadyPresent(nameOrUri: Ior[String, Uri]) extends CosmosError {
+  override def getData: JsonObject = {
+    val jsonMap = nameOrUri match {
+      case Ior.Both(n, u) => Map("name" -> n.asJson, "uri" -> u.asJson)
+      case Ior.Left(n) => Map("name" -> n.asJson)
+      case Ior.Right(u) => Map("uri" -> u.asJson)
+    }
+    JsonObject.fromMap(jsonMap)
+  }
+}
```
cruhland: Added an override here because the default JSON generated for `Ior` is too generic. This override ensures that the result is readable.

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
cruhland: I also changed the return type from `Option[JsonObject]` to simply `JsonObject`, since in the case of no data we can return an empty object.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -137,6 +140,23 @@ object Encoders {
...
+
+  implicit val encodeMediaType: Encoder[MediaType] = deriveFor[MediaType].encoder
+  implicit val encodeHttpMethod: Encoder[HttpMethod] = Encoder.encodeString.contramap(_.getName)
+
+  implicit def encodeIor[A, B](implicit
+    encodeA: Encoder[A],
+    encodeB: Encoder[B]
+  ): Encoder[Ior[A, B]] = deriveFor[Ior[A, B]].encoder
+
+  implicit val encodeCosmosError: Encoder[CosmosError] = deriveFor[CosmosError].encoder
```
cruhland: Here's the generic encoder for `CosmosError`. The additional encoders above it were needed because some of the subclasses had constructor parameters with those types.

cruhland:
I didn't add any tests specifically for this change, although I did need to update some existing tests as a result. I figured that a generic test may not be that useful; we can write specific tests later when we want to ensure that a particular error type gets encoded in responses with the right data.

cruhland:
@mlunoe this will hopefully make your life easier :smile:

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
tamarrow: Can you post an example of this error before and after this PR?

cruhland: Before:

```json
{
  "type": "MarathonBadResponse",
  "message": "Error message from Marathon",
  "data": {
    "errors": [
      {
        "errorKey1": "errorValue1",
        "errorKey2": "errorValue2"
      },
      {
        "errorKey3": 42
      }
    ]
  }
}
```

After:

```json
{
  "type": "MarathonBadResponse",
  "message": "Error message from Marathon",
  "data": {
    "marathonError": {
      "message": "Error message from Marathon",
      "details": [
        {
          "errorKey1": "errorValue1",
          "errorKey2": "errorValue2"
        },
        {
          "errorKey3": 42
        }
      ]
    }
  }
}
```

BenWhitehead: As long as we can prove that the constructed error document is the same before and after this change it's probably okay, but my suspicion is that they are different. 

My guess is that the new one is:
```json
{
  "data": {
     "marathonError": {
         "message": "asdasd",
         "details": [
           {}
         ]
     }
   }
}
```

Where as I think the old one was:
```json
{
  "data": {
     "errors": [
        {}
     ]
   }
}
```

cruhland: Oh yeah, they are of course different. If it's a problem for clients to update then we can just keep all the overrides around.

BenWhitehead: I think we should keep the old declarations then.

tamarrow: Same, differences will mean brokenness for the CLI for places that look for these data objects :)

cruhland: Ok, will revert.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -137,6 +140,23 @@ object Encoders {
     Base64.getEncoder.encodeToString(ByteBuffers.getBytes(bb)).asJson
   }
 
+  implicit val encodeThrowable: Encoder[Throwable] = Encoder.instance(_.getMessage.asJson)
+  implicit val encodeStatus: Encoder[Status] = Encoder.encodeInt.contramap(_.code)
+  implicit val encodeCirceError: Encoder[CirceError] = encodeThrowable.contramap(identity)
+
+  implicit val encodeMediaTypeSubType: Encoder[MediaTypeSubType] =
```
BenWhitehead: Pease surround with `{` and `}` to be consistent with other multi-line encoders.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -137,6 +140,23 @@ object Encoders {
     Base64.getEncoder.encodeToString(ByteBuffers.getBytes(bb)).asJson
   }
 
+  implicit val encodeThrowable: Encoder[Throwable] = Encoder.instance(_.getMessage.asJson)
```
BenWhitehead: I'd prefer if we didn't allow `Throwable` to be encoded. I suspect it's here because of the `causedBy` on `CosmosError` which in most circumstances I think we don't want to externalize.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
cruhland: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
cruhland: Some of the `CosmosError` subclasses have `Throwable` constructor parameters, so that they can pass them to the `CosmosError` constructor as you said. I'll see if there's a way to make Circe/Shapeless ignore those parameters when deriving encoders.

cruhland: @BenWhitehead I haven't found a way to ignore those fields, but I'm considering removing them entirely. Do we need the `causedBy` member of `CosmosError`? It's not used anywhere as far as I can tell; not even in `exceptionErrorResponse` which converts `CosmosError` to `ErrorResponse` so it can be sent in response bodies.

cruhland: Here's where `causedBy` was added: https://github.com/dcos/cosmos/pull/142/files#diff-37a8f9d27c38e0283c9dbeb390b98c86R12

As far as I can tell from that PR, it was added simply to implement this TODO: https://github.com/dcos/cosmos/pull/142/files#diff-6a1a28ac16d9c7356dba32fc769dfdd2L184

But if we're not surfacing anything about that exception in the response body, do we need it? Does Finagle/Finch print the stack trace in the log or something?

BenWhitehead: Finagle/Finch don't print the stack trace for use at this point, we're attempting to handle all exceptions https://github.com/dcos/cosmos/blob/3fa3533ed530b14223dd369c6b047c803a65782a/cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala#L179-L179

I think this was added for local debugging, so that when an exception happens a debugger can break and traverse the caused by. True we're not exposing it to the user, but it would make local debugging more challenging in the circumstance we end up needing it. Is there some way that we can "hint" to circe (maybe shapeless) to filter out any property that extends `Throwable`?

cruhland: I could write a custom encoder for `CosmosError` based on this SO answer from Travis Brown: http://stackoverflow.com/questions/37011894/circe-instances-for-encoding-decoding-sealed-trait-instances-of-arity-0/37015184#37015184

I'll look into it when I have time this week.

cruhland: Actually, maybe the easier solution would just be to override the `getData` implementation for all error classes that have `Throwable` parameters.

jsancio:
@cruhland @BenWhitehead is this ready for merging?

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
BenWhitehead: I think I'm mainly after figuring out if it's possible to get rid of the need for `Encoder[Throwable]`, if we can get rid of that and circe still compiles I'm good.  I think overriding `getData` is going to be more stable in the long run, but it can see the appeal of having it auto generated.

cruhland:
@jsancio It needs one more update, to get rid of `Encoder[Throwable]`. I'll take care of it now.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -137,6 +143,32 @@ object Encoders {
...
+
+  implicit def dropThrowableFromEncodedObjects[K <: Symbol, H <: Throwable, T <: HList](implicit
+    key: Witness.Aux[K],
+    encodeTail: Lazy[DerivedObjectEncoder[T]]
+  ): DerivedObjectEncoder[FieldType[K, H] :: T] =
+    new DerivedObjectEncoder[FieldType[K, H] :: T] {
+      def encodeObject(a: FieldType[K, H] :: T): JsonObject = {
+        encodeTail.value.encodeObject(a.tail)
+      }
+    }
```
cruhland: Here's the Shapeless incantation that prevents `Throwable` fields from appearing in the encoded data.

cruhland:
@jsancio @BenWhitehead updated.

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
```diff
@@ -18,7 +18,12 @@ sealed abstract class CosmosError(causedBy: Throwable = null /*java compatibilit
 
   def status: Status = Status.BadRequest
 
-  def getData: Option[JsonObject] = None
+  def getData: Option[JsonObject] = {
+    this.asJson
+      .asObject
+      .flatMap(_.values.headOption)
+      .flatMap(_.asObject)
```
jsancio: You can merge the `flatMap`s into one `flatMap`. Can we write a command at to what this is doing? This is not clear to me.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
jsancio: What is this actually doing and why? It looks like you are trying to skip over `Throwable`s in the product type. If this true let's write a comment explaining this.

cosmos-server/src/test/scala/com/mesosphere/cosmos/circe/EncodersDecodersSpec.scala
```diff
@@ -79,6 +79,50 @@ class EncodersDecodersSpec extends FreeSpec {
...
+    "RepositoryUriConnection" in {
+      val repo = PackageRepository("Universe", Uri.parse("universe/repo"))
+      val error = RepositoryUriConnection(repository = repo, causedBy = throwable)
+      assertThrowableDropped(error, "causedBy")
+    }
+
+    def assertThrowableDropped[A <: CosmosError with Product](error: A, throwableFieldName: String): Unit = {
+      val encodedFields = error.getData.getOrElse(JsonObject.empty)
+      assert(!encodedFields.contains(throwableFieldName))
+      assertResult(error.productArity - 1)(encodedFields.size)
```
jsancio: Okay but this assumes that the product type only has one field that is `Throwable`.

cosmos-server/src/test/scala/com/mesosphere/cosmos/circe/EncodersDecodersSpec.scala
```diff
@@ -79,6 +79,50 @@ class EncodersDecodersSpec extends FreeSpec {
     }
   }
 
+  "Throwable fields are dropped from encoded objects" - {
+    val throwable = new RuntimeException("BOOM!")
+
+    "PackageFileMissing" in {
+      assertThrowableDropped(PackageFileMissing(packageName = "kafka", cause = throwable), "cause")
```
jsancio: Probably not an issue with this PR but this type uses `cause` yet the rest use `causedBy`. Is that by accident?

jsancio:
LGTM in general. Can I see a sample output for the before and after? Or just the after if the before is difficult to generate.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -137,6 +143,32 @@ object Encoders {
     Base64.getEncoder.encodeToString(ByteBuffers.getBytes(bb)).asJson
   }
 
+  implicit val encodeStatus: Encoder[Status] = Encoder.encodeInt.contramap(_.code)
+
+  implicit val encodeMediaTypeSubType: Encoder[MediaTypeSubType] = {
+    deriveFor[MediaTypeSubType].encoder
+  }
+
+  implicit val encodeMediaType: Encoder[MediaType] = deriveFor[MediaType].encoder
```
BenWhitehead: Please update encoder to use `MediaType#show`[1] instead of deriving an encoder. This will most likely be more compatible with other things that try to parse the media type.

[1] https://github.com/dcos/cosmos/blob/74e915689e5d155e960b346553a3805351c67b99/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaType.scala#L27-L27

BenWhitehead:
Nice job finding out how to filter out `Throwables`! One small comment from me, having some clarification comments around some of the magic would be good for the next person to come along and try to understand what's going on.

:ship: it after fixing the few small comments.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
cruhland: That's right, if any case classes have `Throwable` fields, this will not include them in the generated JSON. I'll add a comment.

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
cruhland: I don't think I can merge them, unless you meant doing `.flatMap(_.values.headOption.flatMap(_.asObject))`, which is harder to read in my opinion. I'll add a comment to explain this.

cosmos-server/src/test/scala/com/mesosphere/cosmos/circe/EncodersDecodersSpec.scala
cruhland: I'll change the method to accept an arbitrary number of field names.

cosmos-server/src/test/scala/com/mesosphere/cosmos/circe/EncodersDecodersSpec.scala
cruhland: It's just the definition of `PackageFileMissing` being inconsistent with the other error types.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
cruhland: Done.

cruhland:
@jsancio Here's an example that shows a few errors at once. First, construct the error value:

```scala
import cats.data.NonEmptyList
import com.mesosphere.cosmos._
import com.mesosphere.cosmos.circe.Encoders._
import com.mesosphere.cosmos.http._
import io.circe.syntax._

// Ordinary error
val err1 = PackageNotFound("cassandra")

// Error with Throwable field
val err2 = PackageFileMissing("kafka", new java.io.IOException("BOOM!"))

// Error with multiple fields
val err3 = UnsupportedContentType(List(MediaType("application", "json"), MediaType("text", "plain")), Some(MediaType("test", "csv")))

// Combine errors together
val errs = NelErrors(NonEmptyList(err1, err2, err3))

// Render to JSON
(errs: Exception).asJson.spaces2
```

Here's what it looked like in the response **before** this change:
```json
{
  "type" : "NelErrors",
  "message" : "OneAnd(com.mesosphere.cosmos.PackageNotFound,List(com.mesosphere.cosmos.PackageFileMissing: java.io.IOException: BOOM!, com.mesosphere.cosmos.UnsupportedContentType))",
  "data" : null
}
```

And here's what it looks like **after** this change:
```json
{
  "type" : "NelErrors",
  "message" : "OneAnd(com.mesosphere.cosmos.PackageNotFound,List(com.mesosphere.cosmos.PackageFileMissing: java.io.IOException: BOOM!, com.mesosphere.cosmos.UnsupportedContentType))",
  "data" : {
    "errs" : [
      {
        "PackageNotFound" : {
          "packageName" : "cassandra"
        }
      },
      {
        "PackageFileMissing" : {
          "packageName" : "kafka"
        }
      },
      {
        "UnsupportedContentType" : {
          "supported" : [
            "application/json",
            "text/plain"
          ],
          "actual" : "test/csv"
        }
      }
    ]
  }
}
```

The `message` field is still ugly, but now the `data` field is way more useful. Also, @BenWhitehead the `MediaType` is being encoded as you asked.

cruhland:
@jsancio Here are the three nested errors from above, treated separately. These are more representative of actual errors that the CLI or the UI would like to handle:

```json
{
  "type" : "PackageNotFound",
  "message" : "Package [cassandra] not found",
  "data" : {
    "packageName" : "cassandra"
  }
}
```

```json
{
  "type" : "PackageFileMissing",
  "message" : "Package file [kafka] not found",
  "data" : {
    "packageName" : "kafka"
  }
}
```

```json
{
  "type" : "UnsupportedContentType",
  "message" : "Unsupported Content-Type: test/csv Accept: [application/json, text/plain]",
  "data" : {
    "supported" : [
      "application/json",
      "text/plain"
    ],
    "actual" : "test/csv"
  }
}
```

jsancio:
LGTM. :shipit:

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
cruhland: Done.

PR #344, opened 2016-05-25
add zk acl support

spacejam:


cosmos-server/src/main/scala/com/mesosphere/cosmos/Flags.scala
```diff
@@ -27,6 +27,14 @@ object zookeeperUri extends GlobalFlag[ZooKeeperUri](
   "The ZooKeeper connection string"
 )
 
+object zkAclUser extends GlobalFlag[String](
+  help = "User for ZK ACL"
```
jsancio: I would spell out "ZK". E.g. "User for ZooKeeper ACL". Same with the zkAclSecret flag.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -299,3 +305,24 @@ object Cosmos extends FinchServer {
...
+) extends ACLProvider {
+  var acls = new java.util.ArrayList[ACL]()
+
+  userSecret.foreach {
+    case (user, secret) => {
+      val id = new Id("digest", s"$user:$secret")
+      val acl = new ACL(ZooDefs.Perms.ALL, id)
+      acls.add(acl)
+    }
+  }
```
jsancio: Can we do this without side effects? E.g.
```
private final class DigestAclProvider private (acls: java.util.ArrayList[ACL]) ...

object DigestAclProvider {
  def apply(userSecret: Option[(String, String)]): DigestAclProvider = {
    userSecret match {
      case Some((user, secret)) =>
        ...
        new DigestAclProvider(java.util.Arrays.asList(acl))
      case None =>
        new DigestAclProvider(java.util.ArrayList[ACL]())
    }
  }
}
```

The code probably doesn't compile.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Flags.scala
```diff
@@ -27,6 +27,14 @@ object zookeeperUri extends GlobalFlag[ZooKeeperUri](
   "The ZooKeeper connection string"
 )
 
+object zookeeperAclUser extends GlobalFlag[String](
+  help = "User for ZK ACL"
```
jsancio: And the help comment.

jsancio:
Looks good to me @spacejam. I take it that in the integration clusters (CCM) we have AuthN enable but it takes any user and secret.

Thanks for doing this!

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
spacejam: I fought sbt for too long over `java.util.ArrayList[ACL]()` over `object java.util.ArrayList is not a value` and I'm not sure how to appease the compiler to make this more idiomatic.  Any ideas?

spacejam: nvm! I was hitting something else before, this one is clear.

spacejam:
Just need to add recursive acl updating and then a better test for it, then will be good to go.  Should be in-progress still but I think I triggered an automatic re-tag somehow

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
jsancio: Sorry Tyler you are right. You need "new j.u.ArrayList[ACL]()".

The reason why the Scala compiler is complaining is because "Symbol()" gets
rewritten to "Symbol.apply()". Which means that Symbol needs to be a value.
Usually that value gets created when you do "object Symbol". The issue here
is that those are Java classes so they don't have that pattern/sugar
implemented.
On Wed, May 25, 2016 at 08:19 Tyler Neely <notifications@github.com> wrote:

> In cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
> <https://github.com/dcos/cosmos/pull/344#discussion_r64592227>:
>
> > @@ -299,3 +305,24 @@ object Cosmos extends FinchServer {
> >    }
> >
> >  }
> > +
> > +private[cosmos] final case class DigestACLProvider(
> > +  userSecret: Option[(String, String)]
> > +) extends ACLProvider {
> > +  var acls = new java.util.ArrayList[ACL]()
> > +
> > +  userSecret.foreach {
> > +    case (user, secret) => {
> > +      val id = new Id("digest", s"$user:$secret")
> > +      val acl = new ACL(ZooDefs.Perms.ALL, id)
> > +      acls.add(acl)
> > +    }
> > +  }
>
> I fought sbt for too long over java.util.ArrayList[ACL]() causing object
> java.util.ArrayList is not a value and I'm not sure how to appease the
> compiler to make this more idiomatic. Any ideas?
>
> —
> You are receiving this because you commented.
>
>
> Reply to this email directly or view it on GitHub
> <https://github.com/dcos/cosmos/pull/344/files/9ba7540b5921086c2a1148c4f4a180ce21125b31#r64592227>
>
-- 
José Armando García Sancio
Mesosphere

spacejam:
@jsancio I would like to create two tests:
1. ensures proper ACLs are added for znodes when a zk user and secret are provided
2. first creates a normal structure without ACLs, ensures that no ACL is set, then is restarted with a user, secret, and the `zookeeperAclAddToExisting` setting enabled which should add ACL's to previously written znodes under the chroot.

Could you give me some tips on where in the testing structure this should fit in?  Is there anything similar that I can look at as an example?  I'm a bit unfamiliar with how these integration tests can be run in a quick iteration loop, any tips?

cruhland:
@spacejam: Unit tests live under `src/test`; integration tests live under `src/it`. The integration tests start up a ZooKeeper test cluster [here](https://github.com/dcos/cosmos/blob/master/project/CosmosIntegrationTestServer.scala#L30-L32).

Unfortunately we don't have any example tests that use ZK, but maybe you can just use some of our other integration tests as examples. As for quick iteration, in the SBT console you can do `~it:test` to run the integration tests each time the source code changes. Using `it:test-only <class name>` (to run a single test class) doesn't seem to work; I'll file an issue.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -266,6 +274,11 @@ object Cosmos extends FinchServer {
         zkClient.close()
       }
 
+      if (zookeeperAclAddToExisting()) {
+        val root = zkUri.path.stripPrefix("/")
+        updateAcls(zkClient, aclProvider, List(root))
+      }
```
jsancio: @spacejam, I am not sure this is going to work when Cosmos is run for the first time. Currently, we have been trying to avoid blocking IO during initialization. The general reason for this is that they 1) tend to be racy and 2) slow down the time from fork to being available.

To get around this we add the necessary logic to the request handling code. For example if you look at ZooKeeperStorage one of the `read` operation try to read and if the znode doesn't exists it tries to create it. It feels to me like we want to add this code there. For example 1) try to read the znode, 2) if it doesn't exists create it with the right ACL. ZooKeeper allows you to specify the ACL when creating the node. During upgrade we need to tell the upgrade team that they need to set the correct ACL to existing znode before restarting Cosmos. We also need to tell the install team that they need to set the permission for `zkUri`. What do you think?

 cc @spacejam @alberts

spacejam: Your suggested procedure is how it's working right now (theoretically, still having testing issues).  By setting aclProvider all new znodes will have an ACL that the provider tells the CuratorFramework about.  We can flip the zookeeperAclAddToExisting flag when we want to do an upgrade, and back off after it starts up the first time.

jsancio: Cool! I didn't know that the ACLProvider was used by the `CreateBuilder` by default.

Regarding upgrade, I think we should do this as part of an upgrade script. It looks to me that the only way this is going to work is if we stop sending traffic to Cosmos, shutdown all of the Cosmos, turn on ACL in zookeeper, run upgrade script that creates the right ACLs and boot all of the Cosmos with the credentials.

Is this the planned upgrade flow? cc @alberts @spacejam

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -298,4 +311,52 @@ object Cosmos extends FinchServer {
...
+    case child :: otherChildren => {
+      val acls = aclProvider.getAclForPath(child)
+
+      zkClient.setACL()
+        .withACL(acls)
+        .forPath(child)
+
+      val newChildren = zkClient.getChildren.forPath(child).asScala.toList
+
+      updateAcls(zkClient, aclProvider, newChildren ++ otherChildren)
```
cruhland: How does this terminate? Won't the `paths` list grow or stay the same size on each recursive call?

spacejam: `child :: otherChildren` on 320 snips the head off on each successive call, I believe

cruhland: Ah, ignore me...didn't see that `newChildren` are the children of `child`, not its siblings.

jsancio:
I am taking over this PR.

jsancio:
Check it out @cruhland . I tested it manually and it works. I am not 100% sure how we can test this in our integration tests.

cruhland:
@jsancio I didn't have time today to review this very far, but I'll look at it first thing tomorrow. Could you rebase it on top of master before then? Thanks!

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
```diff
@@ -0,0 +1,95 @@
...
+    val zkUri = zookeeperUri()
+    logger.info("Using {} for the zookeeper connection", zkUri)
+
+    val zkClientBuilder = CuratorFrameworkFactory
+      .builder()
+      .namespace(zkUri.path.stripPrefix("/"))
+      .connectString(zkUri.connectString)
+      .retryPolicy(new ExponentialBackoffRetry(1000, 3))
+
+    val authInfo = sys.env.get("ZOOKEEPER_USER").zip(sys.env.get("ZOOKEEPER_SECRET")).map {
```
cruhland: Was there a specific reason for using environment variables instead of command-line flags?

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
```diff
@@ -0,0 +1,95 @@
...
+      case (user, secret) =>
+        (
+          s"$user:$secret".getBytes(StandardCharsets.UTF_8),
+          CosmosAclProvider(user, secret)
+        )
+    }
+
+    authInfo.foreach {
+      case (authBytes, aclProvider) =>
+        logger.info("Enabling authorization and ACL provider for ZookKeeper client")
```
cruhland: Nitpick: typo in `ZookKeeper`

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  private[this] def updateAcls(
+    zkClient: CuratorFramework,
+    aclProvider: ACLProvider
+  ): Unit = {
+    updateAcls(
+      zkClient,
+      aclProvider,
+      zkClient.getChildren.forPath("/").asScala.toList.map("/" + _)
```
cruhland: Do we also want to put ACLs on the root node (of the Cosmos namespace)? This skips the root.

cruhland:
Looks good. Just a few comments, and it looks like it's still missing some commits from master.

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
jsancio: More secure. Usually command line flags show in `ps`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
jsancio: No, we don't. That path is passed to Cosmos and not owned by Cosmos. DC/OS is going to make sure that the root znode is permissioned correctly.

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.zookeeper.data.ACL
+import org.apache.zookeeper.data.Id
+
+import com.mesosphere.cosmos.zookeeperUri
+
+object Clients {
+  val logger = org.slf4j.LoggerFactory.getLogger(getClass)
+
+  def createAndInitialize(): CuratorFramework = {
+    val zkUri = zookeeperUri()
```
BenWhitehead: Please pass this value into the method instead of reading directly from the flag. We want to keep flag interaction in `Cosmos.scala` so that it's as far out toward bootstrapping as possible.

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
BenWhitehead: Please pass this value into the method instead of reading directly from the env. We want to keep env interaction in `Cosmos.scala` so that it's as far out toward bootstrapping as possible.

jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
jsancio: Done.

Issue #342, opened 2016-05-23
Support multiple response object versions to Cosmos clients: cli, ui, scripts


jsancio:
Tasks:

1. Refactor the structure of the EndpointHandler and Router
1. Update request reader to route base on accept header
1. Implement the new data type and for the response content based on the accept header

PR #346, opened 2016-05-25
Support several expected values for the Accept header

cruhland:
This PR is mostly refactoring, with a rough pass at the functionality needed for #342. I haven't added tests yet since I'd like feedback on the approach. If it looks good, I'll add a commit to this PR with tests.

cc @jsancio @BenWhitehead

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
@@ -2,28 +2,25 @@ package com.mesosphere.cosmos
 
 import java.nio.file.Path
 
+import com.mesosphere.cosmos.circe.Decoders._
+import com.mesosphere.cosmos.circe.Encoders._
+import com.mesosphere.cosmos.circe.{DispatchingMediaTypedEncoder, MediaTypedDecoder}
+import com.mesosphere.cosmos.handler._
+import com.mesosphere.cosmos.http.MediaTypes
+import com.mesosphere.cosmos.model._
+import com.mesosphere.cosmos.repository.{PackageSourcesStorage, UniverseClient, ZooKeeperStorage}
 import com.netaporter.uri.Uri
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Request, Response, Status}
 import com.twitter.finagle.stats.{NullStatsReceiver, StatsReceiver}
-import com.twitter.util.Future
 import com.twitter.util.Try
 import io.circe.Json
-import io.circe.syntax.EncoderOps
 import io.finch._
 import io.finch.circe._
 import io.github.benwhitehead.finch.FinchServer
 import org.apache.curator.framework.CuratorFrameworkFactory
 import org.apache.curator.retry.ExponentialBackoffRetry
-
-import com.mesosphere.cosmos.circe.Decoders._
-import com.mesosphere.cosmos.circe.Encoders._
-import com.mesosphere.cosmos.handler._
-import com.mesosphere.cosmos.http.{MediaTypes, RequestSession}
-import com.mesosphere.cosmos.model._
-import com.mesosphere.cosmos.repository.PackageSourcesStorage
-import com.mesosphere.cosmos.repository.UniverseClient
-import com.mesosphere.cosmos.repository.ZooKeeperStorage
+import shapeless.HNil
 
 private[cosmos] final class Cosmos(
   uninstallHandler: EndpointHandler[UninstallRequest, UninstallResponse],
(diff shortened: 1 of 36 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -38,127 +36,47 @@ private[cosmos] final class Cosmos(
...
   val packageDeleteSource: Endpoint[Json] = {
-    def respond(t: (RequestSession, PackageRepositoryDeleteRequest)): Future[Output[Json]] = {
-      implicit val (session, request) = t
-      deleteRepositoryHandler(request).map { response =>
-        Ok(response.asJson).withContentType(Some(deleteRepositoryHandler.produces.show))
-      }
-    }
-
-    post("package" / "repository" / "delete" ? deleteRepositoryHandler.reader)(respond _)
+    deleteRepositoryHandler.route(reader => post("package" / "repository" / "delete" ? reader))
```
BenWhitehead: Great cleanup here!

BenWhitehead:
Can we add a test with some case classes and MediaTypes that are disconnected from all our existing model and use them to create a test that shows an endpoint can return different responses? Looking at the signature of `EndpointHandler` I'm failing to see how multiple response types will/would be expressed.

cruhland:
Yep @BenWhitehead that's the one thing I haven't dealt with yet. This change will work for "superficial" modifications to the response body, like omitting or including an optional field. I believe we could technically get away with that for 1.8 if we added `minDcosReleaseVersion` to our `DescribeResponse` model as an optional field, which we always fill in for "v2" responses.

In the long term though we'll need to allow for arbitrary response objects, so I'll look into the test you suggested.

cruhland:
Status update: after the team discussion today, I'm looking into changing the approach slightly. I'll add a comment here once I verify that the new approach works.

cruhland:
I have a way forward, just tying up some loose ends now. New commits coming by EOD.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -198,6 +198,11 @@ object Encoders {
 
   implicit val encodeCosmosError: Encoder[CosmosError] = deriveFor[CosmosError].encoder
 
+  implicit val encodeExampleVersionedAdt: Encoder[ExampleVersionedAdt] = {
+    val encoder = deriveFor[ExampleVersionedAdt].encoder
+    Encoder.instance(encoder.encodeObject(_).values.headOption.getOrElse(Json.obj()))
+  }
```
cruhland: This is a custom encoder for a "versioned response" ADT. It uses the derived encoder but removes the class annotations. See `ExampleVersionedAdt.scala` for the definition of the ADT.

Here's what the plain derived encoder would do for `Version2("bar", 42)`:
```json
{
  "Version2": {
    "a": "bar",
    "b": 42
  }
}
```

And here's what this custom encoder does:
```json
{
  "a": "bar",
  "b": 42
}
```

We want the latter result because clients can determine which version they're getting from the `Content-Type` header.

cosmos-server/src/test/scala/com/mesosphere/cosmos/handler/VersionedResponsesSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+  object Bar {
+    val MediaType: MediaType = versionedJson(2)
+    def fromGenericResponse(foobar: FoobarResponse): Bar = Bar(foobar.bar)
+  }
+
+  def versionedJson(version: Int): MediaType = {
+    MediaTypes.applicationJson.copy(parameters = Some(Map("version" -> s"v$version")))
+  }
+
+  // TODO cruhland: Move this to Encoders as a helper method, and test
```
cruhland: Remove this

cosmos-server/src/test/scala/com/mesosphere/cosmos/handler/VersionedResponsesSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+import com.mesosphere.cosmos.http.{MediaType, MediaTypes, RequestSession}
+import com.twitter.finagle.http.{Method, RequestBuilder}
+import com.twitter.io.Buf
+import com.twitter.util.{Await, Future, Try}
+import io.circe.generic.semiauto
+import io.circe.syntax._
+import io.circe.{Encoder, Json}
+import io.finch.{Endpoint, Input, Output}
+
+final class VersionedResponsesSpec extends UnitSpec {
```
cruhland: This test demonstrates how multiple response versions are dealt with.

cruhland:
@jsancio @BenWhitehead updated. This should satisfy the requirement of having separate case classes for each version of the response object. Please review and I can make changes as necessary.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -298,7 +213,7 @@ object Cosmos extends FinchServer {
       new PackageRepositoryListHandler(sourcesStorage),
       new PackageRepositoryAddHandler(sourcesStorage),
       new PackageRepositoryDeleteHandler(sourcesStorage),
-      CapabilitiesHandler()
+      new CapabilitiesHandler
```
jsancio: Why are are we doing this? Did we remove the `apply()` method?

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/ExampleVersionedAdt.scala
```diff
@@ -0,0 +1,12 @@
...
+/** Intended to be used only in tests.
+  *
+  * For some reason Shapeless fails to create a generic representation of this class if it is
+  * defined in test scope.
+  */
+sealed trait ExampleVersionedAdt
+case object Version0 extends ExampleVersionedAdt
+case class Version1(a: String) extends ExampleVersionedAdt
+case class Version2(a: String, b: Int) extends ExampleVersionedAdt
+case class Version3(b: Int, c: Double) extends ExampleVersionedAdt
```
jsancio: Please move this to be under `test` and not part of the `main` source code.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -198,6 +198,12 @@ object Encoders {
 
   implicit val encodeCosmosError: Encoder[CosmosError] = deriveFor[CosmosError].encoder
 
+  implicit val encodeExampleVersionedAdt: Encoder[ExampleVersionedAdt] = deriveFor[ExampleVersionedAdt].encoder
+//  implicit val encodeExampleVersionedAdt: Encoder[ExampleVersionedAdt] = {
+//    val encoder = deriveFor[ExampleVersionedAdt].encoder
+//    Encoder.instance(encoder.encodeObject(_).values.headOption.getOrElse(Json.obj()))
+//  }
+
```
jsancio: Please remove the comments and move to `test`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/EndpointHandler.scala
```diff
@@ -1,61 +1,53 @@
...
+  import EndpointHandler._
 
   def apply(request: Request)(implicit session: RequestSession): Future[Response]
 
-  private val printer: Printer = Printer.noSpaces.copy(dropNullKeys = true, preserveOrder = true)
-  lazy val encodeResponseType: EncodeResponse[Response] =
-    EncodeResponse.fromString[Response](produces.show, produces.parameters.flatMap(_.get("charset"))) { response =>
-      printer.pretty(encoder(response))
+  private[cosmos] def route(method: Method, path: String*): Endpoint[Json] = {
+    val endpointPath = path.foldLeft[Endpoint[HNil]](/)(_ / _)
```
jsancio: Very cool but why just take and Endpoint[HNil] in? The user can then do `route(get / "cool" / "path")`, no? If we do that then we can give this method a better name since what it looks like it is doing is extracting stuff, decoding, calling handler and decoding.

jsancio: If we do this then we can get rid of the `Method => Endpoint` map.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -2,166 +2,81 @@ package com.mesosphere.cosmos
 
 import java.nio.file.Path
 
+import com.mesosphere.cosmos.circe.Decoders._
+import com.mesosphere.cosmos.circe.Encoders._
+import com.mesosphere.cosmos.handler._
```
jsancio: Is this where we are getting all the `EndpointCoded` implicit?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/EndpointCodec.scala
```diff
@@ -0,0 +1,131 @@
...
+    requestClassTag: ClassTag[Request]
+  ): EndpointCodec[Request, Response, Response] = {
+    EndpointCodec(RequestReaders.standard[Request, Response, Response](accepts, produces), encoder)
+  }
+
+  private[this] def producesOnly[Response](
+    mediaType: MediaType
+  ): Seq[(MediaType, Response => Response)] = Seq((mediaType, identity))
+
+}
```
jsancio: Can we remove all of the `implicit`s above? We have an `EndpointHandler` for each `EndpointCodec`, no? If the mapping is one to one how are implicit helping? If the mapping is not one to one but they are going to get used in `Cosmos.scala` how are implicit helping?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: Since I was changing the constructors of all the `EndpointHandler` subclasses anyway, I updated this to be consistent with the others.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/ExampleVersionedAdt.scala
cruhland: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
cruhland: Whoops, done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/EndpointHandler.scala
cruhland: Will do.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: No, this is importing all of the `EndpointHandler` subclasses (it was in the previous version of the file too).

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/EndpointCodec.scala
cruhland: Yeah, I'm not sure what my reasoning was here, probably was thinking about separating concerns. But since it's a direct dependency it makes more sense to put this information in the `EndpointHandler` subclass files. Will update.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/EndpointHandler.scala
```diff
@@ -1,61 +1,28 @@
...
-      printer.pretty(encoder(response))
-    }
+  private[cosmos] def forRoute(route: Endpoint[HNil]): Endpoint[Json] = {
+    val endpoint = route ? requestReader
 
-}
-
-object EndpointHandler {
+    endpoint { context: EndpointContext[Request, Response, VersionedResponse] =>
+      this(context.requestBody)(context.session).map { response =>
```
cruhland: Change `this(...)` to `apply(...)`

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/EndpointHandler.scala
```diff
@@ -1,61 +1,27 @@
...
-    responseClassTag: ClassTag[Response],
-    session: RequestSession
-  ): EndpointHandler[Request, Response] = {
-    new EndpointHandler[Request, Response] {
-      val accepts = MediaTypes.applicationJson
-      val produces = MediaTypes.applicationJson
-      override def apply(v1: Request)(implicit session: RequestSession): Future[Response] = Future.value(resp)
+    endpoint { context: EndpointContext[Request, Response] =>
+      this(context.requestBody)(context.session).map { response =>
+        val encodedResponse = response.asJson(context.responseEncoder)
```
cruhland: I've updated this so that there's no longer a `responseFormatter: Response => VersionedResponse`; instead the `RequestReader` selects the appropriate `Encoder` based on the value of the `Accept` header and includes that in the `EndpointContext`. See `RequestReaders.scala`.

There's one issue remaining, which is that each `EndpointHandler` subclass still has an `implicit Encoder` constructor parameter. This works with the current code, because every endpoint only has one version of the response that can be encoded. When we update e.g. the `PackageDescribeHandler` to allow more than one `Encoder` to be specified, we'll need one `Encoder` for each acceptable `MediaType`, and so a single `implicit Encoder` parameter won't make any sense. I'm thinking about replacing that parameter with an `implicit MultiEncoder`, where `case class MultiEncoder[A](encoders: (MediaType, Encoder[A])*)`. These would be defined in `Encoders.scala` or a similar file, and imported wherever `EndpointHandler` subclasses are instantiated, e.g. in `Cosmos.scala`.

Take a look at how I've updated `VersionedResponsesSpec` to get an example of how the `MultiEncoder` would work. With the latest changes, we create the following `produces` value for the `FoobarHandler`'s `RequestReader`:

```scala
Seq(
  Foo.MediaType -> Foo.foobarEncoder,
  Bar.MediaType -> Bar.foobarEncoder
)
```

With `MultiEncoder`, there would instead be the following definition:

```scala
implicit val foobarEncoder: MultiEncoder[FoobarResponse] = MultiEncoder(
  Foo.MediaType -> Foo.foobarEncoder,
  Bar.MediaType -> Bar.foobarEncoder
)
```

The `RequestReader` factory method would then take an `implicit produces: MultiEncoder[Res]` as an argument.

Let's discuss further on Monday. cc @BenWhitehead @jsancio

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/MultiEncoder.scala
```diff
@@ -0,0 +1,40 @@
...
+  implicit val packageRepositoryDeleteEncoder: MultiEncoder[PackageRepositoryDeleteResponse] =
+    singleton(MediaTypes.PackageRepositoryDeleteResponse)
+  implicit val packageRepositoryListEncoder: MultiEncoder[PackageRepositoryListResponse] =
+    singleton(MediaTypes.PackageRepositoryListResponse)
+  implicit val packageSearchEncoder: MultiEncoder[SearchResponse] =
+    singleton(MediaTypes.SearchResponse)
+  implicit val packageUninstallEncoder: MultiEncoder[UninstallResponse] =
+    singleton(MediaTypes.UninstallResponse)
+
+  private[this] def singleton[A: Encoder](mediaType: MediaType): MultiEncoder[A] = {
```
cruhland: Rename this

cruhland: rename to `apply`

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/MultiEncoder.scala
```diff
@@ -0,0 +1,40 @@
+package com.mesosphere.cosmos.circe
+
+import com.mesosphere.cosmos.circe.Encoders._
+import com.mesosphere.cosmos.http.{MediaType, MediaTypes}
+import com.mesosphere.cosmos.model._
+import io.circe.Encoder
+
+/** Stores multiple [[com.mesosphere.cosmos.circe.RichEncoder]]s ordered by media type priority. */
+final case class MultiEncoder[A](encoders: RichEncoder[A]*)
```
cruhland: class with vals

cruhland: this could be a partial function

cruhland:
I'm out of time today, but I'll make sure to resolve the conflicts and merge this first thing tomorrow morning.

jsancio:
Yes! 🎉 🎈

PR #348, opened 2016-06-01
Improve PackageFileSchemaMismatch error to pass along the error that caused it

BenWhitehead:
@jsancio quick fix to the error message we ran into today and spoke about.

###### Before
```json
{
   "message": "Package file [index.json] does not match schema",
    "type": "PackageFileSchemaMismatch"
}
```

###### After
```json
{
    "data": {
        "errorMessage": "String: El(DownField(0.3.0),true),El(DownField(versions),true),El(DownArray,true),El(DownField(packages),true)"
    },
    "message": "Package file [index.json] does not match schema",
    "type": "PackageFileSchemaMismatch"
}
```

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
@@ -7,14 +7,12 @@ import java.time.LocalDateTime
 import java.util.Base64
 import java.util.concurrent.atomic.AtomicReference
 import java.util.zip.ZipInputStream
-
 import scala.util.matching.Regex
-
 import cats.data.Xor.{Left, Right}
 import cats.data._
-import cats.std.list._           // allows for traversU in verifySchema
+import cats.std.list._
 import cats.std.option._
-import cats.syntax.apply._       // provides |@|
+import cats.syntax.apply._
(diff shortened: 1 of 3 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
```diff
@@ -106,7 +103,7 @@ final class UniversePackageCache private (
       val repoIndex = parseJsonFile(indexFile)
         .toRightXor(new IndexNotFound(universeBundle))
         .flatMap { index =>
-          index.as[UniverseIndex].leftMap(_ => PackageFileSchemaMismatch("index.json"))
+          index.as[UniverseIndex].leftMap(PackageFileSchemaMismatch("index.json", _))
```
jsancio: Implicit lambda with `_` and `()` always worry me. Do you mind if we make this explicit like `failure => ...`?

jsancio:
LGTM! One minor comment.

cosmos-server/src/main/scala/com/mesosphere/cosmos/UniversePackageCache.scala
BenWhitehead: fixed

PR #352, opened 2016-06-03
Introduce new universe v3 repo schema.

BenWhitehead:
The following changes were made to the project structure to ensure we're
able to handle multiple versions of universe schema in the same
codebase.

* Introduce new `cosmos-json` module that contains circe
  endcoders/decoders for models defined in cosmos-model
* Create new package `com.mesosphere.universe.v2.model` to contain all the
  existing universe model objects and schemas
* Create new package `com.mesosphere.universe.v3.model` to contain all the
  new universe v3 repo model objects and schemas
* Restructure the way subsections are represented in scala packages, this
  helps reduce the number of ambiguous import scopes created in code
  (since you can do a "partial import" in scala and then import from the
  scope that has already been imported) the following pattern change
  took place:
  * `com.mesosphere.cosmos.model.thirdparty.adminrouter` -> `com.mesosphere.cosmos.thirdparty.adminrouter.model`
  * `com.mesosphere.cosmos.model.thirdparty.marathon` -> `com.mesosphere.cosmos.thirdparty.marathon.model`
  * `com.mesosphere.cosmos.model.thirdparty.mesos.master` -> `com.mesosphere.cosmos.thirdparty.mesos.master.model`
* Introduce a new encoder/decoder for each subset of the model
  * `com.mesosphere.universe.v2.model`
  * `com.mesosphere.universe.v3.model`
  * `com.mesosphere.cosmos.thirdparty.adminrouter.model`
  * `com.mesosphere.cosmos.thirdparty.marathon.model`
  * `com.mesosphere.cosmos.thirdparty.mesos.master.model`

PR #355, opened 2016-06-07
Move model of universe into model sub packages

BenWhitehead:
Meant to include this change with #352


jsancio:
Cool. LGTM. Let's talk about how we can remove code duplication. For example we have `v2.model.Images` and `v3.model.Images`. It would be nice if the process of adding new version is a bit lighter way than copying all of the "data types".

Issue #331, opened 2016-05-03
Create class to abstract fetching and loading of universe


cruhland:

cruhland:
This was fixed by #356.

PR #356, opened 2016-06-08
Create UniverseRepositoryFetcher

BenWhitehead:
`UniverseResoitoryFetcher`'s implementation will eventually replace the
implementation of UniverseClient.apply().

What it currently does:
1. Abstracts what it means to fetch a repo from a Universe
2. All request headers are setup for sending of the request
3. All response headers are evaluated to determine how to handle the
   response
4. New universe v3 json documents are supported
5. There is a code path for legacy v2 zip but is not yet implemented
6. Integration tests to very the code can correctly load and fetch a
   universe repo.

Define an Ordering[V3Package] that ensures the "correct" ordering for a
list of packages is available.

Create new CosmosInternalRepository case class to represent the process
internal repo.


BenWhitehead:
Fixes #331

BenWhitehead:
@cruhland please review

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
```diff
@@ -23,6 +23,10 @@ object PackageDefinition {
   case class ReleaseVersion(value: Int) {
     assert(value >= 0, s"Value $value is not >= 0")
   }
+  object ReleaseVersion {
+    import Ordering.Int
+    implicit val packageDefinitionReleaseVersionOrdering: Ordering[ReleaseVersion] = Ordering.by(_.value)
```
cruhland: I see, this is what you mentioned about ordering being ascending. This is fine for now, since it follows the "principle of least surprise" of putting integers in ascending order.

cosmos-model/src/test/scala/com/mesosphere/universe/v3/model/V3PackageSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+  "V3Package" - {
+    "Ordering should work" in {
+      val expected = List(
+        v3Package("pkg1", Version("1.0-1"), ReleaseVersion(1)),
+        v3Package("pkg1", Version("1.0-2"), ReleaseVersion(2)),
+        v3Package("pkg1", Version("1.0-3"), ReleaseVersion(3)),
+        v3Package("pkg2", Version("1.0"), ReleaseVersion(1)),
+        v3Package("pkg2", Version("2.0"), ReleaseVersion(2)),
+        v3Package("pkg3", Version("1.0"), ReleaseVersion(3)),
+        v3Package("pkg4", Version("1.0"), ReleaseVersion(4))
```
cruhland: Could you update this to have an example of the `Version` *decreasing* when the `ReleaseVersion` *increases*? I know we're not sorting by `Version` now, but having that case would potentially catch bugs in the future.

For example:

```scala
v3Package("pkg5", Version("1.9"), ReleaseVersion(1)),
v3Package("pkg5", Version("1.7"), ReleaseVersion(2))
```

BenWhitehead: Updated test with 3 new defs (Simulating a "backport release"):
```scala
v3Package("pkg5", Version("1.0-1"), ReleaseVersion(1)),
v3Package("pkg5", Version("2.0-1"), ReleaseVersion(2)),
v3Package("pkg5", Version("1.1-1"), ReleaseVersion(3))
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -21,3 +33,89 @@ object UniverseClient {
...
+  }
+
+  private[repository] def apply(uri: Uri, dcosReleaseVersion: DcosReleaseVersion)(implicit session: RequestSession): Future[CosmosInternalRepository] = {
+    Stat.timeFuture(stats.stat("fetch")) {
+      Future.value { uri.toURI.toURL.openConnection() }
+        .flatMap { conn =>
+          // Set headers on request
+          conn.setRequestProperty("Accept", MediaTypes.UniverseV3Repository.show)
+          conn.setRequestProperty("Accept-Encoding", "gzip")
+          conn.setRequestProperty("User-Agent", s"cosmos/$cosmosVersion dcos/${dcosReleaseVersion.show }")
```
cruhland: Nitpick: extra space after `dcosReleaseVersion.show`

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -21,3 +33,89 @@ object UniverseClient {
...
+              t._1.map(MediaTypeParser.parse(_).get) -> t._2
+            }
+            .map {
+              case (None, _)     => throw new IllegalStateException("Unable to process response without specified Content-Type")
+              case (Some(ct), ce) => ct -> ce
+            }
+            .map { case (contentType, contentEncoding) =>
+              val in = contentEncoding match {
+                case Some("gzip") => new GZIPInputStream(conn.getInputStream)
+                case _ => conn.getInputStream
```
cruhland: Should we fail-fast here if the encoding has an unexpected value?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -21,3 +33,89 @@ object UniverseClient {
...
+                  ??? //TODO: v2 -> v3 conversion code is wired in here
+                } else {
+                  throw new IllegalStateException(s"Unable to process response Content-Type: '${contentType.show}'")
+                }
+              }
+
+              decode[Repository](repoJsonString) match {
+                case Xor.Left(err)   => throw CirceError(err)
+                case Xor.Right(repo) => repo
+              }
```
cruhland: This is somewhat nitpicky, but I'm guessing that the v2 -> v3 conversion code will go directly from the v2 model case classes to the v3 model case classes. So this call to decode and the pattern match can just be moved inside the first clause of the if-else above.

BenWhitehead: I'd prefer to defer the specific change until we get the code that handles the v2 -> v3 stream. so we don't have to change twice.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: we can

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -21,3 +33,89 @@ object UniverseClient {
...
+
+          // Handle the response
+          Future.value { Option(conn.getHeaderField("Content-Type")) -> Option(conn.getHeaderField("Content-Encoding")) }
+            .map { t =>
+              t._1.map(MediaTypeParser.parse(_).get) -> t._2
+            }
+            .map {
+              case (None, _)     => throw new IllegalStateException("Unable to process response without specified Content-Type")
+              case (Some(ct), ce) => ct -> ce
+            }
```
cruhland: This might be a bit easier to read? It should behave the same as the above code.

```scala
Future.value {
  val contentType = Option(conn.getHeaderField("Content-Type"))
    .getOrElse(throw new IllegalStateException("Unable to process response without specified Content-Type"))
  val parsedContentType = MediaTypeParser.parse(contentType).get
  val contentEncoding = Option(conn.getHeaderField("Content-Encoding"))
  (contentType, contentEncoding)
}
```

cruhland: Hmm, modulo some exceptions getting thrown before others. The lines could be rearranged a bit.

BenWhitehead: will update

cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala
```diff
@@ -40,7 +40,7 @@ object Decoders {
     }
   implicit val decodePackageDefinitionReleaseVersion: Decoder[PackageDefinition.ReleaseVersion] =
     Decoder.instance[PackageDefinition.ReleaseVersion] { (c: HCursor) =>
-      Try { c.as[Int].map(PackageDefinition.ReleaseVersion) } match {
+      Try { c.as[Int].map(PackageDefinition.ReleaseVersion(_)) } match {
```
jsancio: `PackageDefinition.ReleaseVersion.apply` if you want to avoid the `_`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
cruhland: :+1:

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -21,3 +33,89 @@ object UniverseClient {
...
+
+  def apply(uri: Uri)(implicit session: RequestSession): Future[CosmosInternalRepository] = {
+    adminRouter.getDcosVersion().flatMap { dcosVersion =>
+      apply(uri, dcosVersion.version)
+    }
+  }
+
+  private[repository] def apply(uri: Uri, dcosReleaseVersion: DcosReleaseVersion)(implicit session: RequestSession): Future[CosmosInternalRepository] = {
+    Stat.timeFuture(stats.stat("fetch")) {
+      Future.value { uri.toURI.toURL.openConnection() }
```
jsancio: If we use `Future.value` then the `apply` method is going to throw if `uri.toURI.toURL.openConnection()` throws. I think you want to use `Future.apply` or just `Future`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -21,3 +33,89 @@ object UniverseClient {
...
+          // Set headers on request
+          conn.setRequestProperty("Accept", MediaTypes.UniverseV3Repository.show)
+          conn.setRequestProperty("Accept-Encoding", "gzip")
+          conn.setRequestProperty("User-Agent", s"cosmos/$cosmosVersion dcos/${dcosReleaseVersion.show}")
+          session.authorization.foreach { auth =>
+            conn.setRequestProperty("Authorization", auth.headerValue)
+          }
+
+          // Handle the response
+          Future.value { Option(conn.getHeaderField("Content-Type")) -> Option(conn.getHeaderField("Content-Encoding")) }
```
jsancio: Can we standardize on using `Future`? Every time I see `Future.value` we need to prove that when the argument gets evaluated it can never throws. Is that case with with `getHeaderField`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -21,3 +33,89 @@ object UniverseClient {
...
+            .map { t =>
+              t._1.map(MediaTypeParser.parse(_).get) -> t._2
```
jsancio: Let's deconstruct the tuple. Easier to read.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -21,3 +33,89 @@ object UniverseClient {
...
+      case v3: V3Package => v3
+      case V2Package(
+        _, name, version, relVersion, maintainer, description, marathon, tags, selected,
+        scm, website, framework, preInNotes, postInNotes, postUnNotes, licenses, resource, config, pip
+      ) => V3Package(V3PackagingVersion.instance, name, version, relVersion, maintainer, description, tags, selected,
+        scm, website, framework, preInNotes, postInNotes, postUnNotes, licenses,
+        marathon = Some(marathon), resource = resource, config = config, pip = pip
+      )
+    }
+  }
```
jsancio: I think @cruhland is going to introduce a type class called `Converter[R, T]` for converting from `R` to `T`. Maybe we can introduce it in this PR. I am also going to write a bunch of these type classes so it is nice to share them.

project/plugins.sbt
```diff
@@ -6,3 +6,4 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.8.1")
 
 addSbtPlugin("org.scala-sbt.plugins" % "sbt-onejar" % "0.8")
 
+addSbtPlugin("com.github.sdb" % "xsbt-filter" % "0.4")
```
jsancio: Do we need this because of the `build.properties`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -21,3 +33,89 @@ object UniverseClient {
...
+    Option(props.getProperty("cosmos.version")).getOrElse("unknown-version")
+  }
+
+  def apply(uri: Uri)(implicit session: RequestSession): Future[CosmosInternalRepository] = {
+    adminRouter.getDcosVersion().flatMap { dcosVersion =>
+      apply(uri, dcosVersion.version)
+    }
+  }
+
+  private[repository] def apply(uri: Uri, dcosReleaseVersion: DcosReleaseVersion)(implicit session: RequestSession): Future[CosmosInternalRepository] = {
```
cruhland: There's a good amount of error handling in this method that isn't being tested. Could the side-effecting code that deals with `conn` be separated out from the pure code that comes after it (parsing header values, decoding the response body)? Then we could at least have unit tests for the pure stuff.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
cruhland: Ah, if it it's useful for you to have it as well @jsancio then maybe we should add it in this PR. @BenWhitehead ask me for more details.

cruhland:
Done adding comments for now.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: Instead of writing our own, we can use `"com.twitter" % "bijection-core" %% "0.9.2"` to give us the type classes to do this.

project/plugins.sbt
BenWhitehead: Correct.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: I'll relax to using apply

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: I'll relax to using apply

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: will be replaced with a recommendation from charles above.

PR #357, opened 2016-06-08
Update v3 repo schema to use command object instead of pip array directly

BenWhitehead:
Correlates to https://github.com/mesosphere/universe/pull/482


cruhland:
Looks good to me. :ship: it when ready.

PR #359, opened 2016-06-09
Define package describe response encoder supporting versions v1 and v2

cruhland:
Uses Twitter's Bijection library to define converters between case
classes.

This is a starting point, just looking for early feedback before I get to the other responses that need to be updated.

### DIFF
--- a/cosmos-model/src/main/scala/com/mesosphere/cosmos/model/package.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/cosmos/model/package.scala
@@ -1,0 +1,9 @@
+package com.mesosphere.cosmos
+
+import com.mesosphere.universe.v3.model.V3Package
+
+package object model {
+
+  type DescribeResponse = V3Package
+
+}
--- a/cosmos-model/src/main/scala/com/mesosphere/cosmos/model/v1/DescribeResponse.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/cosmos/model/v1/DescribeResponse.scala
@@ -1,4 +1,4 @@
-package com.mesosphere.cosmos.model
+package com.mesosphere.cosmos.model.v1
 
 import com.mesosphere.universe.v2.model.{Command, PackageDetails, Resource}
 import io.circe.JsonObject
--- a/cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
+++ b/cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
@@ -100,7 +100,7 @@ final class PackageDescribeSpec extends FreeSpec with TableDrivenPropertyChecks 
   ): Response = {
     val request = apiClient.requestBuilder(DescribeEndpoint)
       .addHeader("Content-Type", MediaTypes.DescribeRequest.show)
-      .addHeader("Accept", MediaTypes.DescribeResponse.show)
+      .addHeader("Accept", MediaTypes.V1DescribeResponse.show)
       .buildPost(Buf.Utf8(describeRequest.asJson.noSpaces))
     apiClient(request)
   }
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
@@ -4,11 +4,13 @@ import java.nio.file.Path
 
 import com.mesosphere.cosmos.circe.Decoders._
 import com.mesosphere.cosmos.circe.Encoders._
-import com.mesosphere.cosmos.circe.{DispatchingMediaTypedEncoder, MediaTypedDecoder}
+import com.mesosphere.cosmos.circe.{DispatchingMediaTypedEncoder, MediaTypedDecoder, MediaTypedEncoder}
 import com.mesosphere.cosmos.handler._
 import com.mesosphere.cosmos.http.MediaTypes
 import com.mesosphere.cosmos.model._
 import com.mesosphere.cosmos.repository.{PackageSourcesStorage, UniverseClient, ZooKeeperStorage}
+import com.mesosphere.universe.v3.circe.{Encoders => V3Encoders}
+import com.mesosphere.universe.v3.model.V3Package
 import com.netaporter.uri.Uri
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Request, Response, Status}
@@ -27,7 +29,7 @@ private[cosmos] final class Cosmos(
   packageInstallHandler: EndpointHandler[InstallRequest, InstallResponse],
   packageRenderHandler: EndpointHandler[RenderRequest, RenderResponse],
   packageSearchHandler: EndpointHandler[SearchRequest, SearchResponse],
-  packageDescribeHandler: EndpointHandler[DescribeRequest, DescribeResponse],
+  packageDescribeHandler: EndpointHandler[DescribeRequest, model.v1.DescribeResponse],
   packageListVersionsHandler: EndpointHandler[ListVersionsRequest, ListVersionsResponse],
   listHandler: EndpointHandler[ListRequest, ListResponse],
   listRepositoryHandler: EndpointHandler[PackageRepositoryListRequest, PackageRepositoryListResponse],
@@ -272,8 +274,21 @@ object Cosmos extends FinchServer {
   implicit val packageListVersionsEncoder: DispatchingMediaTypedEncoder[ListVersionsResponse] =
     DispatchingMediaTypedEncoder(MediaTypes.ListVersionsResponse)
 
-  implicit val packageDescribeEncoder: DispatchingMediaTypedEncoder[DescribeResponse] =
-    DispatchingMediaTypedEncoder(MediaTypes.DescribeResponse)
+  implicit val packageDescribeV1Encoder: DispatchingMediaTypedEncoder[model.v1.DescribeResponse] =
+    DispatchingMediaTypedEncoder(MediaTypes.V1DescribeResponse)
+
+  implicit val packageDescribeEncoder: DispatchingMediaTypedEncoder[model.DescribeResponse] = {
+    DispatchingMediaTypedEncoder(Seq(
+      MediaTypedEncoder(
+        encoder = V3Encoders.encodeV3Package,
+        mediaType = MediaTypes.V2DescribeResponse
+      ),
+      MediaTypedEncoder(
+        encoder = encodeDescribeResponse.contramap(converter.Response.v3PackageToDescribeResponse),
+        mediaType = MediaTypes.V1DescribeResponse
+      )
+    ))
+  }
 
   implicit val packageInstallEncoder: DispatchingMediaTypedEncoder[InstallResponse] =
     DispatchingMediaTypedEncoder(MediaTypes.InstallResponse)
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
@@ -1,11 +1,11 @@
 package com.mesosphere.cosmos.http
 
 object MediaTypes {
-  private[this] def vnd(kind: String): MediaType =
+  private[this] def vnd(kind: String, version: Int = 1): MediaType =
     MediaType(
       "application",
       MediaTypeSubType(s"vnd.dcos.package.$kind", Some("json")),
-      Map("charset" -> "utf-8", "version" -> "v1")
+      Map("charset" -> "utf-8", "version" -> ("v" + version))
     )
 
   val any = MediaType("*", MediaTypeSubType("*"))
@@ -26,7 +26,6 @@ object MediaTypes {
   val SearchRequest = vnd("search-request")
   val SearchResponse = vnd("search-response")
   val DescribeRequest = vnd("describe-request")
-  val DescribeResponse = vnd("describe-response")
   val ListVersionsRequest = vnd("list-versions-request")
   val ListVersionsResponse = vnd("list-versions-response")
   val CapabilitiesResponse = MediaType(
@@ -50,5 +49,8 @@ object MediaTypes {
     MediaTypeSubType("vnd.dcos.universe.repo", Some("json")),
     Map("charset" -> "utf-8", "version" -> "v3")
   )
+
+  val V1DescribeResponse = vnd("describe-response", 1)
+  val V2DescribeResponse = vnd("describe-response", 2)
 
 }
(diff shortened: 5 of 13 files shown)


cosmos-model/src/main/scala/com/mesosphere/cosmos/model/v1/DescribeResponse.scala
```diff
@@ -1,4 +1,4 @@
-package com.mesosphere.cosmos.model
+package com.mesosphere.cosmos.model.v1
```
cruhland: Moved `DescribeResponse` to a new package to indicate that it's an "old" `v1` response object.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Converters.scala
```diff
@@ -0,0 +1,133 @@
...
+import com.mesosphere.universe.v2.model.PackageDetails
+import com.mesosphere.universe.v2.model.PackageDetailsVersion
+import com.mesosphere.universe.v2.model.PackagingVersion
+import com.mesosphere.universe.v2.model.{Resource => V2Resource}
+import com.mesosphere.universe.v3.model.{Assets => V3Assets}
+import com.mesosphere.universe.v3.model.{Container => V3Container}
+import com.mesosphere.universe.v3.model.{Images => V3Images}
+import com.mesosphere.universe.v3.model.{License => V3License}
+import com.mesosphere.universe.v3.model.PackageDefinition
+import com.mesosphere.universe.v3.model.{Resource => V3Resource}
```
cruhland: Suggestions for importing overloaded names?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Converters.scala
```diff
@@ -0,0 +1,133 @@
...
+import com.twitter.bijection.Conversion.asMethod
+import com.twitter.bijection.{Bijection, Injection}
+import com.twitter.io.Charsets
+
+import scala.util.{Failure, Try}
+
+object Converters {
+
+  implicit val v3PackageToDescribeResponse: Injection[V3Package, DescribeResponse] = {
+    forwardOnly { v3Package =>
```
cruhland: For now, I'm defining these injections in one direction only, because I wanted to get a PR up quickly. Should we adopt a rule to always define these in both directions, even though one direction might not be used?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Converters.scala
```diff
@@ -0,0 +1,133 @@
...
+      DescribeResponse(
+        `package` = v3Package.as[PackageDetails],
+        marathonMustache = v3Package.marathon match {
+          case Some(marathon) =>
+            new String(ByteBuffers.getBytes(marathon.v2AppMustacheTemplate), Charsets.Utf8)
+          case _ =>
+            ""
```
cruhland: Not sure if this is the right thing to do when the Marathon template is not defined in a `version-3.x` package.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Converters.scala
```diff
@@ -0,0 +1,133 @@
...
+        },
+        command = v3Package.pip.map(Command(_)),
+        config = v3Package.config,
+        resource = v3Package.resource.as[Option[V2Resource]]
+      )
+    }
+  }
+
+  implicit val v3PackageToPackageDetails: Injection[V3Package, PackageDetails] = {
```
cruhland: This is technically not an injection because information is lost going from a `V3Package` to `PackageDetails`. Should it be a simple function instead?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Converters.scala
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  implicit val universeV3ContainerToUniverseV2Container: Bijection[V3Container, V2Container] = {
+    val v3ToV2 = (v3: V3Container) => V2Container(docker = v3.docker)
+    val v2ToV3 = (v2: V2Container) => V3Container(docker = v2.docker)
+
+    Bijection.build(v3ToV2)(v2ToV3)
+  }
+
+  implicit val universeV3ImagesToUniverseV2Images: Bijection[V3Images, V2Images] = {
```
cruhland: Here's an example of `v2` and `v3` data structures being identical. There are a few others in this file also. Given the large amount of code required for this, would it make sense for us to reuse older versions of data structures if they haven't changed in a new version?

For example, we could have the `v3` `Resource` case class refer to the `v2` `Assets` and `Images` case classes because they haven't changed.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Converters.scala
```diff
@@ -0,0 +1,133 @@
...
+      iconMedium = v2.iconMedium,
+      iconLarge = v2.iconLarge,
+      screenshots = v2.screenshots
+    )
+
+    Bijection.build(v3ToV2)(v2ToV3)
+  }
+
+  def forwardOnly[A, B](f: A => B): Injection[A, B] = {
+    Injection.build(f)(_ => Failure(new RuntimeException("Cannot convert backwards")))
```
cruhland: Since a failed conversion is a programming error, we should catch it in testing so end users shouldn't see it. Would a custom exception type be useful anyway?

cruhland:
@jsancio ready for review. I'll rebase this shortly.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Converters.scala
```diff
@@ -0,0 +1,141 @@
+package com.mesosphere.cosmos
```
jsancio: Hey @cruhland, let's namespace them based on the grouping of models. For example we can have a set of bijections for Universe and another set of bijections for universe v3 to response objects.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -100,7 +100,7 @@ final class PackageDescribeSpec extends FreeSpec with TableDrivenPropertyChecks
   ): Response = {
     val request = apiClient.requestBuilder(DescribeEndpoint)
       .addHeader("Content-Type", MediaTypes.DescribeRequest.show)
-      .addHeader("Accept", MediaTypes.DescribeResponse.show)
+      .addHeader("Accept", MediaTypes.DescribeResponseV1.show)
```
jsancio: Sometime we use `V#` a suffix and sometimes we use it as a prefix. Let's pick one and go with it.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -272,8 +275,21 @@ object Cosmos extends FinchServer {
...
+      MediaTypedEncoder(
+        encoder = V3Encoders.encodeV3Package,
+        mediaType = MediaTypes.DescribeResponseV2
+      ),
+      MediaTypedEncoder(
+        encoder = encodeDescribeResponse.contramap(Converters.v3PackageToDescribeResponse),
+        mediaType = MediaTypes.DescribeResponseV1
+      )
+    ))
+  }
```
jsancio: Okay for now but I would like us to move these implicits out of here in a future PR.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -272,8 +275,21 @@ object Cosmos extends FinchServer {
...
+    DispatchingMediaTypedEncoder(MediaTypes.DescribeResponseV1)
+
+  implicit val packageDescribeEncoder: DispatchingMediaTypedEncoder[V3Package] = {
+    DispatchingMediaTypedEncoder(Seq(
+      MediaTypedEncoder(
+        encoder = V3Encoders.encodeV3Package,
+        mediaType = MediaTypes.DescribeResponseV2
+      ),
+      MediaTypedEncoder(
+        encoder = encodeDescribeResponse.contramap(Converters.v3PackageToDescribeResponse),
```
jsancio: Very cool!

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -272,8 +275,21 @@ object Cosmos extends FinchServer {
   implicit val packageListVersionsEncoder: DispatchingMediaTypedEncoder[ListVersionsResponse] =
     DispatchingMediaTypedEncoder(MediaTypes.ListVersionsResponse)
 
-  implicit val packageDescribeEncoder: DispatchingMediaTypedEncoder[DescribeResponse] =
-    DispatchingMediaTypedEncoder(MediaTypes.DescribeResponse)
+  implicit val packageDescribeV1Encoder: DispatchingMediaTypedEncoder[DescribeResponse] =
+    DispatchingMediaTypedEncoder(MediaTypes.DescribeResponseV1)
+
+  implicit val packageDescribeEncoder: DispatchingMediaTypedEncoder[V3Package] = {
```
jsancio: Why is the type `V3Package` and not `V2DescribeResponse`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
```diff
@@ -1,11 +1,11 @@
 package com.mesosphere.cosmos.http
 
 object MediaTypes {
-  private[this] def vnd(kind: String): MediaType =
+  private[this] def vnd(kind: String, version: Int = 1): MediaType =
```
jsancio: Okay but it would be nice to remove the default and make this an enumeration.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
```diff
@@ -51,4 +50,7 @@ object MediaTypes {
     Map("charset" -> "utf-8", "version" -> "v3")
   )
 
+  val DescribeResponseV1 = vnd("describe-response", 1)
+  val DescribeResponseV2 = vnd("describe-response", 2)
```
jsancio: I see. Ignore my comment above regarding the `vnd` method since it is private.

jsancio:
LGTM. Let's merge this once my comments are addressed.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: The type parameter is the one from the `EndpointHandler`, i.e. the "business logic" response. Once `PackageDescribeHandler` is updated to use the Universe version-3.x model, it will use this `Encoder`, but for now it's using the one above it, which has `DescribeResponse` as the type parameter as expected.

jsancio: Right but don't we want the business logic handler to be `DescribeRequest => v2.DescribeResponse`?

cruhland: There's no `v2.DescribeResponse`, because it would be identical to `universe.v3.V3Package`. We could create one for symmetry but I don't know if that would be worth it.

jsancio: Okay. Maybe we can have a type alias. I think consistency is important.

cosmos-model/src/main/scala/com/mesosphere/cosmos/model/package.scala
```diff
@@ -0,0 +1,9 @@
+package com.mesosphere.cosmos.model
+
+import com.mesosphere.universe.v3.model.V3Package
+
+object `package` {
```
jsancio: How about `package object pkg { ... }`? We can rename the file to `pkg`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -27,7 +29,7 @@ private[cosmos] final class Cosmos(
   packageInstallHandler: EndpointHandler[InstallRequest, InstallResponse],
   packageRenderHandler: EndpointHandler[RenderRequest, RenderResponse],
   packageSearchHandler: EndpointHandler[SearchRequest, SearchResponse],
-  packageDescribeHandler: EndpointHandler[DescribeRequest, DescribeResponse],
+  packageDescribeHandler: EndpointHandler[DescribeRequest, model.v1.DescribeResponse],
```
jsancio: I like this pattern! In the future maybe we just import `model`.

cosmos-model/src/main/scala/com/mesosphere/cosmos/model/package.scala
cruhland: That would put the type alias in a package named `com.mesosphere.cosmos.model.pkg`. I did it this way so that the full package name would be at the top of the file. It could also be done as:

```scala
package com.mesosphere.cosmos

package object model {
  // ...
}
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -272,8 +274,21 @@ object Cosmos extends FinchServer {
   implicit val packageListVersionsEncoder: DispatchingMediaTypedEncoder[ListVersionsResponse] =
     DispatchingMediaTypedEncoder(MediaTypes.ListVersionsResponse)
 
-  implicit val packageDescribeEncoder: DispatchingMediaTypedEncoder[DescribeResponse] =
-    DispatchingMediaTypedEncoder(MediaTypes.DescribeResponse)
+  implicit val packageDescribeV1Encoder: DispatchingMediaTypedEncoder[model.v1.DescribeResponse] =
+    DispatchingMediaTypedEncoder(MediaTypes.V1DescribeResponse)
+
+  implicit val packageDescribeEncoder: DispatchingMediaTypedEncoder[model.DescribeResponse] = {
```
jsancio: Interesting. Is this the type is ``object `package` { ... }``?

cosmos-model/src/main/scala/com/mesosphere/cosmos/model/package.scala
jsancio: Also, if you look at the Scala library they do this pattern:
```
cat .../com/mesosphere/cosmos/model/package.scala
package com.mesosphere.cosmos

package object model {
  ...
}
```

jsancio:
@cruhland LGTM. One minor command regarding ``object `package```.

PR #361, opened 2016-06-10
handle decoded marathon labels for v2 + v3

tamarrow:


cosmos-model/src/main/scala/com/mesosphere/cosmos/thirdparty/marathon/model/MarathonApp.scala
```diff
@@ -19,6 +19,10 @@ case class MarathonApp(
   def packageVersion: Option[PackageDetailsVersion] = labels.get(MarathonApp.versionLabel).map(PackageDetailsVersion)
 
   def packageRepository: Option[String] = labels.get(MarathonApp.repositoryLabel)
+
+  def packagingVersion: String = labels.getOrElse(MarathonApp.registryVersionLabel, "2.0")
```
cruhland: Could you make the type of this field `com.mesosphere.universe.v3.model.PackagingVersion`, and write a `Decoder` to parse it? We should avoid using magic constants like `"2.0"` where possible.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -55,6 +50,41 @@ private[cosmos] final class ListHandler(
...
+
+  private[this] def newDecodePackageFromMarathon(
+    app: MarathonApp
+  ): NewInstalledPackageInformation = {
+    val b64PkgInfo = app.packageMetadata.getOrElse("")
+    val pkgInfoBytes = Base64.getDecoder.decode(b64PkgInfo)
+    val pkgInfoString = new String(pkgInfoBytes, StandardCharsets.UTF_8)
+    val packagingVersion = app.packagingVersion
+
+    if (packagingVersion == "2.0") {
```
cruhland: There's actually a `Decoder` which handles this already, as long as the encoded document has a `packagingVersion` field: https://github.com/dcos/cosmos/blob/5bd41cb6ec7649edf224f1ebc024a4460c211020/cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala#L55-L62

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -33,15 +36,7 @@ private[cosmos] final class ListHandler(
...
-                    case None =>
-                      val b64PkgInfo = app.labels(MarathonApp.metadataLabel)
-                      val pkgInfoBytes = Base64.getDecoder.decode(b64PkgInfo)
-                      val pkgInfoString = new String(pkgInfoBytes, StandardCharsets.UTF_8)
-
-                      decode[PackageDetails](pkgInfoString) match {
-                        case Xor.Left(err) => throw new CirceError(err)
-                        case Xor.Right(pkgDetails) => InstalledPackageInformation(pkgDetails)
-                      }
+                    case None => decodePackageFromMarathon(app)
```
cruhland: Even better than putting this into a function is passing the function in as a dependency. Then we can more easily switch to the new version when we're ready. Writing tests also becomes easier. Speaking of which, we should probably have some unit tests for the new decode logic.

cosmos-model/src/main/scala/com/mesosphere/cosmos/model/Installation.scala
```diff
@@ -2,7 +2,13 @@ package com.mesosphere.cosmos.model
 
 import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
 
+// TODO: remove
```
cruhland: We can't remove this because it gets encoded as part of the `v1` response to `ListHandler`. Instead, it should be moved to the `com.mesosphere.cosmos.model.v1` package where I've put the old `DescribeResponse`.

cosmos-model/src/main/scala/com/mesosphere/cosmos/model/Installation.scala
```diff
@@ -2,7 +2,13 @@ package com.mesosphere.cosmos.model
...
 case class Installation(
   appId: AppId,
   packageInformation: InstalledPackageInformation
 )
+
+case class NewInstallation(
```
cruhland: Once `Installation` is moved this can just be called `Installation`. Similarly for `InstalledPackageInformation`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -55,6 +50,41 @@ private[cosmos] final class ListHandler(
...
+    val pkgInfoString = new String(pkgInfoBytes, StandardCharsets.UTF_8)
+    val packagingVersion = app.packagingVersion
+
+    if (packagingVersion == "2.0") {
+      decode[V2Package](pkgInfoString) match {
+        case Xor.Left(err) => throw new CirceError(err)
+        case Xor.Right(pkgDetails) => NewInstalledPackageInformation(pkgDetails)
+      }
+    } else {
+      decode[V3Package](pkgInfoString) match {
```
cruhland: How can we be sure this will work? The code that writes the label hasn't been updated to write a `V2`/`V3Package`. We should probably have that encoding logic and this decoding logic in their own class as I mentioned in another comment.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v2/model/Installation.scala
```diff
@@ -0,0 +1,8 @@
+package com.mesosphere.cosmos.rpc.v2.model
+
+import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
+
+case class Installation(
+  appId: AppId,
+  packageInformation: InstalledPackageInformation
```
cruhland: Why not have this field be `packageDefinition: PackageDefinition`, and then we don't need a new `InstalledPackageInformation`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -9,8 +9,11 @@ import com.mesosphere.cosmos.http.RequestSession
 import com.mesosphere.cosmos.thirdparty.marathon.model.MarathonApp
 import com.mesosphere.cosmos.repository.CosmosRepository
 import com.mesosphere.cosmos.rpc.v1.model.{Installation, InstalledPackageInformation, ListRequest, ListResponse}
+import com.mesosphere.cosmos.rpc.v2.model.{Installation => V2Installation, InstalledPackageInformation => V2InstalledPackageInformation}
 import com.mesosphere.universe.v2.model.{PackageDetails, ReleaseVersion}
 import com.mesosphere.universe.v2.circe.Decoders._
+import com.mesosphere.universe.v3.circe.Decoders._
+import com.mesosphere.universe.v3.model.{PackageDefinition, V2Package, V3Package}
```
cruhland: To make things more explicit, and avoid having to rename imports, let's just have:

```scala
```

Then refer to the types as `rpc.v2.model.Installation` or `universe.v3.model.PackageDefinition`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -55,6 +50,33 @@ private[cosmos] final class ListHandler(
     }
   }
 
+  // TODO: remove
```
cruhland: Let's use the convention Jose asked for in my recent PR: put `// TODO (version): Rename this to <original-name>` above the `V3` definition. Then we can search for `version` to find the relevant to-dos.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -55,6 +50,33 @@ private[cosmos] final class ListHandler(
...
+    app: MarathonApp
+  ): V2InstalledPackageInformation = {
+    val b64PkgInfo = app.packageMetadata.getOrElse("")
+    val pkgInfoBytes = Base64.getDecoder.decode(b64PkgInfo)
+    val pkgInfoString = new String(pkgInfoBytes, StandardCharsets.UTF_8)
+
+    decode[PackageDefinition](pkgInfoString) match {
+      case Xor.Left(err) => throw new CirceError(err)
+      case Xor.Right(pkgDetails) => V2InstalledPackageInformation(pkgDetails)
+    }
```
cruhland: Only this `decode`-pattern-match block must be different to handle the Universe 3.0 data types. The other lines can move back to their original location...less duplication to worry about.

cruhland:
@tamarrow looks good, just a few more comments. I guess you'll need to rebase too.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v2/model/Installation.scala
tamarrow: done

tamarrow:
@cruhland updated

cruhland:
LGTM :smile:

PR #362, opened 2016-06-10
Move request/response case classes to an `rpc` package with versioning

cruhland:


cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v2/model/package.scala
```diff
@@ -0,0 +1,11 @@
+package com.mesosphere.cosmos.rpc.v2
+
+import com.mesosphere.universe.v3.model.V3Package
+
+package object model {
+
+  // These type aliases are immutable, like everything in this package
+  // They keep naming consistent across versions
+  type DescribeResponse = V3Package
```
cruhland: @BenWhitehead this kind of type alias should be acceptable, correct? It keeps the version information explicit.

BenWhitehead:
:shipit:

PR #363, opened 2016-06-10
Move request/response encoders/decoders to the versioned `rpc` packages

cruhland:
This is the last big "repackaging" change.


BenWhitehead:
LGTM, double check that any unit tests for encoders/decoders are moved to the corresponding package as well.

cruhland:
Looks like the only specific tests are for `CosmosError`, whose encoders/decoders didn't move. So I'll merge this once the tests pass.

PR #373, opened 2016-06-14
Universe v2 to v3 conversion

jsancio:
I tested this manually by fetching from https://universe.mesosphere.com/repo.

### DIFF
--- a/cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala
+++ b/cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala
@@ -67,7 +67,7 @@ object Decoders {
   implicit val decodeV2PackagingVersion: Decoder[V2PackagingVersion] = {
     Decoder.decodeString.map(V2PackagingVersion(_))
   }
-  implicit val decodeV2Resource: Decoder[V2Resource] = deriveFor[V2Resource].decoder
+  implicit val decodeV3V2Resource: Decoder[V2Resource] = deriveFor[V2Resource].decoder
 
   implicit val decodeV3Package: Decoder[V3Package] = deriveFor[V3Package].decoder
   implicit val decodeV3PackagingVersion: Decoder[V3PackagingVersion] = {
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
@@ -1,30 +1,14 @@
 package com.mesosphere.cosmos.converter
 
 import com.mesosphere.cosmos.converter.Universe._
+import com.mesosphere.cosmos.internal
 import com.mesosphere.cosmos.rpc
 import com.mesosphere.universe
-import com.mesosphere.universe.common.ByteBuffers
 import com.twitter.bijection.Conversion.asMethod
 import com.twitter.io.Charsets
+import scala.util.Try
 
 object Response {
-
-  def v3PackageToDescribeResponse(
-    v3Package: universe.v3.model.V3Package
-  ): rpc.v1.model.DescribeResponse = {
-    rpc.v1.model.DescribeResponse(
-      `package` = v3V3PackageToV2PackageDetails(v3Package),
-      marathonMustache = v3Package.marathon match {
-        case Some(marathon) =>
-          new String(ByteBuffers.getBytes(marathon.v2AppMustacheTemplate), Charsets.Utf8)
-        case _ =>
-          ""
-      },
-      command = v3Package.command.as[Option[universe.v2.model.Command]],
-      config = v3Package.config,
-      resource = v3Package.resource.map(v3V3ResourceToV2Resource)
-    )
-  }
 
   def v2InstallResponseToV1InstallResponse(x: rpc.v2.model.InstallResponse): rpc.v1.model.InstallResponse = {
     rpc.v1.model.InstallResponse(
@@ -34,25 +18,25 @@ object Response {
     )
   }
 
-  private[this] def v3V3PackageToV2PackageDetails(
-    v3Package: universe.v3.model.V3Package
-  ): universe.v2.model.PackageDetails = {
-    universe.v2.model.PackageDetails(
-      packagingVersion = v3Package.packagingVersion.as[universe.v2.model.PackagingVersion],
-      name = v3Package.name,
-      version = v3Package.version.as[universe.v2.model.PackageDetailsVersion],
-      maintainer = v3Package.maintainer,
-      description = v3Package.description,
-      tags = v3Package.tags.as[List[String]],
-      selected = v3Package.selected,
-      scm = v3Package.scm,
-      website = v3Package.website,
-      framework = v3Package.framework,
-      preInstallNotes = v3Package.preInstallNotes,
-      postInstallNotes = v3Package.postInstallNotes,
-      postUninstallNotes = v3Package.postUninstallNotes,
-      licenses = v3Package.licenses.as[Option[List[universe.v2.model.License]]]
-    )
+  def packageDefinitionToDescribeResponse(
+      packageDefinition: internal.model.PackageDefinition)
+    : rpc.v1.model.DescribeResponse = {
+    // TODO(version): We are throwing. Not sure this is the right thing
+    packageDefinition
+      .as[Try[universe.v3.model.V2Package]]
+      .map { v2Package =>
+        rpc.v1.model.DescribeResponse(
+            `package` = v2Package.as[universe.v2.model.PackageDetails],
+            marathonMustache =
+              new String(universe.common.ByteBuffers
+                           .getBytes(v2Package.marathon.v2AppMustacheTemplate),
+                         Charsets.Utf8),
+            command = v2Package.command.as[Option[universe.v2.model.Command]],
+            config = v2Package.config,
+            resource =
+              v2Package.resource.as[Option[universe.v2.model.Resource]]
+        )
+      }
+      .get
   }
-
 }
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
@@ -1,86 +1,310 @@
 package com.mesosphere.cosmos.converter
 
-import com.mesosphere.universe.v2.{model => v2}
-import com.mesosphere.universe.v3.{model => v3}
+import com.mesosphere.cosmos.PackageFileMissing
+import com.mesosphere.cosmos.internal
+import com.mesosphere.universe
 import com.twitter.bijection.Conversion.asMethod
 import com.twitter.bijection.{Bijection, Injection}
-
+import java.nio.ByteBuffer
+import scala.util.Failure
 import scala.util.Try
 
 object Universe {
 
-  implicit val v3V3PackagingVersionToV2PackagingVersion: Injection[v3.V3PackagingVersion, v2.PackagingVersion] = {
-    val fwd = (x: v3.V3PackagingVersion) => v2.PackagingVersion(x.v)
-    val rev = (x: v2.PackagingVersion) => Try(v3.V3PackagingVersion(x.toString))
+  implicit val v2PackagingVersionToV3V2PackagingVersion: Bijection[
+      universe.v2.model.PackagingVersion,
+      universe.v3.model.V2PackagingVersion] = {
+
+    Bijection.build { (_: universe.v2.model.PackagingVersion) =>
+      universe.v3.model.V2PackagingVersion.instance
+    } { (x: universe.v3.model.V2PackagingVersion) =>
+      universe.v2.model.PackagingVersion(x.v)
+    }
+  }
+
+  implicit val v3V2PackageToV2PackageDetails: Injection[
+      universe.v3.model.V2Package, universe.v2.model.PackageDetails] = {
+    Injection.build { (value: universe.v3.model.V2Package) =>
+      universe.v2.model.PackageDetails(
+          packagingVersion =
+            value.packagingVersion.as[universe.v2.model.PackagingVersion],
+          name = value.name,
+          version = value.version.as[universe.v2.model.PackageDetailsVersion],
+          maintainer = value.maintainer,
+          description = value.description,
+          tags = value.tags.as[List[String]],
+          selected = value.selected,
+          scm = value.scm,
+          website = value.website,
+          framework = value.framework,
+          preInstallNotes = value.preInstallNotes,
+          postInstallNotes = value.postInstallNotes,
+          postUninstallNotes = value.postUninstallNotes,
+          licenses = value.licenses.as[Option[List[universe.v2.model.License]]]
+      )
+    } { _ =>
+      Try(throw new IllegalArgumentException(
+              "Cannot convert version 2 package details to version 3 package"))
+    }
+  }
+
+  implicit val v3V2PackageToInternalPackageDefinition: Injection[
+      universe.v3.model.V2Package, internal.model.PackageDefinition] = {
+    Injection.build { (value: universe.v3.model.V2Package) =>
+      internal.model.PackageDefinition(
+          value.packagingVersion,
+          value.name,
+          value.version,
+          value.releaseVersion,
+          value.maintainer,
+          value.description,
+          value.tags,
+          value.selected,
+          value.scm,
+          value.website,
+          value.framework,
+          value.preInstallNotes,
+          value.postInstallNotes,
+          value.postUninstallNotes,
+          value.licenses,
+          None,
+          Some(value.marathon),
+          value.resource.as[Option[universe.v3.model.V3Resource]],
+          value.config,
+          value.command
+      )
+    } { (value: internal.model.PackageDefinition) =>
+      // TODO(version): deal with minDcosReleaseVersion
+      for {
+        packagingVersion <- Try(value.packagingVersion.asInstanceOf[
+                                   universe.v3.model.V2PackagingVersion])
+        resource <- value.resource
+                     .as[Try[Option[universe.v3.model.V2Resource]]]
+        marathon <- Try(value.marathon.getOrElse(
+                           throw PackageFileMissing("package.json")))
+      } yield {
+        universe.v3.model.V2Package(
+            packagingVersion,
+            value.name,
+            value.version,
+            value.releaseVersion,
+            value.maintainer,
+            value.description,
+            marathon,
+            value.tags,
+            value.selected,
+            value.scm,
+            value.website,
+            value.framework,
+            value.preInstallNotes,
+            value.postInstallNotes,
+            value.postUninstallNotes,
+            value.licenses,
+            resource,
+            value.config,
+            value.command
+        )
+      }
+    }
+  }
+
+  implicit val v3V3PackageToInternalPackageDefinition: Injection[
+      universe.v3.model.V3Package, internal.model.PackageDefinition] = {
+    Injection.build { (value: universe.v3.model.V3Package) =>
+      internal.model.PackageDefinition(
+          value.packagingVersion,
+          value.name,
+          value.version,
+          value.releaseVersion,
+          value.maintainer,
+          value.description,
+          value.tags,
+          value.selected,
+          value.scm,
+          value.website,
+          value.framework,
+          value.preInstallNotes,
+          value.postInstallNotes,
+          value.postUninstallNotes,
+          value.licenses,
+          value.minDcosReleaseVersion,
+          value.marathon,
+          value.resource,
+          value.config,
+          value.command
+      )
+    } { (value: internal.model.PackageDefinition) =>
+      Try(value.packagingVersion
+            .asInstanceOf[universe.v3.model.V3PackagingVersion]).map {
+        packagingVersion =>
+          universe.v3.model.V3Package(
+              packagingVersion,
+              value.name,
+              value.version,
+              value.releaseVersion,
+              value.maintainer,
+              value.description,
+              value.tags,
+              value.selected,
+              value.scm,
+              value.website,
+              value.framework,
+              value.preInstallNotes,
+              value.postInstallNotes,
+              value.postUninstallNotes,
+              value.licenses,
+              value.minDcosReleaseVersion,
+              value.marathon,
+              value.resource,
+              value.config,
+              value.command
+          )
+      }
+    }
+  }
+
+  // TODO(version): Can we write this as an Bijection?
+  implicit val v3PackageDefinitionToInternalPackageDefinition: Injection[
+      universe.v3.model.PackageDefinition,
+      internal.model.PackageDefinition] = {
+    Injection.build { (value: universe.v3.model.PackageDefinition) =>
+      value match {
+        case v2Package: universe.v3.model.V2Package =>
+          v2Package.as[internal.model.PackageDefinition]
+        case v3Package: universe.v3.model.V3Package =>
+          v3Package.as[internal.model.PackageDefinition]
+      }
+    } { (value: internal.model.PackageDefinition) =>
+      Try(throw new IllegalArgumentException("TODO(version): FIX ME???"))
+    }
+  }
+
+  implicit val v2PackageDetailsVersionToV3PackageDefinitionVersion: Bijection[
+      universe.v2.model.PackageDetailsVersion,
+      universe.v3.model.PackageDefinition.Version] = {
+    Bijection.build { (x: universe.v2.model.PackageDetailsVersion) =>
+      universe.v3.model.PackageDefinition.Version(x.toString)
+    } { (x: universe.v3.model.PackageDefinition.Version) =>
+      universe.v2.model.PackageDetailsVersion(x.toString)
+    }
+  }
+
+  // TODO(version): Make this a Bijection and reverse
+  implicit val v3PackageDefinitionTagToString: Injection[
+      universe.v3.model.PackageDefinition.Tag, String] = {
+
+    val fwd = (x: universe.v3.model.PackageDefinition.Tag) => x.value
+    val rev = (x: String) => Try(universe.v3.model.PackageDefinition.Tag(x))
 
     Injection.build(fwd)(rev)
   }
 
-  implicit val v3PackageDefinitionVersionToV2PackageDetailsVersion: Bijection[v3.PackageDefinition.Version, v2.PackageDetailsVersion] = {
-    val fwd = (x: v3.PackageDefinition.Version) => v2.PackageDetailsVersion(x.toString)
-    val rev = (x: v2.PackageDetailsVersion) => v3.PackageDefinition.Version(x.toString)
+  // TODO(version): Make this a Bijection and reverse
+  implicit val v3LicenseToV2License: Injection[
+      universe.v3.model.License, universe.v2.model.License] = {
 
-    Bijection.build(fwd)(rev)
-  }
-
-  implicit val v3PackageDefinitionTagToString: Injection[v3.PackageDefinition.Tag, String] = {
-    val fwd = (x: v3.PackageDefinition.Tag) => x.value
-    val rev = (x: String) => Try(v3.PackageDefinition.Tag(x))
-
-    Injection.build(fwd)(rev)
-  }
-
-  implicit val v3LicenseTov2License: Injection[v3.License, v2.License] = {
-    val fwd = (x: v3.License) => v2.License(name = x.name, url = x.url.toString)
-    val rev = (x: v2.License) =>
-      Common.uriToString.invert(x.url).map(url => v3.License(name = x.name, url = url))
-
-    Injection.build(fwd)(rev)
-  }
-
-  implicit val v3CommandToV2Command: Bijection[v3.Command, v2.Command] = {
-    val fwd = (x: v3.Command) => v2.Command(pip = x.pip)
-    val rev = (x: v2.Command) => v3.Command(pip = x.pip)
-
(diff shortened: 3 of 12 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+  ): universe.v3.model.Repository = {
+    val bundle = new ZipInputStream(inputStream)
+    // getNextEntry() returns null when there are no more entries
+    val universeRepository = Iterator
+      .continually(Option(bundle.getNextEntry()))
+      .takeWhile(_.isDefined)
+      .flatten
+      .filter(!_.isDirectory)
+      .map(entry => Paths.get(entry.getName))
+      .filter(entryPath => entryPath.getNameCount == 7) // Only keep leave nodes
```
jsancio: Fix comment: "Only keep leaf nodes".

cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v2/circe/MediaTypedEncoders.scala
```diff
@@ -4,18 +4,22 @@ import com.mesosphere.cosmos.circe.{DispatchingMediaTypedEncoder, MediaTypedEnco
 import com.mesosphere.cosmos.converter
 import com.mesosphere.cosmos.http.MediaTypes
 import com.mesosphere.cosmos.rpc
+import com.mesosphere.cosmos.internal
 import com.mesosphere.universe
 
 object MediaTypedEncoders {
 
-  implicit val packageDescribeEncoder: DispatchingMediaTypedEncoder[rpc.v2.model.DescribeResponse] = {
+  implicit val packageDescribeEncoder: DispatchingMediaTypedEncoder[internal.model.PackageDefinition] = {
```
jsancio: We agree to make these `rpc.v2.model.*Response` but I am not doing it here.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+          logger.info("Parsing entry: {}", entryPath)
+          val packageName = entryPath.getName(4).toString
+          val releaseVersion = Integer.parseInt(entryPath.getName(5).toString)
+
+          val oldV2Package = state.getOrElse(
+              (packageName, releaseVersion),
+              V2PackageInformation()
           )
+
+          val buffer = StreamIO.buffer(bundle).toByteArray()
```
cruhland: We should say in a comment that bundle is being mutated.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+              oldV2Package
+            }
+
+          state + (((packageName, releaseVersion), newV2Package))
+      }
+
+    universe.v3.model.Repository(
+        universeRepository.toList.flatMap {
+          case ((packageName, releaseVersion), packageInfo) =>
+            packageInfo.packageDetails.map {
```
cruhland: To be consistent with `UniversePackageCache`, we should throw an error if this file is missing.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+              details =>
+                // This skips packages that don't contain package.json
+                val marathon = packageInfo.marathonMustache.getOrElse(
```
cruhland: To be consistent with `UniversePackageCache`, we should throw an error if this file is missing.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+                    details.website,
+                    details.framework,
+                    details.preInstallNotes,
+                    details.postInstallNotes,
+                    details.postUninstallNotes,
+                    details.licenses.map {
+                      licenses =>
+                        licenses.flatMap {
+                          license =>
+                            v3LicenseToV2License.invert(license) match {
```
cruhland: See if we can change `V2Package` to be less strict when it comes to licenses: we want to support any license values that would have been considered valid with the Universe 2.x model.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+          val releaseVersion = Integer.parseInt(entryPath.getName(5).toString)
+
+          val oldV2Package = state.getOrElse(
+              (packageName, releaseVersion),
+              V2PackageInformation()
           )
+
+          val buffer = StreamIO.buffer(bundle).toByteArray()
+
+          val newV2Package =
```
BenWhitehead: Please add clarifying curly braces around all the if branches

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+    // getNextEntry() returns null when there are no more entries
+    val universeRepository = Iterator
+      .continually(Option(bundle.getNextEntry()))
+      .takeWhile(_.isDefined)
+      .flatten
+      .filter(!_.isDirectory)
+      .map(entry => Paths.get(entry.getName))
+      .filter(entryPath => entryPath.getNameCount == 7) // Only keep leave nodes
+      .foldLeft(Map.empty[(String, Int), V2PackageInformation]) {
+        (state, entryPath) =>
```
BenWhitehead: can we factor the body of this `foldLeft` into it's own method?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+            } else if (entryPath.getFileName() == Paths.get("command.json")) {
+              oldV2Package.copy(
+                  command = Some(
+                      parseAndVerify[universe.v2.model.Command](
+                          entryPath, new String(buffer))
+                  )
+              )
+            } else {
+              // We have an extra file for some reason. Just ignore it.
+              logger.warn(
```
BenWhitehead: Please remove log message. If it's important for the user's operation it should be returned in the response.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+    val universeRepository = Iterator
+      .continually(Option(bundle.getNextEntry()))
+      .takeWhile(_.isDefined)
+      .flatten
+      .filter(!_.isDirectory)
+      .map(entry => Paths.get(entry.getName))
+      .filter(entryPath => entryPath.getNameCount == 7) // Only keep leave nodes
+      .foldLeft(Map.empty[(String, Int), V2PackageInformation]) {
+        (state, entryPath) =>
+          logger.info("Parsing entry: {}", entryPath)
```
BenWhitehead: Please remove log statement.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+                    details.postUninstallNotes,
+                    details.licenses.map {
+                      licenses =>
+                        licenses.flatMap {
+                          license =>
+                            v3LicenseToV2License.invert(license) match {
+                              case Success(l) => Some(l)
+                              case Failure(err) =>
+                                // Drop unconvertable licenses
+                                logger.warn(
```
BenWhitehead: Please remove log message. If it's important for the user's operation it should be returned in the response.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -1,43 +1,69 @@
...
+    new FnUniverseClient(function)
+  }
 }
 
-class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver: StatsReceiver = NullStatsReceiver) {
+class UniverseRepositoryFetcher(
+    implicit statsReceiver: StatsReceiver = NullStatsReceiver) {
 
   private[this] val stats = statsReceiver.scope("repositoryFetcher")
+  private[this] val logger = LoggerFactory.getLogger(getClass)
```
BenWhitehead: Please remove logger after removing log statements.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
     Option(props.getProperty("cosmos.version")).getOrElse("unknown-version")
   }
 
-  def apply(uri: Uri)(implicit session: RequestSession): Future[CosmosInternalRepository] = {
-    adminRouter.getDcosVersion().flatMap { dcosVersion =>
```
BenWhitehead: What code is calling `adminRouter.getDcosVersion` now that this method is removed?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -49,93 +75,233 @@ class UniverseRepositoryFetcher(adminRouter: AdminRouter)(implicit statsReceiver
...
+      }
+
+    universe.v3.model.Repository(
+        universeRepository.toList.flatMap {
+          case ((packageName, releaseVersion), packageInfo) =>
+            packageInfo.packageDetails.map {
+              details =>
+                // This skips packages that don't contain package.json
+                val marathon = packageInfo.marathonMustache.getOrElse(
+                    ByteBuffer.allocate(0))
```
BenWhitehead: This this `ByteBuffer.allocate(0)` needs to stick around (not sure it does) create a constant in `ByteBuffers` of this so that we're not allocating a new ByteBuffer every time we fetch.

cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala
```diff
@@ -67,7 +67,7 @@ object Decoders {
   implicit val decodeV2PackagingVersion: Decoder[V2PackagingVersion] = {
     Decoder.decodeString.map(V2PackagingVersion(_))
   }
-  implicit val decodeV2Resource: Decoder[V2Resource] = deriveFor[V2Resource].decoder
+  implicit val decodeV3V2Resource: Decoder[V2Resource] = deriveFor[V2Resource].decoder
```
cruhland: For consistency should all of these decoders be renamed in the same way as the Universe v2 decoders?

BenWhitehead: my preference would be to not rename any of them, and instead be more explicit with scoping the use of these. it can be hard to keep track of everything when we have many many place in the code that are v1, v2 and v3.  I think we should try our best to limit the number of locations we have overlapping versions of serialization/deserialization.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
```diff
@@ -1,48 +1,34 @@
...
-      command = v3Package.command.as[Option[v2.Command]],
-      config = v3Package.config,
-      resource = v3Package.resource.map(v3V3ResourceToV2Resource)
-    )
+  def packageDefinitionToDescribeResponse(
+      packageDefinition: internal.model.PackageDefinition)
+    : DescribeResponse = {
+    // TODO(version): We are throwing. Not sure this is the right thing
+    packageDefinition
+      .as[Try[universe.v3.model.V2Package]]
```
cruhland: This is because the `PackageDefinition` might have more information in it than a `V2Package` does, correct? Such as the `minDcosReleaseVersion`, or `Resources.cli`.

The right answer here depends on whether packages published to Universe pre-1.8 are being kept in version-2.x format or are being converted to version-3.x format. If it's the former, then we're pretty safe in throwing here because old clients will still be able to see the packages they care about. If it's the latter, then we really shouldn't throw because then old clients would not be able to see any packages.

However, should we even care whether the packages are stored in version-2.x or version-3.x format? It feels like mixing of concerns. The client chooses the format they want via the `Accept` header; if the `v1` response omits some information, they can just request a `v2` response instead.

Maybe it's better to think about it as being either pro-backwards-compatibility for older clients (dropping information, not throwing) vs. pro-upgrade-pressure for older clients (throwing, not dropping information).

BenWhitehead: I'm also curious where the throw comes from. I don't think we can create a bijection between v2 and v3 models since it's not possible to do a round trip between them since there's inherent data loss (in our current scenario) going from v3 to v2.

I think it is possible to say "v2 -> v3 is good", "v3 -> v2 has a projection", but "v3 -> v2 -> v3 doesn't necessarily work" because info present in v3 can be lost going to v3.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
```diff
@@ -1,86 +1,308 @@
...
-  implicit val v3V3PackagingVersionToV2PackagingVersion: Injection[v3.V3PackagingVersion, v2.PackagingVersion] = {
-    val fwd = (x: v3.V3PackagingVersion) => v2.PackagingVersion(x.v)
-    val rev = (x: v2.PackagingVersion) => Try(v3.V3PackagingVersion(x.toString))
+  implicit val v2PackagingVersionToV3V2PackagingVersion: Bijection[
+      universe.v2.model.PackagingVersion,
+      universe.v3.model.V2PackagingVersion] = {
 
-    Injection.build(fwd)(rev)
+    Bijection.build { (x: universe.v2.model.PackagingVersion) =>
+      universe.v3.model.V2PackagingVersion(x.toString)
```
cruhland: Do we ever expect `PackagingVersion` to have a value other than `2.0`? If so, this line can throw an `AssertionError`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
```diff
@@ -1,86 +1,308 @@
...
+          value.postUninstallNotes,
+          value.licenses,
+          None,
+          Some(value.marathon),
+          value.resource.as[Option[universe.v3.model.V3Resource]],
+          value.config,
+          value.command
+      )
+    } { (value: internal.model.PackageDefinition) =>
+      // TODO(version): deal with minDcosReleaseVersion
```
cruhland: Where is this direction of the conversion used? In that context, if there's a distinction between a `minDcosReleaseVersion` that's `None` vs. one that's `Some`, then we can't safely do this conversion.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
```diff
@@ -1,86 +1,308 @@
...
+      } yield {
+        universe.v3.model.V2Package(
+            packagingVersion,
+            value.name,
+            value.version,
+            value.releaseVersion,
+            value.maintainer,
+            value.description,
+            value.marathon.getOrElse(universe.v3.model
+                  .Marathon(ByteBuffer.allocate(0))), // TODO(version): Is this correct?
```
cruhland: This is similar to the situation in `UniverseClient`, no? In v2 packages, the `marathon.json.mustache` file is required, so if one doesn't exist it's an error. So we either need to throw an error here or make `V2Package` more lenient with an `Option[Marathon]`.

cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala
jsancio: One of the places where we use them is in the code that goes from an internal model to both `rpc.v1.model` or `rpc.v2.model` where each of those classes get encoded with `universe.v2.circe.Encoders` and `universe.v1.circe.Decoders` respectively. To get around this we would need explicit encoders like we do in `DispatchingMediaTypedEncoder`.

Another solution to this problem is to make the encoders and decoders implicit in the companion objects. In that case Scala will pull in that implicit scope automatically and won't have the symbol shadowing we are seeing.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
jsancio: First, you can never go `V3Package => V2Package` or vice versa. This is basically because `V3Package` is statically bound to `packagingVersion=3.0` while `V2Package` is statically bound to `packagingVersion=2.0`.

The reason for `internal.model.PackageDefinition => Try[V2Package]` is because `internal.model.PackageDefinition` is conceptually the union of both `V2Package` and `V3Package` so it is only convertible to `V2Package` when it represents that union.

Now, the reason for this union is so that internal Cosmos code doesn't have to do pattern matching all over the code and instead handle presence or absence fields systematically. E.g. missing `minDcosReleaseVersion` or missing `marathonMustache`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/internal/model/PackageDefinition.scala
```diff
@@ -0,0 +1,31 @@
...
+import io.circe.JsonObject
+
+case class PackageDefinition(
+    packagingVersion: universe.v3.model.PackagingVersion,
+    name: String,
+    version: universe.v3.model.PackageDefinition.Version,
+    releaseVersion: universe.v3.model.PackageDefinition.ReleaseVersion,
+    maintainer: String,
+    description: String,
+    tags: List[universe.v3.model.PackageDefinition.Tag] = Nil,
```
cruhland: This has been mentioned elsewhere, but several of these types are stricter for `universe.v3` than for `v2`, so mapping from `universe.v2` into this type will not work. The tags are one example: the `PackageDefinition.Tag` constructor throws an `AssertionError` if the `String` value of the tag doesn't match a regex.

Other examples are `PackageDefinition.ReleaseVersion` and `License` (the `url` field is a typed `Uri`).

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
jsancio: I can change this to `universe.v3.model.V2PackagingVersion.instance`. Does that work for you?

cruhland: Sure, since that indicates that the value of the input `x` doesn't matter (it will always be the same).

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
jsancio: Yeah. I haven't thought about it but we can return a `Failure` if `minDcosReleaseVersion.map(_  > 1.6.1).getOrElse(false)`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
jsancio: Right. I'll update the comment to say that this should be a `Try` and add to the `flatMap`s above.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
cruhland: That came up in the review meeting earlier. Jose pointed out that the body uses the `bundle` variable, which has side effects and thus it might be better to keep it within a single scope.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v2/circe/MediaTypedEncoders.scala
cruhland: Jose and I had a followup conversation. In the long term, these type parameters should all be set to types from the internal model (not necessarily containing `Response` in their names, although that might make it easier to understand what they are intended for).

This is because we want to take the result of the "business logic" in an `EndpointHandler`, which will be an `internal.model` type, and encode it to JSON. The encoding process will first map the internal model to some versioned response model from the `rpc` package, which will then be Circe-encoded to JSON.

Currently some of these parameters might be set to an `rpc.v2` type directly, just to avoid creating a duplicate `internal.model` type.

cosmos-server/src/main/scala/com/mesosphere/cosmos/internal/model/PackageDefinition.scala
jsancio: @tamarrow @BenWhitehead Do test for this regex match in our Universe CI? If not, can we add an issue for it?

jsancio: @cruhland We have `Injection[universe.v3.model.PackageDefinition.Tag, String]` so I assume that I deal with this somewhere or the types won't match 😄 .

tamarrow: no checks afaik. added: https://github.com/mesosphere/universe-checker/issues/2

cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v2/circe/MediaTypedEncoders.scala
```diff
@@ -4,18 +4,22 @@ import com.mesosphere.cosmos.circe.{DispatchingMediaTypedEncoder, MediaTypedEnco
...
     DispatchingMediaTypedEncoder(Seq(
       MediaTypedEncoder(
-        encoder = universe.v3.circe.Encoders.encodeV3Package,
+        encoder = universe.v3.circe.Encoders.encodeV3Package.contramap { (pkgDefinition: internal.model.PackageDefinition) =>
+          // TODO(version): This throws. Need to figure out how we want to handle this errors
+          converter.Universe.v3V3PackageToInternalPackageDefinition.invert(pkgDefinition).get
+        },
```
cruhland: Why does this throw? Shouldn't any `PackageDefinition` be convertible to a `V3Package`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/internal/model/PackageDefinition.scala
jsancio: I wrote a `TODO(version)` to revisit this in `UniverseClient.scala`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v2/circe/MediaTypedEncoders.scala
cruhland: Is it just the `packagingVersion` that causes trouble?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: I don't know yet. Maybe something in `UniversePackageCache` or maybe `Cosmos`. Let's talk about it as the next set of changes.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: I am not sure I follow. All of the if branches have curly braces.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: sorry, I meant something like this:

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Yes. Unfortunately this closure closes over "global" state in the `bundle` variable.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: Fair enough

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Removed the `ByteBuffer`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: I have a `TODO(version)` comment to investigate this in the universe converter file.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v2/circe/MediaTypedEncoders.scala
jsancio: Yeah. This is because of `_.packagingVersion`: https://github.com/dcos/cosmos/pull/373/files#diff-ab09d261938f01cee38c97067021aeffR136

jsancio:
Thanks @BenWhitehead and @cruhland for the review. I am done with my changes for now. It should be good to merge if we are happy.

cruhland:
@jsancio LGTM. The remaining issues have TODO comments so we can handle them later.

PR #375, opened 2016-06-14
universe v3 equivalent of PackageSearchHandler

tamarrow:
#370 

We probably want to update the SearchResult to include new info (namely resource.cli) but I'll leave that for another pr.


cruhland:
LGTM.

PR #377, opened 2016-06-14
Create stubs for Universe v3 equivalents of Cosmos and MultiRepository

cruhland:


cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
```diff
@@ -113,3 +113,16 @@ final class MultiRepository (
...
+// TODO (version): Delete this
+final class V3MultiRepository(
+  packageRepositoryStorage: PackageSourcesStorage,
+  universeClient: UniverseClient
+) extends V3PackageCollection {
+
+  override def getPackageByPackageVersion(
+    packageName: String,
+    packageVersion: Option[universe.v3.model.PackageDefinition.Version]
+  ): Future[(universe.v3.model.V3Package, Uri)] = ???
```
jsancio: Okay. After we merge the unverser v2 to universe v3 PR we should change this to `internal.model.PackageDefinition`.

jsancio:
Is this PR ready for review? Correct me if I am wrong but this PR doesn't do what the title says, no?

cruhland:
@jsancio I updated the title to be more accurate. The intention is to reduce the number of surprises when it comes time to integrate all our changes.

cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
jsancio: This should return `Future[(internal.model.PackageDefinition, Uri)]`.

jsancio:
LGTM after we address one small comment!

cruhland:
@jsancio updated. Changing the type required some of the endpoint handlers to change, so you should take another look. Changes are in a separate commit.

jsancio:
LGTM. I see that we are using encoders explicitly in a few places because of the `universe.v2` and `universe.v3` conflict. Type classes lose some of their appeal if we have to use them explicitly.

Issue #371, opened 2016-06-14
Create a Universe 3.0 equivalent for ListHandler


cruhland:
Name the class `V3ListHandler`, and put it in the same file as the original, with the comment `// TODO (version): Rename to ListHandler`.

The class must use `V3PackageCollection` and `V3CosmosRepository` (doesn't exist yet) instead of their Universe 2.0 equivalents.

PR #380, opened 2016-06-15
universe v3 equivalent of ListHandler

tamarrow:
#371

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
@@ -38,5 +38,31 @@ object Response {
         )
       }
       .get
+    }
+
+   private[this] def getV2ResourceFromPackageDefinition(
+     packageDef: universe.v3.model.PackageDefinition
+   ): Option[universe.v2.model.Resource] = {
+      packageDef match {
+        case v2Package: universe.v3.model.V2Package =>
+          v2Package.resource.as[Option[universe.v2.model.Resource]]
+        case v3Package: universe.v3.model.V3Package =>
+          v3Package.resource.as[Option[universe.v2.model.Resource]]
+    }
   }
+
+  def v2ListResponseToV1ListResponse(x: rpc.v2.model.ListResponse): rpc.v1.model.ListResponse = {
+    rpc.v1.model.ListResponse(
+      packages = x.packages.map { y =>
+        rpc.v1.model.Installation(
+          appId = y.appId,
+          packageInformation = rpc.v1.model.InstalledPackageInformation(
+              packageDefinition = y.packageInformation.as[universe.v2.model.PackageDetails],
+              resourceDefinition = getV2ResourceFromPackageDefinition(y.packageInformation)
+          )
+        )
+      }
+    )
+  }
+
 }
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
@@ -22,6 +22,57 @@ object Universe {
     }
   }
 
+  implicit val v3V2PackageDetailsToPackageDefinition: Bijection[
+      universe.v2.model.PackagingVersion,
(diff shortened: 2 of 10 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
```diff
@@ -22,6 +22,71 @@ object Universe {
...
+      universe.v3.model.V3PackagingVersion] = {
+
+    Bijection.build { (_: universe.v2.model.PackagingVersion) =>
+      universe.v3.model.V3PackagingVersion.instance
+    } { (x: universe.v3.model.V3PackagingVersion) =>
+      universe.v2.model.PackagingVersion(x.v)
+    }
+  }
+
+  implicit val v3PackagingVersionToV2PackagingVersion: Injection[
```
cruhland: The name of this and the one above should be swapped.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
```diff
@@ -38,5 +38,20 @@ object Response {
...
+    }
+
+  def v2ListResponseToV1ListResponse(x: rpc.v2.model.ListResponse): rpc.v1.model.ListResponse = {
+    rpc.v1.model.ListResponse(
+      packages = x.packages.map { y =>
+        rpc.v1.model.Installation(
+          appId = y.appId,
+          packageInformation = rpc.v1.model.InstalledPackageInformation(
+              packageDefinition = y.packageInformation.as[universe.v2.model.PackageDetails],
+              resourceDefinition = y.packageInformation.as[Option[universe.v2.model.Resource]]
```
cruhland: This usage of `as` seems to go against the intended use of the Bijection library. Here, we're taking a `universe.v3.model.PackageDefinition` and "converting" it to a `universe.v2.model.Resource`. But there's not a valid bijection or injection between the two types. Instead, we probably should do something like this:

```scala
resourceDefinition = y.packageInformation.resource.as[Option[universe.v2.model.Resource]]
```

That would use the `Bijection/Injection` that already exists for `Resource`.

cruhland: Ah, part of the problem is that `packageInformation.resource` doesn't exist, we'd have to pattern match. That's still better handled outside of `Bijection/Injection` IMO.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/CosmosRepository.scala
```diff
@@ -14,3 +17,16 @@ trait CosmosRepository extends PackageCollection {
...
+// TODO (version): Rename to CosmosRepository
+/** A repository of packages that can be installed on DCOS. */
+trait V3CosmosRepository extends PackageCollection {
+
+  def repository: PackageRepository
+
+  def getPackageByReleaseVersion(
+    packageName: String,
+    releaseVersion: ReleaseVersion
+  ): Future[(V3Package, Uri)]
```
cruhland: We don't need to include the `Uri`, right?

cruhland:
@tamarrow just a few comments, good to merge once they are addressed. Don't worry too much about the `Bijection` stuff, we'll probably clean some of that up in a separate PR.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/CosmosRepository.scala
tamarrow: removed

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
tamarrow: done

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
tamarrow: updated

tamarrow:
thanks @cruhland

PR #381, opened 2016-06-15
Update v3-repo-schema.json to ensure resource documents set additionalProperties false

BenWhitehead:
Update based on https://github.com/mesosphere/universe/pull/491


cruhland:
LGTM. Do we need to update any of the v2 schema files? And why isn't there a `v3-resource-schema.json` file in Cosmos?

BenWhitehead:
@cruhland none of our models need to be updated (since circe drops additional properties at decode time).

`v3-resource-schema.json` is used to validate the resource file of a package with `packagingVersion` of `3.0`, cosmos only cares about the schema of the consumption document.

PR #382, opened 2016-06-16
Refactor the universe v3 code to use new client

jsancio:


cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -89,9 +110,6 @@ class UniverseRepositoryFetcher(
             "User-Agent",
             s"cosmos/$cosmosVersion dcos/${dcosReleaseVersion.show}"
         )
-        session.authorization.foreach { auth =>
-          conn.setRequestProperty("Authorization", auth.headerValue)
-        }
```
cruhland: Why was this deleted?

cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
```diff
@@ -116,19 +122,20 @@ final class MultiRepository (
...
   override def getPackagesByPackageName(
-    packageName: String
+      packageName: String
   ): Future[List[internal.model.PackageDefinition]] = ???
 
   override def getPackageByPackageVersion(
-    packageName: String,
-    packageVersion: Option[universe.v3.model.PackageDefinition.Version]
+      packageName: String,
+      packageVersion: Option[universe.v3.model.PackageDefinition.Version]
```
cruhland: Some extra whitespace was added here.

cruhland:
LGTM after comments are addressed.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: this has to stick around so that we can communicate with admin router when it's auth is enabled.

jsancio: I think this class should only talk to Universe and should not care about DC/OS authorization. There are a few benefits to this but the main benefit is that we will never accidentally leak authorization tokens to external Universes. What do we think?

cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
jsancio: I see. This is my autoformatter.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: The auth token is only needed for admin router, not universe you are correct. I should have expanded the diff collapse to see this is for universe fetch step not the previous request to admin router to get the dcos version.

PR #385, opened 2016-06-16
Introduce coursier for dependency resolution

BenWhitehead:
This PR introduces coursier[1] to the project for dependency resolution rather than the default ivy resolver included with sbt. It is significantly faster at resolving and downloading the dependencies of the project. Below are the runtimes for the project with and without coursier.

With the project already compiled, and an empty cache for ivy (`rm -rf ~/.ivy2/cache`):
```bash
time sbt compile it:compile
<snip>

real    5m13.391s
user    4m35.728s
sys     0m3.244s
```

With the project already compiled, and an empty coursier cache (`rm -rf ~/.coursier/cache`):
```bash
time sbt compile it:compile
<snip>

real    1m27.734s
user    4m51.496s
sys     0m2.680s
```

I've also updated the set of resolvers to ensure that Maven Central has the highest priority, this will hopefully help reduce the number of build failures due to 503's from maven.twttr.com.

[1] https://github.com/alexarchambault/coursier

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -113,11 +113,31 @@ object CosmosBuild extends Build {
 
     exportJars := true,
 
-    resolvers ++= Seq(
-      "Clojars Repository" at "http://clojars.org/repo",
-      "Conjars Repository" at "http://conjars.org/repo",
-      "Twitter Maven" at "http://maven.twttr.com",
-      "finch-server" at "http://storage.googleapis.com/benwhitehead_me/maven/public"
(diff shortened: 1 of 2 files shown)


BenWhitehead:
@cruhland please review

cruhland:
:ship: it!

BenWhitehead:
Introduces some issue that results in test failures. closing.

BenWhitehead:
Depends on #389

PR #388, opened 2016-06-16
Add cosmos-json as an aggregate to sbt config

BenWhitehead:


cruhland:
@BenWhitehead I was suspicious about how that assertion could be failing, so I did some investigation. I think what's actually been happening is that the `cosmos-json` project's tests were not being run by `sbt test`. On master, try running these commands at the SBT console:

```scala
> test
> cosmos-json/test
```

The first one should have "all" tests pass while the second will have two test failures. The failure occurred in #385 for the first time because of [this change](https://github.com/dcos/cosmos/pull/385/files#diff-798d1bfac5c6765ba2f0c34e37c073edR213).

This probably means that those test failures are legit and that we can add the `sbt-coursier` plugin without problems once they are fixed.

Issue #374, opened 2016-06-14
Wire together all Universe 3.0 classes and remove their old counterparts


cruhland:
This is the big one.

PR #391, opened 2016-06-17
Fix conversion failure when encoding Marathon package metadata label

cruhland:
Contributes to #374.

### DIFF
--- a/cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageInstallIntegrationSpec.scala
+++ b/cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageInstallIntegrationSpec.scala
@@ -237,22 +237,52 @@ private object PackageInstallIntegrationSpec extends Matchers with TableDrivenPr
     ("cassandra", PackageDetailsVersion("foobar"))
   )
 
+  private val HelloWorldCommand: Json = Map(
+    "pip" -> List(
+      "dcos<1.0".asJson,
+      "git+https://github.com/mesosphere/dcos-helloworld.git#dcos-helloworld=0.1.0".asJson
+    ).asJson
+  ).asJson
+
   private val HelloWorldLabels = StandardLabels(
-    Map(
-      ("DCOS_PACKAGE_METADATA", "eyJ3ZWJzaXRlIjoiaHR0cHM6Ly9naXRodWIuY29tL21lc29zcGhlcmUvZGNvcy1" +
-        "oZWxsb3dvcmxkIiwibmFtZSI6ImhlbGxvd29ybGQiLCJwb3N0SW5zdGFsbE5vdGVzIjoiQSBzYW1wbGUgcG9zdC" +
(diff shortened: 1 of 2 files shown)


cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageInstallIntegrationSpec.scala
```diff
@@ -237,22 +237,52 @@ private object PackageInstallIntegrationSpec extends Matchers with TableDrivenPr
...
-        "sInByZUluc3RhbGxOb3RlcyI6IkEgc2FtcGxlIHByZS1pbnN0YWxsYXRpb24gbWVzc2FnZSJ9"),
-      ("DCOS_PACKAGE_COMMAND", "eyJwaXAiOlsiZGNvczwxLjAiLCJnaXQraHR0cHM6Ly9naXRodWIuY29tL21lc29z" +
-        "cGhlcmUvZGNvcy1oZWxsb3dvcmxkLmdpdCNkY29zLWhlbGxvd29ybGQ9MC4xLjAiXX0="),
-      "DCOS_PACKAGE_REGISTRY_VERSION" -> "2.0",
-      "DCOS_PACKAGE_NAME" -> "helloworld",
-      "DCOS_PACKAGE_VERSION" -> "0.1.0",
-      "DCOS_PACKAGE_SOURCE" -> DefaultRepositories().getOrThrow(1).uri.toString,
-      "DCOS_PACKAGE_RELEASE" -> "0"
-    )
+    packageMetadata = Map(
```
cruhland: I had to update this because we're now putting the full `PackageDefinition` into the package metadata label. To make it easier to update in the future I replaced the base-64-encoded label text with its decoded JSON definition.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageInstallHandler.scala
```diff
@@ -87,9 +88,9 @@ object PackageInstallHandler {
 
       // TODO(version): This can throw
       override def packageMetadataJson: Json = pkg
-        .as[Try[universe.v3.model.V3Package]]
+        .as[Try[universe.v3.model.PackageDefinition]]
```
cruhland: This accommodates `V2Package` and `V3Package`.

jsancio:
LGTM! 🚢 it

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageInstallIntegrationSpec.scala
BenWhitehead: I don't think we want to do this. having json documents in json strings is a bad thing. I also don't think we want to encode the entire package into this label. between the config.json and marathon template this can easily get to very large sizes. If these documents are large it will place a lot of stress on marathon due to the zk 1mb node size limit.

PR #392, opened 2016-06-17
bugfix for getPackagesByPackageName

tamarrow:
This fixes the failing listVersions test in `PackageDescribeSpec`

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
@@ -30,7 +30,9 @@ final class MultiRepository(
         }
       } map (_.flatten)
     } map { packages =>
-      packages.headOption.getOrElse(throw PackageNotFound(packageName))
+      packages
+        .find(!_.isEmpty)
+        .getOrElse(throw PackageNotFound(packageName))
     }
   }
 


cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
```diff
@@ -30,7 +30,9 @@ final class MultiRepository(
         }
       } map (_.flatten)
     } map { packages =>
-      packages.headOption.getOrElse(throw PackageNotFound(packageName))
+      packages
+        .filterNot(_.isEmpty)
+        .headOption.getOrElse(throw PackageNotFound(packageName))
```
jsancio: How about `.find(!_.isEmpty).getOrElse(...)`?

tamarrow: done

jsancio:
LGTM! 🚢 it.

cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
```diff
@@ -30,7 +30,9 @@ final class MultiRepository(
         }
       } map (_.flatten)
     } map { packages =>
-      packages.headOption.getOrElse(throw PackageNotFound(packageName))
+      packages
+        .find(!_.isEmpty)
+        .getOrElse(throw PackageNotFound(packageName))
     }
```
cruhland: I'm wondering if we should have this code instead?

```scala
repositories().flatMap { repositories =>
  Future.collect {
    // ...
  } map (_.flatten.flatten.toList)
}
```

cruhland: I verified locally that this change makes `PackageDescribeSpec` pass.

jsancio: Well, this is not how the old Cosmos works. I am okay changing it but let's not do it in this PR and let's talk about the consequences before doing it.

tamarrow: Merging into @cruhland's branch as is, and we can always change it later.

PR #393, opened 2016-06-17
Fix code causing failure of PackageSearchSpec

BenWhitehead:


BenWhitehead:
@cruhland @jsancio please review

jsancio:
LGTM! 🚢 it.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -232,7 +232,7 @@ final class DefaultUniverseClient(
                      */
                     universe.v3.model.PackageDefinition.Tag(tag)
                 },
-                details.selected,
+                details.selected.orElse(Some(false)),
```
cruhland: This suggests to me that the `selected` field on `internal.model.PackageDefinition` should have type `Boolean` instead of `Option[Boolean]`, because a package is either selected or not. The `Option` here is required only for JSON encoding/decoding.

BenWhitehead: The model objects we've got in cosmos reflect the schema as it's defined for universe, which in the case of selected means an optional field. When we added it to version 2 models we set it to default to false when read so we continue to do that here. But it's only from the standpoint of cosmos, not from the schema.

cruhland: Yeah, I'm fine with `selected` being optional in the Universe schema. I'll ask a more specific question: why does this fix the test failure, and should the fix be made here? If we always translate `universe.v3.model.V2Package` and `universe.v3.model.V3Package` into `internal.model.PackageDefinition`, then the default value of `internal.model.PackageDefinition.selected` should be determined during that translation, not here.

Issue #396, opened 2016-06-20
Replace GitHub URLs in PackageRepositorySpec


cruhland:
In the "Issue #219" test we are using two GitHub URLs to test error cases. These should be replaced with URLs of resources that we control.

jsancio:
This has been fixed.

PR #398, opened 2016-06-20
Fix test failures in PackageRepositorySpec

cruhland:
Contributes to #374.

### DIFF
--- a/cosmos-server/src/it/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
+++ b/cosmos-server/src/it/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
@@ -145,7 +145,7 @@ final class PackageRepositorySpec
       assertAddFailure(unsupportedUriMsg(uri), bogusRepository)
     }
     def unsupportedUriMsg(uri: Uri): String = {
-      s"Repository URI [${uri}] uses an unsupported scheme. Only http and https are supported"
+      s"Repository URI [$uri] uses an unsupported scheme. Only http and https are supported"
     }
   }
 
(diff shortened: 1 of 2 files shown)


cosmos-server/src/it/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
```diff
@@ -171,18 +171,22 @@ final class PackageRepositorySpec
...
-      assertInvalidRepo("https://github.com/mesosphere/dcos-cli/archive/master.zip")
+      val uriText = "https://github.com/mesosphere/dcos-cli/archive/master.zip"
+      val expectedMsg = s"Index file missing for repo [$uriText]"
+      assertInvalidRepo(uriText, expectedMsg)
     }
 
     "non-Zip-encoded repo bundle" in {
       // TODO: Use a more reliable URI
-      assertInvalidRepo("https://github.com/mesosphere/dcos-cli")
+      val uriText = "https://mesosphere.com/"
```
cruhland: @BenWhitehead suggestions for an alternative URL? Issue #396 tracks the full goal.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -135,80 +135,41 @@ final class DefaultUniverseClient(
...
-            } else {
-              // We have an extra file for some reason. Just ignore it.
-              oldV2Package
-            }
-          }
+    universeRepository.version match {
+      case Some(version) if version.toString.startsWith("2.") => // Valid
+      case Some(version) => throw new UnsupportedRepositoryVersion(version)
+      case _ => throw IndexNotFound(sourceUri)
+    }
```
cruhland: @jsancio this ensures that the Universe repository version gets checked before the package files (see below), as `PackageRepositorySpec` expects.

jsancio: Why don't we just remove the test? The code doesn't need to look at `index.json` anymore so it is strange that we look at it just to throw an exception. What do you think?

cruhland: Well, we still look at it to perform repo version validation. We could look at the `version.json` file instead, but then we'd probably have to create a new error type (e.g. `VersionFileNotFound`), which the CLI and UI would have to be updated to handle...

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -257,6 +218,71 @@ final class DefaultUniverseClient(
...
+
+      state.copy(packages = state.packages + ((packageKey, packageFiles)))
+    } else {
+      state
+    }
+  }
+
+  private[this] def processPackageFiles(
+    packageFiles: Map[String, (Path, Array[Byte])]
+  ): V2PackageInformation = {
```
jsancio: Okay but just wanted to point out that this roughly doubles the memory requirements. We roughly need to store the Zip in memory (`Map[String, (Path, Array[Byte])]`) and the decoded zip `V2PackageInformation`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: I see. I didn't notice the `UnsupportedRepositoryVersion` exception. LGTM!

jsancio:
LGTM! 🚢 it. Thanks for the clean up!

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
cruhland: Good point. I think it's okay for this PR, but if we want to address it soon, @BenWhitehead suggested that we should use applicative error types (instead of exceptions) when processing the zip file. That would allow us to decode the case classes in one pass, while reporting all errors that occurred to keep the tests happy.

If you think this is a good idea I can create an issue.

PR #406, opened 2016-06-21
Wire through RequestSession for UniverseClient

BenWhitehead:


jsancio:
LGTM. :shipit:

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -202,8 +202,7 @@ object Cosmos extends FinchServer {
     // TODO(version): How are we going to get the DC/OS release version??
```
cruhland: Delete this comment?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -44,7 +42,13 @@ final class UniverseClient(implicit statsReceiver: StatsReceiver = NullStatsRece
     Option(props.getProperty("cosmos.version")).getOrElse("unknown-version")
   }
 
-  override def apply(
+  def apply(repository: PackageRepository)(implicit session: RequestSession): Future[internal.model.CosmosInternalRepository] = {
+    adminRouter.getDcosVersion().flatMap { dcosVersion =>
```
cruhland: Do we have to look this up every time? Or can we cache it?

cruhland:
LGTM too. Two minor comments/questions.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
BenWhitehead: I think we need to look it up every time. Imagine a scenario where an admin is updating the masters and a new leading master is elected, but the users request gets routed to an existing cosmos, we want to make sure that cosmos is operating relative to the clusters.

BenWhitehead:
going to retarget master now that #374 has been merged.

Issue #409, opened 2016-06-22
Ensure Cosmos passes all existing CLI tests


cruhland:
- [ ] Run our DC/OS CLI integration tests against DC/OS testing channel: https://github.com/dcos/dcos/pull/268

PR #413, opened 2016-06-23
Enforce an explicit schema for the "package metadata" Marathon label

cruhland:
As a result, the ListHandler response will not contain the new
package fields from the Universe 3.0 schema.

This gets us closer to the CLI integration tests passing (#409). Some things to follow up on:
- @tamarrow this change doesn't fix the CLI test failures because we are now including the `selected` field explicitly in the label even if it wasn't specified in the package. So the CLI tests will need to be updated to expect that field.
- At one point we were running the CLI integration tests against Cosmos regularly. Should we start doing that again? It would have caught the `selected` discrepancy I mentioned above much earlier.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Label.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Label.scala
@@ -1,0 +1,85 @@
+package com.mesosphere.cosmos.converter
+
+import com.mesosphere.cosmos.converter.Universe._
+import com.mesosphere.cosmos.internal
+import com.mesosphere.cosmos.label
+import com.mesosphere.cosmos.rpc
+import com.mesosphere.universe
+import com.twitter.bijection.Conversion.asMethod
+import com.twitter.bijection.Conversion
+
+object Label {
+
+  implicit val internalPackageDefinitionToLabelV1PackageMetadata: Conversion[
+    internal.model.PackageDefinition, label.v1.model.PackageMetadata] =
+    Conversion.fromFunction { (x: internal.model.PackageDefinition) =>
+      label.v1.model.PackageMetadata(
+        packagingVersion = x.packagingVersion.as[universe.v2.model.PackagingVersion],
+        name = x.name,
+        version = x.version.as[universe.v2.model.PackageDetailsVersion],
+        maintainer = x.maintainer,
+        description = x.description,
+        tags = x.tags.as[List[String]],
+        selected = Some(x.selected),
+        scm = x.scm,
+        website = x.website,
+        framework = Some(x.framework),
+        preInstallNotes = x.preInstallNotes,
+        postInstallNotes = x.postInstallNotes,
+        postUninstallNotes = x.postUninstallNotes,
+        licenses = x.licenses.as[Option[List[universe.v2.model.License]]],
+        images = x.resource.flatMap(_.images).as[Option[universe.v2.model.Images]]
+      )
+    }
+
+    // TODO(version): This and the conversion below form a bijection
+  implicit val labelV1PackageMetadataToRpcV1InstalledPackageInformation: Conversion[
+    label.v1.model.PackageMetadata, rpc.v1.model.InstalledPackageInformation] =
+    Conversion.fromFunction { (x: label.v1.model.PackageMetadata) =>
+      rpc.v1.model.InstalledPackageInformation(
+        packageDefinition = universe.v2.model.PackageDetails(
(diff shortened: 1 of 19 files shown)


cruhland:
If there's more work that needs to be done on this, I won't have time to address it today. Please go ahead and add changes to it and merge if needed.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Label.scala
```diff
@@ -0,0 +1,83 @@
...
+        scm = x.packageDefinition.scm,
+        website = x.packageDefinition.website,
+        framework = x.packageDefinition.framework,
+        preInstallNotes = x.packageDefinition.preInstallNotes,
+        postInstallNotes = x.packageDefinition.postInstallNotes,
+        postUninstallNotes = x.packageDefinition.postUninstallNotes,
+        licenses = x.packageDefinition.licenses,
+        images = x.resourceDefinition.flatMap(_.images)
+      )
+    }
```
jsancio: Is there a reason why you decided to use `Conversion` for `rpcV1InstalledPackageInformationToLabelV1PackageMetadata` and `labelV1PackageMetadataToRpcV1InstalledPackageInformation` instead of `Bijection`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
```diff
@@ -427,4 +427,16 @@ object Universe {
...
+    Conversion.fromFunction { (x: internal.model.PackageDefinition) =>
+      universe.v2.model.PackageDetails(
+        packagingVersion = x.packagingVersion.as[universe.v2.model.PackagingVersion],
+        name = x.name,
+        version = x.version.as[universe.v2.model.PackageDetailsVersion],
+        maintainer = x.maintainer,
+        description = x.description
+      )
+    }
+
```
jsancio: You may not need this. We should have bijections/injections for `internal.model.PackageDefinition => universe.v3.model.PackageDefinition` and `universe.v3.model.PackageDefintion => universe.v2.model.PackageDetails`.

cosmos-server/src/test/scala/com/mesosphere/cosmos/converter/LabelSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  private[this] def roundTrip[A, B](a: A)(implicit
+    aToB: Conversion[A, B],
+    bToA: Conversion[B, A]
+  ): A = {
+    a.as[B].as[A]
+  }
+
+}
```
jsancio: Cool.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Label.scala
cruhland: Yep, it's because `InstalledPackageInformation` has the full `Resources` object, so converting to `PackageMetadata` potentially loses information. I really only need that direction for the test though, so maybe its definition should be moved under `src/test`.

jsancio:
LGTM!

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
cruhland: Hmm yeah this is old code I forgot to clean up. It can be deleted, since I don't think it's used anywhere.

jsancio:
Okay. I'll take over this PR.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Universe.scala
jsancio: Looks like we are using this code. CLI tests are failing because this code is dropping fields. Working on a fix.

tamarrow:
@jsancio is this ready to merge once the tests pass?

jsancio:
@tamarrow @BenWhitehead This is ready for review.

Both `test_package` and `test_service` in the CLI pass. I had to make some minor changes that I will pass the the CLI repo soon.

BenWhitehead:
@jsancio feel free to merge, though I think there's some cleanup we're going to want to take care of when we take care of the `todo (version)`s in the code. (I'm not sure the new metadata object should go in `com.mesosphere.cosmos.label` or someplace else. Having just a label package feels kinda orphaned from the marathon aspect which prompts it to be there at all.

PR #420, opened 2016-06-28
Clean up several TODO(version) related to conversions

BenWhitehead:
Conversions have been updated to be as explicit as possible on what is
allowed and what isn't.

* Update many stub injection/bijections have been replaced by
  Conversions
* ListHandler has been updated to try and remove the cyclic dependency on
  needing to convert bidirectionally between
  `internal.model.PackageDefinition` and `universe.v3.model.PackageDefinition`

Formatting of Universe.scala has been updated to try and make
declarations as consistent as possible.

Update CosmosIntegrationTestClient to be more tolerant of receiving a
token that isn't >= 10 characters. Prior to the fix
CosmosIntegrationTestClient would result in an
ExceptionInInitializerError if the token passed in was < 10 characters.


BenWhitehead:
@jsancio @tamarrow please review

jsancio:
LGTM! :shipit:

PR #424, opened 2016-06-29
Ensure DescribeResponse to not expose releaseVersion

### LINKED ISSUE #418: Remove "releaseVersion" from "rpc.v2.model.DescribeResponse"
We explicitly decided not to expose the release version in our "rpc.v1.model". We need to make sure that we don't start exposing it again in "rpc.v2.model". 

https://github.com/dcos/cosmos/blob/master/cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v2/model/DescribeResponse.scala#L10

cc @tamarrow @BenWhitehead

BenWhitehead:
Fixes #418


BenWhitehead:
Manually ran test suite against existing `testing/master` ccm cluster.

PR #427, opened 2016-06-30
Delete unused error types

### LINKED ISSUE #426: Delete `NelErrors` type
It's no longer used after the deletion of `UniversePackageCache`. If we have a reason to respond with multiple errors in the future, let's design a better replacement for `NelErrors` at that time.

cruhland:
Also delete `PackageImportHandler` since it is not used either.

Fixes #426.


tamarrow:
LGTM

jsancio:
Interesting. No MesosError? Do you happen to know why we don't need that error anymore?

cruhland:
@jsancio looks like both `MarathonAppMetadataError` and `MesosRequestError` were never used -- see [the PR that introduced them](https://github.com/dcos/cosmos/pull/82/files#diff-37a8f9d27c38e0283c9dbeb390b98c86R23).

PR #432, opened 2016-07-07
Fix /package/list so that it actually adheres to the v1 schema

BenWhitehead:
Fix a regression that was introduced that broke the schema of the v1 response for package list.


BenWhitehead:
Manually tested and verified with @cruhland . This handler will probably need to be refactored before it's feasible to write unit or integration tests for it.

cruhland:
LGTM!

Issue #124, opened 2016-01-30
sort results from `list` 


tamarrow:
Cosmos should send result for installed services from `package/install` sorted by package first and application id second.

jsancio:
@tamarrow how are we sorting them in the DC/OS CLI? Is it by `appId`.

PR #433, opened 2016-07-07
Resolved Issue #124

larioj:
The list operation now returns packages sorted by name and appId. A test of the issue was added as well as some private helper functions.


cruhland:
Thanks for the PR, @larioj! I'm leaving the office now, but I'll review first thing tomorrow.

cruhland:
Discussed some improvements with @larioj offline, waiting for update.

cosmos-model/src/main/scala/com/mesosphere/cosmos/thirdparty/marathon/model/AppId.scala
```diff
@@ -19,4 +19,6 @@ object AppId {
 
   def apply(s: String): AppId = new AppId(if (s.startsWith("/")) s else "/" + s)
 
+  implicit val ordering: Ordering[AppId] = Ordering.by[AppId, String](_.toString)
```
cruhland: For future reference, in most cases the type parameters to a generic method can be inferred by the compiler and don't need to be written explicitly. In this case, either of the following would work:

```scala
implicit val ordering: Ordering[AppId] = Ordering.by(_.toString)
implicit val ordering = Ordering.by[AppId, String](_.toString)
```

The first case is generally preferred since it gives an explicit type to the `ordering` variable.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
```diff
@@ -134,6 +134,70 @@ final class PackageListIntegrationSpec
...
+    assertResult(installResponse.packageVersion)(packageVersion)
+
+    uninstallResponse
+  }
+
+  private[this] def packageInstallMany(packageName: List[String]): List[InstallResponse] =
+    packageName map packageInstall
+
+  private[this] def packageUninstallMany(installResponses: List[InstallResponse]): List[UninstallResponse] =
+    installResponses map packageUninstall
```
cruhland: These methods are so short it's probably better to not have them. You can just write the code inline in the test method.

I also noticed that you don't use the return value from `packageUninstall`. You might as well leave off the return value from that method (change its return type to `Unit`) and then you can uninstall all packages with `installResponses foreach packageUninstall`.

cruhland:
@larioj nice work! Just a few comments and then it'll be good to merge.

cruhland:
LGTM, :ship: it!

PR #434, opened 2016-07-07
Update finch/circe config to exclude null properties from serialized json

### LINKED ISSUE #56: Circe inserts null values for omitted optional fields
When generating JSON for a case class, Circe will insert `null` values for omitted optional fields, instead of leaving them out.

For example, given the case class

```scala
case class Foo(bar: Int, baz: Option[String])
```

and the instance

```scala
val foo = Foo(bar = 5, baz = None)
```

Circe will generate

```json
{"bar": 5, "baz": null}
```

when we would prefer that it generates

```json
{"bar": 5}
```

Should we add this behavior to Circe, or just write custom encoders for Cosmos classes?

This is currently affecting JSON encoding of the `PackageDefinition` class.

BenWhitehead:
Fixes #56

### DIFF
--- a/cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
+++ b/cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
@@ -93,9 +93,6 @@ final class PackageDescribeSpec extends FreeSpec with TableDrivenPropertyChecks 
     val Right(packageJson) = packageInfo.cursor.get[PackageDetails]("package")
     assertResult(HelloworldPackageDef)(packageJson)
 
-    val Right(resourceJson) = packageInfo.cursor.get[Resource]("resource")
-    assertResult(HelloworldResource)(resourceJson)
-
     val Right(configJson) = packageInfo.cursor.get[Json]("config")
     assertResult(HelloworldConfigDef)(configJson)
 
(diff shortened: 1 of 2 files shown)


cruhland:
Easiest. Change. Ever.

:ship: it!

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -93,9 +93,6 @@ final class PackageDescribeSpec extends FreeSpec with TableDrivenPropertyChecks
     val Right(packageJson) = packageInfo.cursor.get[PackageDetails]("package")
     assertResult(HelloworldPackageDef)(packageJson)
 
-    val Right(resourceJson) = packageInfo.cursor.get[Resource]("resource")
-    assertResult(HelloworldResource)(resourceJson)
```
cruhland: If we keep this here, but test for a cursor decode error, then this would give us a test to ensure `null` fields are being dropped.

BenWhitehead: I don't think this test method is the correct location for this sort of assertion.

This type of test probably makes more sense for a suite of tests that verify server config separate from backend data or endpoint specific implementation.

BenWhitehead: Issue to track https://github.com/dcos/cosmos/issues/435

PR #436, opened 2016-07-11
Update readme to include version and compatiblity information

BenWhitehead:


README.md
```diff
@@ -58,3 +58,107 @@ export ZOOKEEPER_SECRET <secret>
...
+
+|       | application/zip (version-2.x) | application/vnd.dcos.universe.repo+json;charset=utf-8;version=v3 |
+| ----- | ----------------------------- | ---------------------------------------------------------------- |
+| 0.1.0 | Not Supported                 | Not Supported                                                    |
+| 0.1.1 | Not Supported                 | Not Supported                                                    |
+| 0.1.2 | Not Supported                 | Not Supported                                                    |
+| 0.1.3 | Not Supported                 | Not Supported                                                    |
+| 0.1.4 | Not Supported                 | Not Supported                                                    |
+| 0.1.5 | Not Supported                 | Not Supported                                                    |
+| 0.2.0 | Not Supported                 | Supported                                                        |
```
cruhland: Shouldn't the `application/zip` column have "Supported" for all entries?

cruhland: Also, to make the table less of a chore to read through, could we summarize it into two rows, `0.1.x` and `0.2.0`?

README.md
```diff
@@ -58,3 +58,107 @@ export ZOOKEEPER_SECRET <secret>
...
+
+|       |    2.0    |      3.0      |
+| ----- | --------- | ------------- |
+| 0.1.0 | Supported | Not Supported |
+| 0.1.1 | Supported | Not Supported |
+| 0.1.2 | Supported | Not Supported |
+| 0.1.3 | Supported | Not Supported |
+| 0.1.4 | Supported | Not Supported |
+| 0.1.5 | Supported | Not Supported |
+| 0.2.0 | Supported | Supported     |
```
cruhland: Can this table be summarized to `0.1.x` and `0.2.0` rows as well?

README.md
```diff
@@ -58,3 +58,107 @@ export ZOOKEEPER_SECRET <secret>
...
+
+### API Method Version Compatibility
+
+The following requests have constraints based on the version of the package from universe
```
cruhland: From the word "constraints" it sounds like we're really trying to convey when the endpoints will *not* work. If that's the case, this section could be made much shorter by documenting only the endpoint limitations, and stating that everything else will work.

For example, for `/package/install` we only need to say the following:

```
Content-Type: application/vnd.dcos.package.install-request+json;charset=utf-8;version=v1
Accept:       application/vnd.dcos.package.install-response+json;charset=utf-8;version=v1
```
A v1 install can fail when the package being installed was published as a Universe package with `packagingVersion` 3.0 and the package has no marathon template defined.

README.md
```diff
@@ -58,3 +58,107 @@ export ZOOKEEPER_SECRET <secret>
...
+
+#### `/package/describe`
+
+```
+Conent-Type: application/vnd.dcos.package.describe-request+json;charset=utf-8;version=v1
```
cruhland: Typo in `Conent-Type` -- applies to all occurrences below.

README.md
```diff
@@ -58,3 +58,107 @@ export ZOOKEEPER_SECRET <secret>
...
+Accept:      application/vnd.dcos.package.install-response+json;charset=utf-8;version=v1
+```
+A v1 install can succeed in the following scenarios:
+
+1. The package being installed was published as a Universe package with `packagingVersion` 2.0
+2. The package being installed was published as a Universe package with `packagingVersion` 3.0 and the package has a marathon template defined
+
+```
+Conent-Type: application/vnd.dcos.package.install-request+json;charset=utf-8;version=v1
+Accept:      application/vnd.dcos.package.install-response+json;charset=utf-8;version=v1
```
cruhland: This should have `version=v2`.

README.md
```diff
@@ -58,3 +58,107 @@ export ZOOKEEPER_SECRET <secret>
...
+```
+A v2 install can succeed in the following scenarios:
+
+1. The package being installed was published as a Universe package with `packagingVersion` 2.0
+2. The package being installed was published as a Universe package with `packagingVersion` 3.0 and the package has a marathon template defined
+3. The package being installed was published as a Universe package with `packagingVersion` 3.0 and the package has a `.cli` object defined in it's resource set
```
cruhland: Here's a good example where saying what can fail would be easier to interpret. In this case, the only scenario that can fail is a `packagingVersion` 3.0 package that has neither a marathon template nor a `.cli` object, correct? Or will even that case succeed?

README.md
BenWhitehead: Good catch, editing error. Will update.

README.md
BenWhitehead: sure

cruhland:
@BenWhitehead back to you.

cruhland:
@BenWhitehead LGTM

PR #440, opened 2016-07-15
Tests for `ReleaseVersion` converters

cruhland:
- Streamline definition of `universe.v3.model.PackageDefinition.ReleaseVersion`
- Add tests for and clean up converters for `ReleaseVersion`
- Use converters to define `Decoder`s for `ReleaseVersion`
- Define `ConversionFailureInjection` which enables a single error message template to be used through a chain of `Injection`s and `Bijection`s


cosmos-model/src/main/scala/com/mesosphere/cosmos/converter/Common.scala
```diff
@@ -0,0 +1,85 @@
...
+    Bijection.build(fwd)(rev)
+  }
+
+  implicit val v3ReleaseVersionToInt: Injection[universe.v3.model.PackageDefinition.ReleaseVersion, Int] = {
+    val fwd = (x: universe.v3.model.PackageDefinition.ReleaseVersion) => x.value
+    val rev = universe.v3.model.PackageDefinition.ReleaseVersion.apply _
+
+    ConversionFailureInjection(Injection.build(fwd)(rev)) { from =>
+      s"Expected integer value >= 0 for release version, but found [$from]"
+    }
```
cruhland: Here's an example use of `ConversionFailureInjection`. By using this here, the chained conversions below will automatically use the provided error message template.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
```diff
@@ -20,17 +21,17 @@ object PackageDefinition {
...
-    def max(that: ReleaseVersion): ReleaseVersion = {
-      if (value >= that.value) this
-      else that
-    }
-  }
   object ReleaseVersion {
-    import Ordering.Int
-    implicit val packageDefinitionReleaseVersionOrdering: Ordering[ReleaseVersion] = Ordering.by(_.value)
+
+    def apply(value: Int): Try[ReleaseVersion] = {
```
cruhland: Using an explicit `Try` instead of an assertion.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
```diff
@@ -20,17 +21,17 @@ object PackageDefinition {
     val packageDetailsTagPattern = Pattern.compile(packageDetailsTagRegex)
   }
 
-  case class ReleaseVersion(value: Int) {
-    assert(value >= 0, s"Value $value is not >= 0")
+  final class ReleaseVersion private(val value: Int) extends AnyVal
```
cruhland: Making this not a case class ensures that the constructor is actually `private` -- otherwise the auto-generated `apply` for case classes subverts it.

cruhland: By subclassing `AnyVal`, the `toString` representation still gives a way to see what `value` is.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
```diff
@@ -20,17 +21,17 @@ object PackageDefinition {
...
   object ReleaseVersion {
-    import Ordering.Int
-    implicit val packageDefinitionReleaseVersionOrdering: Ordering[ReleaseVersion] = Ordering.by(_.value)
+
+    def apply(value: Int): Try[ReleaseVersion] = {
+      if (value >= 0) Success(new ReleaseVersion(value))
+      else Failure(new IllegalArgumentException("negative value"))
+    }
+
+    implicit val ordering: Ordering[ReleaseVersion] = Ordering.by(_.value)
```
BenWhitehead: Let's keep the long name here on the ordering to try and reduce implicit scope collision.

cruhland: Implicits defined in companion objects don't need to be imported, which I think means they are not vulnerable to name collision. But I'm not completely sure, so I'll rename it to be safe.

BenWhitehead:
Overall looks good.

As mentioned offline, i'm not crazy about the `foo.as[Try[Bar]]` but we can chat about that for a different change.

PR #441, opened 2016-07-15
Update Service client to set User-Agent header

BenWhitehead:


cosmos-server/src/main/scala/com/mesosphere/cosmos/BuildProperties.scala
```diff
@@ -0,0 +1,30 @@
+package com.mesosphere.cosmos
+
+import java.util.Properties
+
+private[cosmos] class BuildProperties private[cosmos](resourceName: String) {
```
cruhland: Jose convinced me that we should move away from using `private[<package>]` and just stick to `private`, `private[this]`, or public. In this particular case, I think you can just have it be `private` and your test will still be able to access it because it's defined in the same package, `com.mesosphere.cosmos`.

cruhland: Ah, didn't see the scoping on the constructor. Presumably you could make that `private` and just test that it's constructable through the `apply` method.

cosmos-server/src/main/scala/com/mesosphere/cosmos/BuildProperties.scala
```diff
@@ -0,0 +1,30 @@
...
+    props
+  }
+}
+
+object BuildProperties {
+  private[this] val loaded = new BuildProperties("/build.properties")
+
+  def apply() = loaded
+
+  implicit class BuildPropertiesOps(val bp: BuildProperties) {
```
cruhland: This should extend `AnyVal` to eliminate instance allocation.

cosmos-server/src/main/scala/com/mesosphere/cosmos/ServiceClient.scala
```diff
@@ -14,6 +14,7 @@ import org.jboss.netty.handler.codec.http.HttpMethod
 abstract class ServiceClient(baseUri: Uri) {
 
   private[this] val cleanedBaseUri: String = Uris.stripTrailingSlash(baseUri)
+  private[this] val cosmosVersion: String = BuildProperties().cosmosVersion
```
cruhland: Why not just make this a local variable in `baseRequestReader`?

BenWhitehead: So we don't pollute the heap with many allocations for the implicit conversion of the class for each request built.

cruhland:
:ship: it

PR #442, opened 2016-07-18
List handler refactoring

larioj:
Refactored the code to be more testable and easier to read.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
@@ -2,6 +2,7 @@ package com.mesosphere.cosmos.handler
 
 import java.nio.charset.StandardCharsets
 import java.util.Base64
+
 import cats.data.Xor
 import com.mesosphere.cosmos.{AdminRouter, CirceError}
 import com.mesosphere.cosmos.converter.Label._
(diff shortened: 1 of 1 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -10,84 +11,86 @@ import com.mesosphere.cosmos.label
...
-import scala.util.Try
-
-private[cosmos] final class ListHandler(
-    adminRouter: AdminRouter,
-    repositories: (Uri) => Future[Option[CosmosRepository]]
-) extends EndpointHandler[rpc.v1.model.ListRequest, rpc.v1.model.ListResponse] {
 
-  override def apply(request: rpc.v1.model.ListRequest)(implicit session: RequestSession): Future[rpc.v1.model.ListResponse] = {
-    adminRouter.listApps().flatMap { applications =>
+private[cosmos] final class ListHandler(adminRouter: AdminRouter,
```
BenWhitehead: Style: Cosmos uses more of a "cut down" instead of "wrap" when lines are longer than 120 characters.

Checkout [`UninstallHandler.scala`](https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala#L21-L24) for an example of this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -10,84 +11,86 @@ import com.mesosphere.cosmos.label
...
+        case (app, Some(repo)) => {
+          // Will not fail because of satisfiesRequest
+          val (Some(pkgName), Some(pkgReleaseVersion)) = (app.packageName, app.packageReleaseVersion)
+          repo
+            .getPackageByReleaseVersion(pkgName, pkgReleaseVersion)
+            .map(pkgInfo => Installation(app.id, pkgInfo.as[InstalledPackageInformation]))
+        }
+        case (app, None) => Future.value(Installation(app.id, decodeInstalledPackageInformation(app)))
+      }
+    }.map(_.toList)
```
BenWhitehead: The `.map(_.toList)` can be avoided by returning a `Seq` and deferring the `_.toList` until constructing the response object. (This will reduce heap allocations)

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -10,84 +11,86 @@ import com.mesosphere.cosmos.label
...
+  private[this] def getRepositoryAssociations(repositories: (Uri) => Future[Option[CosmosRepository]],
+                                              apps: List[MarathonApp])
+  : Future[List[(MarathonApp, Option[CosmosRepository])]] = {
+    Future.collect {
+      apps.map { app =>
+        // Will not fail because of satisfiesRequest
+        val repoUri = app.packageRepository.get
+        repositories(repoUri).map(repo => (app, repo))
       }
+    }.map(_.toList)
```
BenWhitehead: The `.map(_.toList)` can be avoided by returning a `Seq` and deferring the `_.toList` until constructing the response object. (This will reduce heap allocations)

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -10,84 +11,86 @@ import com.mesosphere.cosmos.label
...
-private[cosmos] final class ListHandler(
-    adminRouter: AdminRouter,
-    repositories: (Uri) => Future[Option[CosmosRepository]]
-) extends EndpointHandler[rpc.v1.model.ListRequest, rpc.v1.model.ListResponse] {
 
-  override def apply(request: rpc.v1.model.ListRequest)(implicit session: RequestSession): Future[rpc.v1.model.ListResponse] = {
-    adminRouter.listApps().flatMap { applications =>
+private[cosmos] final class ListHandler(adminRouter: AdminRouter,
+                                        repositories: (Uri) => Future[Option[CosmosRepository]]
+                                       ) extends EndpointHandler[rpc.v1.model.ListRequest, rpc.v1.model.ListResponse] {
```
BenWhitehead: Since it's already imported, lets used the short name `ListRequest` and `ListResponse` here instead of the longer form. We should then be able to remove the import of the rpc package.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -10,84 +11,86 @@ import com.mesosphere.cosmos.label
...
-        (app.id, app.packageReleaseVersion, app.packageName, app.packageRepository, app.packageMetadata)
-      }
+  override def apply(request: rpc.v1.model.ListRequest)
+                    (implicit session: RequestSession): Future[rpc.v1.model.ListResponse] = {
+    for {
+      apps <- getApplications(adminRouter, request)
+      repoAssocs <- getRepositoryAssociations(repositories, apps)
+      installs <- getInstallations(repoAssocs)
+    } yield {
+      ListResponse(installs.sortBy(install =>
```
BenWhitehead: after removing `.map(_.toList)` from the private methods, it can be added here.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -10,84 +11,86 @@ import com.mesosphere.cosmos.label
...
-            }
-          } map { installedPackageInfo =>
-            rpc.v1.model.Installation(appId, installedPackageInfo)
-          }
+  private[this] def getRepositoryAssociations(repositories: (Uri) => Future[Option[CosmosRepository]],
+                                              apps: List[MarathonApp])
+  : Future[List[(MarathonApp, Option[CosmosRepository])]] = {
+    Future.collect {
+      apps.map { app =>
+        // Will not fail because of satisfiesRequest
```
BenWhitehead: Instead of adding a comment that claims the method will succeed as long as another method was called before it, it would be good to add more of a guarantee to enforce the expected state.

We can create a private inner case class that contains everything we need it to after it has been filtered.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -10,84 +11,86 @@ import com.mesosphere.cosmos.label
...
-    } map { installations =>
-      rpc.v1.model.ListResponse(installations.sortBy(i =>
-        (i.packageInformation.packageDefinition.name ,i.appId)))
-    }
+  private[this] def getInstallations(assocs: List[(MarathonApp, Option[CosmosRepository])])
+                                    (implicit session: RequestSession): Future[List[Installation]] = {
+    Future.collect {
+      assocs map {
+        case (app, Some(repo)) => {
+          // Will not fail because of satisfiesRequest
```
BenWhitehead: Instead of adding a comment that claims the method will succeed as long as another method was called before it, it would be good to add more of a guarantee to enforce the expected state.

We can create a private inner case class that contains everything we need it to after it has been filtered.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
```diff
@@ -10,84 +11,86 @@ import com.mesosphere.cosmos.label
...
 
-      val filtered = appData.filter {
-        case (appId, Some(releaseVersion), Some(packageName), Some(repositoryUri), metaData) =>
-          request.packageName.forall(_ == packageName) && request.appId.forall(_ == appId)
-        case _ => false
-      }
+  private[this] def getApplications(adminRouter: AdminRouter, request: ListRequest)
+                                   (implicit session: RequestSession): Future[List[MarathonApp]] = {
+    adminRouter.listApps().map { response =>
+      response.apps.filter(satisfiesRequest(_, request))
```
BenWhitehead: instead of doing just a `filter` here, we could do a collect that filters and results in a custom inner case class with the data structure in a form we can guarantee will work for the other methods.

BenWhitehead:
In the commit message, there are some typos:
*  `sesible` -> `sensible` 
* `recivers` -> `receivers`
* `organizasation` -> `organization`
* `repositiories` -> `repositories`

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ListHandler.scala
larioj: The change to Seq worked out nicely since ListResponse actually takes a Seq. No need for a conversion :)

PR #444, opened 2016-07-18
InstallResponse Conversion for V2 to V1

aeyakovenko:
- Add Conversion trait for InstallResponse
- Update MediaTypeEncoders to use the 'as' method for conversion
- Conversion Tests


cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
```diff
@@ -7,11 +7,14 @@ import com.mesosphere.cosmos.{ServiceMarathonTemplateNotFound, internal, rpc}
 import com.mesosphere.universe
 import com.mesosphere.universe.common.ByteBuffers
 import com.twitter.bijection.Conversion.asMethod
+import com.twitter.bijection.{Conversion}
```
BenWhitehead: Curly braces are not needed for a single class import

cosmos-server/src/test/scala/com/mesosphere/cosmos/converter/ResponseSpec.scala
```diff
@@ -0,0 +1,45 @@
+package com.mesosphere.cosmos.converter
+
+import com.mesosphere.cosmos.converter.Universe._
+import com.mesosphere.cosmos.label
+import com.mesosphere.cosmos.rpc
+import com.twitter.bijection.Conversion
+import com.twitter.bijection.Conversion.asMethod
+import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
+import com.mesosphere.cosmos.{ServiceMarathonTemplateNotFound}
```
BenWhitehead: Curly braces are not needed for a single class import

cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v2/circe/MediaTypedEncoders.scala
```diff
@@ -33,7 +35,7 @@ object MediaTypedEncoders {
       ),
       MediaTypedEncoder(
         encoder = rpc.v1.circe.Encoders.encodeInstallResponse.contramap { (x: rpc.v2.model.InstallResponse) =>
-          converter.Response.v2InstallResponseToV1InstallResponse(x).get()
+          x.as.get()
```
BenWhitehead: We've been explicit in setting the type arg when using `.as` something like `x.as[Try[rpc.v1.model.InstallResponse]].get()` would get this for us.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
```diff
@@ -7,11 +7,14 @@ import com.mesosphere.cosmos.{ServiceMarathonTemplateNotFound, internal, rpc}
...
 import com.twitter.util.Try
 
 object Response {
-
-  def v2InstallResponseToV1InstallResponse(x: rpc.v2.model.InstallResponse): Try[rpc.v1.model.InstallResponse] = {
+  implicit val internalV2ResponseToV1Response: Conversion[
+    rpc.v2.model.InstallResponse,
+    Try[rpc.v1.model.InstallResponse]
+  ] = Conversion.fromFunction { (x: rpc.v2.model.InstallResponse) =>
```
BenWhitehead: This val can probably be expressed as an `Injection[rpc.v1.model.InstallResponse, rpc.v2.model.InstallResponse]`.  This is the case because a v1 response can always be represented as a v2 response, but not necessarily the other direction.  With the injection defined, this conversion would then just be an `invert` of the injection.

cosmos-server/src/test/scala/com/mesosphere/cosmos/converter/ResponseSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+final class ResponseSpec extends FreeSpec {
+
+  val vstring = "9.87.654.3210"
+  val name = "InstallResponse"
+  val ver = universe.v3.model.PackageDefinition.Version(vstring)
+  val notes = List(None, Some("post install notes"))
+  val clis = List(None, Some(Cli(None)))
+  val appid = AppId("foobar")
+
+  "Conversion[rpc.v2.model.InstallResponse,Try[rpc.v1.model.InstallResponse]]" - {
```
BenWhitehead: @cruhland Can probably show you how to set up a property table for these test cases.

Or you could checkout `UserOptionSpec.scala` and the docs: http://www.scalatest.org/user_guide/table_driven_property_checks

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
aeyakovenko: I thought as well, @cruhland wanted just a one way conversion since the other one isn't used anywhere else.  Its easy enough to add an Injection since all the other V2 members are Optional

BenWhitehead: Unidirectional is fine with me.

cosmos-server/src/test/scala/com/mesosphere/cosmos/converter/ResponseSpec.scala
aeyakovenko: Generator seems like a better choice, otherwise i have to permute all the options between notes and clis by hand in the table.

BenWhitehead: Sounds good

BenWhitehead:
:ship: it when integration tests pass

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
cruhland: I'm not opposed to providing both directions, I just don't like adding code that isn't going to be used. One way that it can be useful is that you can do a "round trip" test of `x == backwards(forwards(x))`, testing both directions in a small amount of code. Should we agree to write those tests, and always use bidirectional conversions?

PR #445, opened 2016-07-19
PackageDefintion Conversion

aeyakovenko:
- Implement trait Conversion[internal.model.PackageDefinition, Try[rpc.v1.model.DescribeResponse]]

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
@@ -24,9 +24,10 @@ object Response {
     }
   }
 
-  def packageDefinitionToDescribeResponse(
-    packageDefinition: internal.model.PackageDefinition
-  ): Try[rpc.v1.model.DescribeResponse] = {
+  implicit val internalPackageDefinitionToDescribeResponse: Conversion[
+    internal.model.PackageDefinition,
+    Try[rpc.v1.model.DescribeResponse]
+  ] = Conversion.fromFunction { (packageDefinition: internal.model.PackageDefinition) =>
     Try(packageDefinition.marathon.map(_.v2AppMustacheTemplate).getOrElse(
       throw ServiceMarathonTemplateNotFound(packageDefinition.name, packageDefinition.version)
     )).map { b64MarathonTemplate =>
@@ -54,5 +55,4 @@ object Response {
       )
     }
   }
-
 }
--- a/cosmos-server/src/test/scala/com/mesosphere/cosmos/converter/ResponseSpec.scala
+++ b/cosmos-server/src/test/scala/com/mesosphere/cosmos/converter/ResponseSpec.scala
@@ -45,20 +45,23 @@ final class ResponseSpec extends FreeSpec {
   }
   "Conversion[internal.model.PackageDefinition,Try[rpc.v1.model.DescribeResponse]]" - {
     "success" in {
+      val marathon = "dGVzdGluZw=="
       val p =  internal.model.PackageDefinition(universe.v3.model.V3PackagingVersion,
                                                 name,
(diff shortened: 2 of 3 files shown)


cosmos-server/src/test/scala/com/mesosphere/cosmos/converter/ResponseSpec.scala
```diff
@@ -18,31 +16,63 @@ import org.scalatest.prop.GeneratorDrivenPropertyChecks.forAll
...
+      n <- Gen.oneOf(clis)
+      c <- Gen.oneOf(notes)
+    } yield (rpc.v2.model.InstallResponse(name, ver, None, n, c))
+
+
     "success" in {
       val v1 = rpc.v1.model.InstallResponse(name, ver.as[universe.v2.model.PackageDetailsVersion], appid)
 
-      forAll(validV2s) { x => assert(x.as[Try[rpc.v1.model.InstallResponse]] == Return(v1)) }
+      forAll(validV2s) { x => assertResult(x.as[Try[rpc.v1.model.InstallResponse]])(Return(v1)) }
```
cruhland: `assertResult` wants the "expected" value as the first argument, and the "actual" value as the second. The arguments are in different argument lists so you can do something like:

```scala
assertResult(myExpectedValue) {
  val foo = ...
  val bar = ...
  foo ++ bar
}
```

cosmos-server/src/test/scala/com/mesosphere/cosmos/converter/ResponseSpec.scala
```diff
@@ -18,31 +16,63 @@ import org.scalatest.prop.GeneratorDrivenPropertyChecks.forAll
...
+                                                "maintainer",
+                                                "description",
+                                                 marathon = Some(universe.v3.model.Marathon(java.nio.ByteBuffer.allocate(marathon.length).put(marathon.getBytes))))
+      val d = universe.v2.model.PackageDetails(universe.v2.model.PackagingVersion("3.0"),
+                                               name,
+                                               ver.as[universe.v2.model.PackageDetailsVersion],
+                                               "maintainer",
+                                               "description",
+                                               selected = Some(false),
+                                               framework = Some(false))
```
cruhland: This doesn't test all fields of the `PackageDefinition`. Fortunately, there's some existing test data you can use. In `MarathonLabelsSpec`, there's a `MinimalPackageDefinition` and a `MaximalPackageDefinition` -- could you move them into a test utility class somewhere so they're easier to find?

cruhland: Not as useful for this change, but there are also some repository JSON files for testing located under `cosmos-json/src/test/resources/com/mesosphere/universe/v3/circe/RepositoryEncoderDecoderSpec/`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
```diff
@@ -24,9 +24,10 @@ object Response {
     }
   }
 
-  def packageDefinitionToDescribeResponse(
-    packageDefinition: internal.model.PackageDefinition
-  ): Try[rpc.v1.model.DescribeResponse] = {
+  implicit val internalPackageDefinitionToDescribeResponse: Conversion[
+    internal.model.PackageDefinition,
+    Try[rpc.v1.model.DescribeResponse]
+  ] = Conversion.fromFunction { (packageDefinition: internal.model.PackageDefinition) =>
```
cruhland: Not sure if you saw my (late) comment on your last PR ([here](https://github.com/dcos/cosmos/pull/444/files#r71259679)), but if it would be easier to test then I'd be in favor using an `Injection` here and testing that some round trip conversions succeed while others fail.

PR #448, opened 2016-07-23
make test for list handler clean up on failure

larioj:
will also split test into two

### DIFF
--- a/cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
+++ b/cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
@@ -1,6 +1,7 @@
 package com.mesosphere.cosmos
 
 import java.util.UUID
+
 import cats.data.Xor
 import com.mesosphere.cosmos.http.MediaTypes
 import com.mesosphere.cosmos.label.v1.model.PackageMetadata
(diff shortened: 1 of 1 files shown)


cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
```diff
@@ -136,18 +136,37 @@ final class PackageListIntegrationSpec
...
+        "linkerd",
+        "linkerd",
+        "zeppelin",
+        "jenkins",
+        "cassandra")
+      val installResponses = apps map packageInstall
+      try {
+        assert(installResponses.size == apps.size)
+      } finally {
+        installResponses foreach packageUninstall
```
BenWhitehead: What would happen if an uninstall failed for one of the install responses processed here?

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
```diff
@@ -136,18 +136,37 @@ final class PackageListIntegrationSpec
...
+    "all installed packages" in {
+      val apps = List(
+        "linkerd",
+        "linkerd",
+        "zeppelin",
+        "jenkins",
+        "cassandra")
+      val installResponses = apps map packageInstall
+      try {
+        assert(installResponses.size == apps.size)
```
BenWhitehead: it would be good to have a stronger assertion here. if there was a bug in the install logic it'd be possible for this assertion to succeed. Something simple like name and version would be sufficient.  The package repo we use for tests is stable with a fixed set of packages.

BenWhitehead:
@larioj Please rebase and address comments from previous review or we'll need to close this since it's been open for > 2 weeks.

larioj:
@BenWhitehead Sorry about that

BenWhitehead:
merge it

PR #450, opened 2016-07-25
Test/universe conversion

aeyakovenko:
- added definitions for v2 model and v3 model V2/V3 Package defs
- test that cover conversion between minimal and maximal defs


aeyakovenko:
I left out tests for Images, Container, Assets, Resource, Command and LIcense.  They are tested by the data structures that wrap them.  Let me know if i should add them.

BenWhitehead:
:+1: on using `in` instead of `-` for the test declarations.

Feel free to merge after integration tests succeed.

PR #458, opened 2016-07-30
Integration test cleanup for UniverseClient

cruhland:
Also added a test for the correct handling of nonexistent repo files.


cosmos-server/src/it/scala/com/mesosphere/cosmos/repository/UniverseClientSpec.scala
```diff
@@ -49,6 +57,53 @@ final class UniverseClientSpec extends FreeSpec {
...
+
+    "should fail to fetch a nonexistent repo file" in {
+      val version = DcosReleaseVersionParser.parseUnsafe("0.0")
+      val repoUri = baseRepoUri / "doesnotexist.json"
+      val result = universeClient(PackageRepository("badRepo", repoUri), version)
+      val Throw(GenericHttpError(method, uri, status)) = Await.result(result.liftToTry)
+
+      assertResult("GET")(method.getName)
+      assertResult(repoUri)(uri)
+      assertResult(403)(status.code)
```
cruhland: According to HTTP, this status code should technically be a 502 Bad Gateway, because Cosmos is acting as a gateway to the repo server, which responded with the 403. We could add the upstream server's status code as a new field in `GenericHttpError`, which should show up under the `data` field in the JSON error response we generate. What do you think?

BenWhitehead: I don't think cosmos is functioning as a gateway (it just so happens that fetches of configured repos happen in the path of a request, something that we hope to decouple in the future), and since we're not functioning as a gateway that basically leaves us with 500 as the only possible status code.

There also already exists more robust handling for finagle clients: https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/Services.scala#L67 when this code is re-written to use the finagle client it will be easier to handle things more specifically.

aeyakovenko: is the 403 caused by users input?  seems like we should forward the permission errors to the client.

cruhland: @aeyakovenko technically yes, but the user input is provided in a completely different request, so it could be confusing to send 403. To trigger the error, the client would first do a `POST /package/repository/add` with the bad repo URI, which would respond with a 200 indicating the URI was successfully added to the repository list. Then, when the user does a `POST /package/install`, for example, Cosmos will attempt to resolve the URI and *that's* when the 403 is generated.

cruhland: @BenWhitehead so do we want to have Cosmos respond with 500 here, but include the "upstream" status in the error data? Or just forward the "upstream" status as it's currently doing?

BenWhitehead: My preference is that we just forward along and make it more robust in 1.9

PR #459, opened 2016-08-01
Test/multi repository spec

aeyakovenko:
- Unit tests for MultiRepository.scala
- Removed unreachable exception


cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
```diff
@@ -51,7 +51,6 @@ final class MultiRepository(
             .map(Some(_))
             .handle {
               case PackageNotFound(_) => None
-              case VersionNotFound(_, _) => None
```
aeyakovenko: unless this is a bug, https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/CosmosRepository.scala#L74 and should return VersionNotFound

BenWhitehead: I think this is the "fallback" point for multi repo since an individual repo will throw `VersionNotFound` if it has the package, but not version in it. @cruhland can you verify?

aeyakovenko: CosmosRepository will never throw a VersionNotFound, and throwing it here doesn't quite make sense either.

https://github.com/dcos/cosmos/blob/a08ac8132e7a7f069c1c2c1358d91dc05ae56974/cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala#L60

Basically it only throws it if the "version" argument is supplied, regardless if it finds anything matching that name.  Should the behavior be to only throw the VersionNotFound error if the name exists?

cruhland: @aeyakovenko actually, in the code you linked, `MultiRepository` will only throw `PackageNotFound` or `VersionNotFound` if no packages were found at all (hence `packages.headOption.getOrElse`).

I agree that the distinction of `PackageNotFound` vs. `VersionNotFound` is pretty trivial, essentially `VersionNotFound` is only thrown if an explicit version was supplied in the request. Perhaps an earlier implementation of `CosmosRepository.getPackageByPackageVersion` actually threw `VersionNotFound` if there were packages matching the name, but not the version, provided by the request -- which would explain why `MultiRepository` tries to handle `VersionNotFound`.

I'd actually be in favor of being more explicit, and have `CosmosRepository.getPackageByPackageVersion` throw `VersionNotFound`. If that's not a good idea for some reason, then the next best thing would be to not bother throwing `VersionNotFound` from `MultiRepository` either, to be consistent.

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+import cats.data.Ior
+import cats.data.Ior.{Left,Right,Both}
+
+final class MultiRepositorySpec extends FreeSpec {
+
+  case class TestClient(repos: List[PackageRepository] = Nil, ls: List[PackageDefinition] = Nil) extends UniverseClient {
+    def apply(repo: PackageRepository)(implicit session: RequestSession): Future[CosmosInternalRepository] = Future { 
+      val rv = repos.filter( _ == repo).flatMap(_ => ls)
+      rv match {
+        case Nil => throw PackageNotFound("foobar")
```
cruhland: Why does this throw if the list is empty? Shouldn't it just return an empty repository?

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+      }
+    }
+  }
+  case class TestStorage(initial:List[PackageRepository] = Nil) extends PackageSourcesStorage {
+    var cache: List[PackageRepository] = initial
+    def read(): Future[List[PackageRepository]] =  Future { cache }
+
+    def readCache(): Future[List[PackageRepository]] =  Future { cache }
+
+    def add(index: Option[Int], packageRepository: PackageRepository): Future[List[PackageRepository]] = sys.error("add")
```
cruhland: Since this is a ScalaTest class, let's use `assert(false, "add")` instead of `sys.error`?

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+
+    def delete(nameOrUri: Ior[String, Uri]): Future[List[PackageRepository]] =  sys.error("delete")
```
cruhland: Same here, use `assert`?

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+  }
+  "getRepository" - {
+    "empty" in {
+      val c = new MultiRepository(TestStorage(), TestClient())
+      assertResult(None)(Await.result(c.getRepository(Uri.parse("/test"))))
+    }
+    "many" in {
+      val one = PackageRepository("one", Uri.parse("/one"))
+      val two = PackageRepository("two", Uri.parse("/two"))
```
cruhland: What if a repository has the same URI as another?

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+      val storage = TestStorage(repos)
+      val cls = List(TestUtil.MinimalPackageDefinition)
+      val client = TestClient(repos, cls)
+      val c = new MultiRepository(storage, client)
+      val ver = TestUtil.MinimalPackageDefinition.version
+
+      assertResult(Return(cls))(Try(Await.result(c.getPackagesByPackageName("minimal"))))
+
+      assertResult(Return((cls.head, u)))(Try(Await.result(c.getPackageByPackageVersion("minimal", None))))
+      assertResult(Return((cls.head, u)))(Try(Await.result(c.getPackageByPackageVersion("minimal", Some(ver)))))
```
cruhland: Add test case where name is `minimal` and version is `Some(x)` where `x` is not a version that is present?

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+
+      assertResult(Return("minimal"))(Try(Await.result(c.search(None)).head.name))
```
cruhland: Verify the full list of results rather than just the `head`?

```scala
assertResult(Return(List("minimal")))(Try(Await.result(c.search(None)).map(_.name)))
```

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+      assertResult(Return("minimal"))(Try(Await.result(c.search(Some("minimal"))).head.name))
+    }
```
cruhland: Add a test to show that searching for something not related to `minimal` produces no results?

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+    "invalid repo" in {
+      val invalid = List(PackageRepository("invalid", Uri.parse("/invalid")))
+      val repos = List(PackageRepository("minimal", Uri.parse("/test")))
+      val storage = TestStorage(repos)
+      val cls = List(TestUtil.MinimalPackageDefinition)
+      val client = TestClient(invalid, cls)
+      val c = new MultiRepository(storage, client)
```
cruhland: What situation is this trying to test?

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+      val client = TestClient(repos, cls)
+      val c = new MultiRepository(storage, client)
+      val ver = TestUtil.MinimalPackageDefinition.version
+      val badver = TestUtil.MaximalPackageDefinition.version
+
+      assertResult(Throw(new PackageNotFound("MAXIMAL")))(Try(Await.result(c.getPackagesByPackageName("MAXIMAL"))))
+
+      assertResult(Throw(new PackageNotFound("MAXIMAL")))(Try(Await.result(c.getPackageByPackageVersion("MAXIMAL", None))))
+      assertResult(Throw(new VersionNotFound("MAXIMAL", ver)))(Try(Await.result(c.getPackageByPackageVersion("MAXIMAL", Some(ver)))))
+      assertResult(Throw(new VersionNotFound("minimal", badver)))(Try(Await.result(c.getPackageByPackageVersion("minimal", Some(badver)))))
```
cruhland: Ah ok, this covers my [comment above](https://github.com/dcos/cosmos/pull/459/files#r73073358).

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,139 @@
...
+    }
+    "from many" in {
+      val u = Uri.parse("/valid")
+      val repos = List(PackageRepository("valid", u))
+      val storage = TestStorage(repos)
+      val cls = List(TestUtil.MaximalPackageDefinition, TestUtil.MinimalPackageDefinition)
```
cruhland: This tests a single repo, multiple package scenario; what about multiple repo, single (or multiple) package?

cruhland:
@aeyakovenko left some suggestions for additional tests, let me know what you think.

cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
aeyakovenko: @cruhland Yea, I agree.  Anyone depend on the VersionNotFound exception?

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
aeyakovenko: yea, you are right.  i got confused between the UniverseClient and CosmosRepository.

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
aeyakovenko: uses `find` so the first matching one is returned.  Seems like a bug

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
aeyakovenko: assert returns Unit, seems like sys.error makes more sense for unimplemented methods?

BenWhitehead: Assert returns unit if the condition holds, which in the case of `false` will fail and an assertionerror will be throw, which will be picked up by the testing libraries and given special formating and handling instead of a generic runtime exception.

cruhland: Oh I see what you're saying @aeyakovenko, because `assert` returns `Unit` it doesn't type check. Given that the test library will still report failure if a generic exception is thrown, keeping the `sys.error` calls should be fine. Alternatively you could use [`???` which throws `NotImplementedError`](http://www.scala-lang.org/api/current/index.html#scala.Predef$@???:Nothing).

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
aeyakovenko: since it stores in the results in the Map, you get the last one added as the repo

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
aeyakovenko: That my client uri is different from the repo uri used by storage.  It's verifying that my test code correctly interprets the repo argument in `TestClient.apply` and returns an empty cosmos repository.  Probably not super necessary.

cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
aeyakovenko: @cruhland @BenWhitehead Fixed CosmosRepository to return VersionNotFound if the name matches but version doesn't.
https://github.com/dcos/cosmos/pull/459/files#diff-1376a283ca99009cb99d101e7f1ca9e5L65

cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
```diff
@@ -24,11 +24,9 @@ final class MultiRepository(
     repositories().flatMap { repositories =>
       Future.collect {
         repositories.map { repository =>
-          repository.getPackagesByPackageName(packageName).map(Some(_)).handle {
```
aeyakovenko: this is unreachable because `CosmosRepository.getPackagesByName` doesn't throw, just returns an empty list

https://github.com/dcos/cosmos/blob/f95ce5026ff20ccae435a9c9f245c84ee800e50a/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/CosmosRepository.scala#L83

cruhland: Heh, wow, didn't even notice that!

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/CosmosRepository.scala
```diff
@@ -65,13 +66,16 @@ final class DefaultCosmosRepository(
...
+      val ns = internalRepository.packages.filter { pkg =>
+        pkg.name == packageName
+      }
+      val vs = ns find { pkg =>
         packageVersion.map(_ == pkg.version).getOrElse(true)
-      } map { pkg =>
-        (pkg, repository.uri)
-      } getOrElse {
-        throw PackageNotFound(packageName)
+      }
```
cruhland: Slightly more efficient:

```scala
val vs = packageVersion match {
  case Some(pkgVer) => ns.find(_.version == pkgVer)
  case _ => ns.headOption
}
```

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
cruhland: Discussed offline -- this isn't really a concern, since

1. We reject duplicate URIs sent to `/package/repository/add`
1. Even if duplicate URIs somehow ended up in the repo list, from an end user's perspective the behavior would be identical (modulo different repo names, which is essentially harmless).

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,232 @@
...
+      assertResult(one)(Await.result(c.getRepository(Uri.parse("/one"))).get.repository)
+      assertResult(two)(Await.result(c.getRepository(Uri.parse("/two"))).get.repository)
+    }
+    "many same uri" in {
+      val one = PackageRepository("one", Uri.parse("/same"))
+      val two = PackageRepository("two", Uri.parse("/same"))
+      val c = new MultiRepository(TestStorage(List(one,two)), TestClient())
+      assertResult(None)(Await.result(c.getRepository(Uri.parse("/zero"))))
+      //get last one added
+      assertResult(two)(Await.result(c.getRepository(Uri.parse("/same"))).get.repository)
```
cruhland: Ah, I forgot about `getRepository`. In line with my [above comment](https://github.com/dcos/cosmos/pull/459/files#r73252147), let's not constrain the behavior too much here since it doesn't really matter. Can you weaken the test to the following?

```scala
assertResult(two.uri)(Await.result(c.getRepository(Uri.parse("/same"))).get.repository.uri)
```

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,232 @@
...
+      }
+      val rmax = PackageRepository("max", Uri.parse("/MAXIMAL"))
+      val rmin = PackageRepository("min", Uri.parse("/minimal"))
+      val min2 = TestUtil.MinimalPackageDefinition.copy(version = Version("1.2.4"))
+      val max2 = TestUtil.MaximalPackageDefinition.copy(version = Version("9.9.9"))
+      val clientdata = List((rmax, List(TestUtil.MaximalPackageDefinition,max2)),
+                            (rmin, List(TestUtil.MinimalPackageDefinition,min2)))
+      val storage = TestStorage(List(rmax,rmin))
+      val client = TestMultiClient(clientdata)
+      val c = new MultiRepository(storage, client)
```
cruhland: Needing all this setup code is a sign that we need to refactor `MultiRepository` at some point...

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/MultiRepositorySpec.scala
```diff
@@ -0,0 +1,232 @@
...
+      assertResult(Return(List("MAXIMAL", "minimal")))(Try(Await.result(c.search(None)).map(_.name).sorted))
+      assertResult(Return(List("minimal")))(Try(Await.result(c.search(Some("minimal"))).map(_.name)))
+      val min2ver = min2.version
+      assertResult(Return(List(Set(minver, min2ver))))(Try(Await.result(c.search(Some("minimal"))).map(_.versions.keys)))
+
+      val max2ver = max2.version
+      assertResult(Return(List(Set(maxver, max2ver))))(Try(Await.result(c.search(Some("MAXIMAL"))).map(_.versions.keys)))
+    }
+    "multi repo multi different packages" in {
+      case class TestMultiClient(repos: List[(PackageRepository,List[PackageDefinition])] = Nil) extends UniverseClient {
```
cruhland: Isn't this the same class that was defined in the test above?

cruhland:
@aeyakovenko two comments that you might want to address, then it's good to go!

cruhland:
LGTM, :ship: it

PR #462, opened 2016-08-09
Package Publish V1

larioj:
Using this for testing


cosmos-json/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/Decoders.scala
```diff
@@ -2,6 +2,7 @@ package com.mesosphere.cosmos.rpc.v1.circe
 
 import com.mesosphere.cosmos.rpc.v1.model._
 import com.mesosphere.cosmos.thirdparty.marathon.circe.Decoders._
+import com.mesosphere.packagestore.v3.circe.Decoders._
```
cruhland: The version of data structures in `packagestore` should probably be `v1`, not `v3`, since they are distinct from the Universe v3 definitions.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PublishResponse.scala
```diff
@@ -0,0 +1,5 @@
+package com.mesosphere.cosmos.rpc.v1.model
+
+import com.mesosphere.universe.v3.model.PackageDefinition
+
+case class PublishResponse(packageDefinition: PackageDefinition)
```
cruhland: We might not want to send the full Universe v3 package definition in the response. What will be useful for the CLI and the UI to display after a successful publish?

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
```diff
@@ -0,0 +1,98 @@
+package com.mesosphere.packagestore.v3.model
```
cruhland: Again, this should be `v1`, not `v3`.

cruhland: And actually, this file should be moved to the `com.mesosphere.cosmos.rpc.v1.model` package, because this data structure is used in the `PublishRequest`.

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
```diff
@@ -0,0 +1,98 @@
...
+
+import com.mesosphere.universe.v3.model._
+import io.circe.JsonObject
+
+sealed abstract class PackageBundle {
+  def toPackageDefinition(releaseVersion: PackageDefinition.ReleaseVersion): PackageDefinition
```
cruhland: We prefer not to have members on pure data objects, to increase the potential for reuse. Instead, create an `Injection` or a `Bijection` to do the conversion. See the `converter` package in the `cosmos-model` project.

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
```diff
@@ -0,0 +1,98 @@
...
+      postInstallNotes,
+      postUninstallNotes,
+      licenses,
+      minDcosReleaseVersion,
+      marathon,
+      resource,
+      config,
+      command
+    )
+}
```
cruhland: Add a newline here.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -82,6 +82,13 @@ private[cosmos] final class Cosmos(
     route(post("package" / "repository" / "delete"), deleteRepositoryHandler)(RequestReaders.standard)
   }
 
+  // TODO: This will need to change. Doing this to avoid compiler errors when changing Cosmos Signature.
```
cruhland: Could you put a token after the `TODO` to indicate who/what should clean this up? Your email name or a GitHub issue number would be good choices.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -82,6 +82,13 @@ private[cosmos] final class Cosmos(
...
+  val storage = Storage()
+  val packagePublish: Endpoint[Json] =
+    route(post("package" / "publish"), new PackagePublishHandler(storage))(RequestReaders.standard)
+  val repositoryServe: Endpoint[Json] =
+    route(get("service" / "ps" / "repository"), new RepositoryServeHandler(storage))(RequestReaders.noBody)
```
cruhland: This is mostly a matter of opinion, but what about something like `"package" / "storage" / "repo"`? Even though it's intended to be temporary, we should still give clear names because these things often last longer than planned.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Storage.scala
```diff
@@ -0,0 +1,29 @@
+package com.mesosphere.cosmos
+
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicInteger
+
+import com.mesosphere.packagestore.v3.model.PackageBundle
+import com.mesosphere.universe.v3.model.{PackageDefinition, Repository}
+
+case class Storage() {
+  private[this] val nextReleaseVersion: AtomicInteger = new AtomicInteger(0)
```
cruhland: While a single counter for the entire repo does ensure that all uploaded packages are unique, it's not the same behavior as Universe. To match our users' expectations, we should have a counter per package name.

We can also have the counter be implicit in the `packages` data structure if we define it as a `ConcurrentMap[String, ConcurrentLinkedQueue[PackageBundle]]`. The `getRepository` method would traverse `packages`, converting each `PackageBundle` to a `PackageDefinition` using the index of the bundle in its queue.

What do you think? cc @jsancio @BenWhitehead

cosmos-server/src/main/scala/com/mesosphere/cosmos/Storage.scala
```diff
@@ -0,0 +1,29 @@
+package com.mesosphere.cosmos
+
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicInteger
+
+import com.mesosphere.packagestore.v3.model.PackageBundle
+import com.mesosphere.universe.v3.model.{PackageDefinition, Repository}
+
+case class Storage() {
```
cruhland: Since this has private fields, it shouldn't be a case class. And it should have a more descriptive name. How about `final class PackageStorage`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
```diff
@@ -12,6 +12,14 @@ object MediaTypes {
   val applicationJson = MediaType("application", MediaTypeSubType("json"), Map("charset" -> "utf-8"))
   val applicationZip = MediaType("application", MediaTypeSubType("zip"))
 
+  val repositoryServeResponse = MediaType(
```
cruhland: This is identical to `UniverseV3Repository`, why make a different one?

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
```diff
@@ -12,6 +12,14 @@ object MediaTypes {
...
+    "application",
+    MediaTypeSubType("vnd.dcos.universe.repo", Some("json")),
+    Map("charset" -> "utf-8", "version" -> "v3")
+  )
+  val publishRequest = vnd("publish-request")
+  val publishResponse = vnd("publish-response")
```
cruhland: Please capitalize the above two `val`s to indicate that they are program constants.

cruhland:
@larioj added some comments. Is this code intended to be a throw-away prototype? If not, we should at least have an integration test to validate its behavior, and ideally some unit tests as well for the parts that aren't going to change soon.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PublishResponse.scala
larioj: The reason why I thought it useful to send back the package definition is because it is a more general version of sending the stuff that changed (i.e. the releaseVersion). This tells the user: "you uploaded a package bundle, whatever that may be, and we turned it into a package definition that you can now use. Here is what that definition looks like. Use whatever information you may be interested in."

cosmos-json/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/Decoders.scala
larioj: They are distinct but they correspond to universe v3 data structures. Does that matter? Or should the versioning be independent?

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
larioj: Thanks

BenWhitehead:
Do not merge this before 0.2.0 is released

cosmos-server/src/main/scala/com/mesosphere/cosmos/Storage.scala
larioj: That a good idea. I can do the transformation after I turn the queue into an array.

cosmos-json/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/Decoders.scala
BenWhitehead: the "package store" is v1, the fact that it can handle universe v3 documents because the schemas happen to overlap should not impact the versioning of the package store.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PublishResponse.scala
BenWhitehead: We're not doing REST, we're doing RPC. The pattern of "echo back what we resolved" is a pretty poor one IMO. if the user needs/can get back at the thing we should provide a reference to them where they can get it.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Storage.scala
BenWhitehead: I would like us to take as many steps as is practical to kill `releaseVersion` as soon as possible. But the recommendation to use a global counter was also before this got overhauled.

larioj: I could keep a map of counters.

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
```diff
@@ -0,0 +1,98 @@
+package com.mesosphere.packagestore.v3.model
+
+import com.mesosphere.universe.v3.model._
+import io.circe.JsonObject
+
+sealed abstract class PackageBundle {
```
BenWhitehead: What does this class do that [`universe.v3.model.PackageDefinition`](https://github.com/dcos/cosmos/blob/master/cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala#L8) does not?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Storage.scala
larioj: Yeah, that's a better name.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
larioj: I did not know. Will change.

BenWhitehead: Agreed, just use `UniverseV3Repository`

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
larioj: It does not have the releaseVersion

cruhland:
@BenWhitehead regarding the 0.2.0 release: do we want to make that now? If we discover that bugfixes are needed, we can always make a branch off of the `v0.2.0` tag to create further 0.2.x releases.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PublishResponse.scala
larioj: So I should not send back anything?

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
larioj: Should I even have a "packagestore" namespace?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
larioj: Will do

larioj:
@cruhland I agree on the tests. I will get started on that as soon as I finish the above changes.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PublishResponse.scala
cruhland: For now, let's not send anything; return an empty JSON object.

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
cruhland: Yes, `packagestore` can be used for definitions *internal* to the package store, such as any data structures or code that only that service will use.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Storage.scala
cruhland: @larioj re: O(n) performance -- even with your current implementation, the package array still needs to be traversed to convert it into JSON, so it's already O(n). Let's test it out first and address performance later if needed.

cruhland: @BenWhitehead any suggestions for avoiding `releaseVersion`? We have to provide *some* value for the `releaseVersion` field, since it's required by the Universe v3 schema.

cosmos-json/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/Decoders.scala
jsancio: I agree with @BenWhitehead. This should be `...v1...`

cosmos-json/src/main/scala/com/mesosphere/packagestore/v3/circe/Decoders.scala
```diff
@@ -0,0 +1,22 @@
+package com.mesosphere.packagestore.v3.circe
```
jsancio: Change this to `...v1...`.

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
```diff
@@ -0,0 +1,98 @@
...
+  framework: Option[Boolean] = None,
+  preInstallNotes: Option[String] = None,
+  postInstallNotes: Option[String] = None,
+  postUninstallNotes: Option[String] = None,
+  licenses: Option[List[License]] = None,
+  resource: Option[V2Resource] = None,
+  config: Option[JsonObject] = None,
+  command: Option[Command] = None
+) extends PackageBundle {
+  override def toPackageDefinition(releaseVersion: PackageDefinition.ReleaseVersion): PackageDefinition =
```
jsancio: Same here. Please use `Injection`, `Bijection` or `Convertion`.

cosmos-model/src/main/scala/com/mesosphere/packagestore/v3/model/PackageBundle.scala
```diff
@@ -0,0 +1,98 @@
...
+  postInstallNotes: Option[String] = None,
+  postUninstallNotes: Option[String] = None,
+  licenses: Option[List[License]] = None,
+  minDcosReleaseVersion: Option[DcosReleaseVersion] = None,
+  marathon: Option[Marathon] = None,
+  resource: Option[V3Resource] = None,
+  config: Option[JsonObject] = None,
+  command: Option[Command] = None
+) extends PackageBundle {
+  override def toPackageDefinition(releaseVersion: PackageDefinition.ReleaseVersion): PackageDefinition =
```
jsancio: Same here. Please use `Injection`, `Bijection` or `Convertion`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
jsancio: Yes. Let's create an issue in Github. Explaining exactly what still needs to get done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Storage.scala
jsancio: Let's make an interface that this class implements. Let's also file an issue in github to address moving this out of Cosmos in a later PR. Please leave a TODO comment here that points to that issue.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Storage.scala
BenWhitehead: When the repo is pulled by cosmos the value could be set to the global value, nothing in the universe definition dictates that the values have to be consecutive. We could have one global set of values spread across all packages, but still relative for each package.

BenWhitehead:
@cruhland RE: the release of 0.2.0 I want to wait until the enterprise EA has taken place and cosmos has been tested in that environment. Lots of config around ssl and security that is not accounted for outside enterprise builds.

cosmos-json/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/Encoders.scala
```diff
@@ -22,9 +20,20 @@ object Encoders {
...
+      case v2: V2PackageBundle => v2.asJson
+      case v3: V3PackageBundle => v3.asJson
+    }
+
   implicit val publishResponse: Encoder[PublishResponse] = deriveFor[PublishResponse].encoder
   implicit val publishRequest: Encoder[PublishRequest] = deriveFor[PublishRequest].encoder
 
+  import com.mesosphere.universe.v2.circe.Encoders._
+  import com.mesosphere.universe.v3.circe.Encoders.{encodeLicense => encodeV3License}
+
```
larioj: I made this change because these imports were shadowing the implicit variables

cosmos-json/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/Encoders.scala
```diff
@@ -21,6 +20,20 @@ object Encoders {
...
+      case v2: V2PackageBundle => v2.asJson
+      case v3: V3PackageBundle => v3.asJson
+    }
+
+  implicit val publishResponse: Encoder[PublishResponse] = deriveFor[PublishResponse].encoder
+  implicit val publishRequest: Encoder[PublishRequest] = deriveFor[PublishRequest].encoder
+
+  import com.mesosphere.universe.v2.circe.Encoders._
+  import com.mesosphere.universe.v3.circe.Encoders.{encodeLicense => encodeV3License}
+
```
larioj: These imports were shadowing implicit variables

cruhland: Which implicits? Could they be renamed instead to avoid name clashes?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -82,6 +83,13 @@ private[cosmos] final class Cosmos(
     route(post("package" / "repository" / "delete"), deleteRepositoryHandler)(RequestReaders.standard)
   }
 
+  // TODO: @larioj [jlariosmurillo@mesosphere.io] this will need to change: see Issue #463
```
cruhland: Let's just take care of this now? Just declare two new arguments to the `Cosmos` constructor and move the `new LocalPackageStorage` to `Cosmos.apply`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -0,0 +1,5 @@
+package com.mesosphere.cosmos.storage
+
+trait PackageStorage {
+  //TODO: Issue #464
```
cruhland: For now, put the method signatures from `LocalPackageStorage` in this trait? We can change them later, I just don't want to commit incomplete definitions.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackagePublishHandler.scala
```diff
@@ -0,0 +1,18 @@
+package com.mesosphere.cosmos.handler
+
+import com.mesosphere.cosmos.http.RequestSession
+import com.mesosphere.cosmos.rpc.v1.model.{PublishRequest, PublishResponse}
+import com.mesosphere.cosmos.storage.LocalPackageStorage
+import com.twitter.util.Future
+
+private[cosmos] final class PackagePublishHandler(
+  storage: LocalPackageStorage
```
cruhland: Just use the `PackageStorage` trait here, after addressing [this comment](https://github.com/dcos/cosmos/pull/462/files#r75762625).

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/RepositoryServeHandler.scala
```diff
@@ -0,0 +1,17 @@
+package com.mesosphere.cosmos.handler
+
+import com.mesosphere.cosmos.http.RequestSession
+import com.mesosphere.cosmos.storage.LocalPackageStorage
+import com.mesosphere.universe.v3.model.Repository
+import com.twitter.util.Future
+
+private[cosmos] final class RepositoryServeHandler(
+  storage: LocalPackageStorage
```
cruhland: Just use the `PackageStorage` trait here, after addressing [this comment](https://github.com/dcos/cosmos/pull/462/files#r75762625).

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalPackageStorage.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.twitter.bijection.Conversion.asMethod
+
+final class LocalPackageStorage extends PackageStorage {
+  private[this] val bundlesByName = new ConcurrentHashMap[String, ConcurrentLinkedQueue[PackageBundle]]()
+
+  def getRepository: Repository = {
+    val packages =
+      bundlesByName
+        .values()
+        .toArray(Array[ConcurrentLinkedQueue[PackageBundle]]())
```
cruhland: You should be able to use the `JavaConverters` for this instead:

```scala
import scala.collection.JavaConverters._

bundlesByName
  .values()
  .asScala
  .flatMap { // ...
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalPackageStorage.scala
```diff
@@ -0,0 +1,41 @@
...
+        .flatMap { bundles: ConcurrentLinkedQueue[PackageBundle] =>
+          bundles
+            .toArray(Array[PackageBundle]())
```
cruhland: Same here? Use `JavaConverters.asScala`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalPackageStorage.scala
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  def putPackageBundle(packageBundle: PackageBundle): Unit = {
+    val name = packageBundle match {
+      case v2: V2PackageBundle => v2.name
+      case v3: V3PackageBundle => v3.name
+    }
+    bundlesByName.putIfAbsent(name, new ConcurrentLinkedQueue[PackageBundle]())
+    bundlesByName.get(name).add(packageBundle)
+    ()
```
cruhland: This should be a little bit safer, and it only looks up `name` in the map once:

```scala
val emptyBundles = new ConcurrentLinkedQueue[PackageBundle]()
val bundles = Option(bundlesByName.putIfAbsent(name, emptyBundles)) match {
  case Some(b) => b
  case _ => emptyBundles
}
bundles.add(packageBundle)
```

cruhland:
@larioj should be good to merge after my comments are addressed. Also, forget -- did we decide that a test should be written for this PR or not? It needs to get written eventually.

cosmos-json/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/Encoders.scala
BenWhitehead: Agreed, since this isn't actually changing a lexical scope it would be better to ensure that things are explicit about removing collision rather than expecting the re-order to continue working.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PublishRequest.scala
```diff
@@ -0,0 +1,3 @@
+package com.mesosphere.cosmos.rpc.v1.model
+
+case class PublishRequest(packageBundle: PackageBundle)
```
BenWhitehead: I'm leery of having the `PublishRequest` use a type not defined by Universe. This will result in a cyclic dependency between the model of universe and the model of cosmos, where I think cosmos is always a client of Universe and Universe knows nothing specific about cosmos.

I realize there is a potential complication here since `PackageBundle` removes the `releaseVersion` property from the document  but I think we should try and avoid the cycle.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
BenWhitehead: Agreed, fix this before the merge.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalPackageStorage.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.concurrent.{ConcurrentHashMap, ConcurrentLinkedQueue}
+
+import com.mesosphere.cosmos.converter.Common.BundleToPackageDefinition
+import com.mesosphere.cosmos.rpc.v1.model.{PackageBundle, V2PackageBundle, V3PackageBundle}
+import com.mesosphere.universe.v3.model.{PackageDefinition, Repository}
+import com.twitter.bijection.Conversion.asMethod
+
+final class LocalPackageStorage extends PackageStorage {
+  private[this] val bundlesByName = new ConcurrentHashMap[String, ConcurrentLinkedQueue[PackageBundle]]()
```
BenWhitehead: Read from storage is by far the much more common case of use, so I think we should prefer to move more of the complex logic into the write operation in favor of a faster read operation.

Also, having nested concurrent data structures isn't necessarily what we want here, since the scope of changes isn't being made atomic (since there are two data structures, each of which is safe in isolation, but not necessarily together).

This class also doesn't have any guards in place to error to a user when multiple users try and publish the same package at the same time.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalPackageStorage.scala
larioj: The toArray method is not for converting to Scala. It is for converting the concurrent queue to an array since we can manipulate an array easier. The toArray method is provided by the queue. I don't know if there is a Scala conversion for concurrent queues.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalPackageStorage.scala
larioj: Yeah, this is a better way to do it. Thanks for the tip. I forgot that Option produces None if given null.

larioj:
@cruhland So there are tests written, but I was gonna push them in a separate PR. But I'm fine with adding them to this PR. Either way is fine by me.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PublishRequest.scala
larioj: Could we move packageBundle to Universe. This would make sense since at some point Universe will have to deal with bundles.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalPackageStorage.scala
larioj: @BenWhitehead @cruhland This is an extension of the discussion we had on the previous PR. See the outdated diff above that has seven comments. I could change it back to the version that used a single counter and did the transformation up front.

In regards to "This class also doesn't have any guards in place to error to a user when multiple users try and publish the same package at the same time." It is impossible for a user to publish the same package twice, since each would get a different release version.

BenWhitehead: There would be a different release version, yes, but possibly a different "latest" in the case that the same package version is used.

larioj: Yes, it may be the case that if Alice publishes Kafka 1.0, and Bob publishes Kafka 1.0 at around the same time, then both try to run the latest Kafka one of them will see the one that got written last not necessarily the package they published. But I don't see a way to solve this problem with the current endpoints. The problem boils down to the fact that publish and run are not atomic. They are separate operations. Anything might have occurred between the publish and the run. The only solution I see to this problem is to provide an endpoint that both installs and runs the package, that way we can guarantee that the operations are atomic.

BenWhitehead: I'm not advocating publish and install be atomic, I advocating telling the user that their publish failed because another publish happened at the same time.

If the user is publishing packages with specific versions then they can disambiguate between each other, but if they aren't then we should provide some info so that they know the state of the repo has shifted since the last time they looked at it.

We do this today when configuring the set of repos cosmos will track, we use a "compare and set" approach and error if the comparison fails. Java's [`AtomicReferece#compareAndSet`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicReference.html#compareAndSet-V-V-) can get us a similar semantic in memory as compared to zk.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -0,0 +1,8 @@
+package com.mesosphere.cosmos.storage
+
+import com.mesosphere.universe.v3.model.{PackageBundle, Repository}
+
+trait PackageStorage {
+  def getRepository: Repository
+  def putPackageBundle(packageBundle: PackageBundle): Unit
```
cruhland: Let's wrap the return values of these methods in `Future`, since going forward they will be accessing external resources.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InMemoryPackageStorage.scala
```diff
@@ -0,0 +1,25 @@
...
+    Repository(
+      packages.get.toList
+    )
+
+  override def putPackageBundle(packageBundle: PackageBundle): Unit = {
+    val oldp = packages.get
+    val newp = oldp :+ (packageBundle, PackageDefinition.ReleaseVersion(oldp.size).get).as[PackageDefinition]
+
+    if(packages.compareAndSet(oldp, newp))
+      throw RepositoryChangedDuringPublish()
```
cruhland: Syntax nitpick: always use curly braces with control flow blocks, even if they are just one line.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -107,6 +107,8 @@ object Encoders {
   }
 
   private[this] def msgForCosmosError(err: CosmosError): String = err match {
+    case RepositoryChangedDuringPublish() =>
+      "Repository Changed During Publish. Please try again."
```
cruhland: This message made me realize that if two packages with different names are published at the same time, we'll still return this error. But we really only need to fail when two packages with _the same name and version_ are published. @BenWhitehead what do you think? A more complex data structure would be required.

cruhland:
@larioj almost there, just a few comments.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
larioj: We can do it on just name by keeping a map of counters.

larioj: @cruhland it makes sense to do it on just name since the "what's the latest" issue is present regardless of version

cruhland: @larioj ah, right. Name should be fine.

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
```diff
@@ -41,6 +41,8 @@ sealed abstract class CosmosError(causedBy: Throwable = null /*java compatibilit
   def getHeaders: Map[String, String] = Map.empty
 }
 
+case class RepositoryChangedDuringPublish() extends CosmosError
```
BenWhitehead: This error should probably set the status value to `409 Conflict` since that is what the error is, it's not that the request the user sent is malformed, just that the state changed over the course of the request.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -82,6 +85,14 @@ private[cosmos] final class Cosmos(
     route(post("package" / "repository" / "delete"), deleteRepositoryHandler)(RequestReaders.standard)
   }
 
+  val packagePublish: Endpoint[Json] = {
+    route(post("package" / "publish"), packagePublishHandler)(RequestReaders.standard)
+  }
+
+  val repositoryServe: Endpoint[Json] = {
+    route(get("package" / "storage" / "repository"), repositoryServeHandler)(RequestReaders.noBody)
+  }
```
BenWhitehead: I don't think we want this endpoint in here for the first release, I can see the use for it testing/debugging right now but I think we need to remove this as soon as is practical.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
BenWhitehead: I think we're going to want to have a more robust/complex data structure to represent how to merge multiple repositories together, but is probably outside the scope of this PR.

BenWhitehead:
Any tests that you have I would like to see added in this PR, they can be expanded and refined more in the future but I'd prefer to not merge new functionality void of tests.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
larioj: @BenWhitehead This endpoint is necessary for the functionality. This is the URL that the user needs to add to  the list of repos to get access to local storage.

BenWhitehead: What URL does the user add? this outlines the path but not necessarily the ip/port.

larioj:
@BenWhitehead @cruhland PR 3 changes have been made please review and merge if it looks good

cosmos-server/src/it/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
```diff
@@ -113,9 +123,13 @@ object CosmosIntegrationTestClient extends Matchers {
...
       requestMediaType: MediaType,
       responseMediaType: MediaType,
-      status: Status = Status.Ok
+      status: Status = Status.Ok,
+      method: String = "POST"
     )(implicit decoder: Decoder[Res], encoder: Encoder[Req]): Xor[ErrorResponse, Res] = {
-      val request = buildPost(path, requestBody, requestMediaType, responseMediaType)
+      val request =
+        if(method == "POST") buildPost(path, requestBody, requestMediaType, responseMediaType)
+        else buildGet(path, responseMediaType)
```
cruhland: As I said earlier, I'm okay with using a `String` for the HTTP method for now, but let's throw an error if `method` is not `POST` or `GET`.

cosmos-server/src/test/scala/com/mesosphere/cosmos/InMemoryPackageStorageSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+  "The repository must contain all published packages in any order" - {
+    val storage = new InMemoryPackageStorage()
+    val bundles: List[(BundleDefinition, PackageDefinition)] = (0 to 8).map { i =>
+      val (b, _) = TestUtil.MaximalV3ModelV2PackageDefinition.as[(V2Bundle, ReleaseVersion)]
+      val bundle = b.copy(name = "package" + (i / 3))
+      val pkg = (bundle, ReleaseVersion(i % 3).get).as[V2Package]
+      (bundle, pkg)
+    }.toList
+
+    "when no packages have been uploaded" - {
```
cruhland: The blocks of code that contain test assertions should use `in` instead of the hyphen in order for errors to be reported correctly:

```scala
"when no packages have been uploaded" in {
```

cosmos-server/src/test/scala/com/mesosphere/cosmos/InMemoryPackageStorageSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.scalatest.FreeSpec
+import com.twitter.bijection.Conversion.asMethod
+import com.twitter.util.Await
+import com.mesosphere.cosmos.converter.Common._
+
+final class InMemoryPackageStorageSpec extends FreeSpec {
+
+  "The repository must contain all published packages in any order" - {
+    val storage = new InMemoryPackageStorage()
+    val bundles: List[(BundleDefinition, PackageDefinition)] = (0 to 8).map { i =>
```
cruhland: These variables are shared between tests:

- `storage` is stateful, and by making use of that, your tests are not independent (they require a specific execution order). This is usually a bad thing because it makes it harder to change the tests. You should generate a new `InMemoryPackageStorage` for each test.
- `bundles` is a constant, and doesn't depend on any definitions in `FreeSpec`, so putting it in a companion object might be a good idea.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageStorageIntegrationSpec.scala
```diff
@@ -0,0 +1,87 @@
...
+  "The package storage repository will serve a repository that" - {
+    "contains all published packages in any order" - {
+
+
+      val bundles: List[(BundleDefinition, PackageDefinition)] = (0 to 8).map { i =>
+        val (b, _) = TestUtil.MaximalV3ModelV2PackageDefinition.as[(V2Bundle, ReleaseVersion)]
+        val bundle  = b.copy(name = "package" + (i / 3))
+        val pkg = (bundle , ReleaseVersion(i % 3).get).as[V2Package]
+        (bundle , pkg)
+      }.toList
```
cruhland: This looks the same as the data in your other test class. Can you have one test reference the definition from the other, or factor it out into another place like `TestUtil`?

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageStorageIntegrationSpec.scala
```diff
@@ -0,0 +1,87 @@
...
+          assert(getRepo.packages.contains(pkg))
+        }
+      }
+
+      "such that for all packages with the same name, the release version " +
+        "ordering matches the publishing ordering" - {
+        val expected = bundles.map(_._2).sortBy(nameAndRelease)
+        val actual = getRepo.packages.sortBy(nameAndRelease)
+        assertResult(expected)(actual)
+      }
```
cruhland: Most of the above test cases are already being covered by the `InMemoryPackageStorageSpec`. Since this is an integration test, it should focus on testing whether the combined system of components works correctly, under the assumption that the individual components work correctly in isolation. So you probably just need a test that publishes several bundles and verifies that they are all present in the data from the repo endpoint.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageStorageIntegrationSpec.scala
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+  }
+}
```
cruhland: Nitpick: add newline.

cruhland:
@larioj added a few comments on the test code

larioj:
@cruhland thanks :)

cosmos-json/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/Decoders.scala
```diff
@@ -20,6 +22,20 @@ object Decoders {
...
+  implicit val decodeV2PackageBundle = deriveFor[V2Bundle].decoder
+  implicit val decodeV3PackageBundle = deriveFor[V3Bundle].decoder
+  implicit val decodePackageBundle: Decoder[BundleDefinition] = {
+    Decoder.instance { (hc: HCursor) =>
+      hc.downField("packagingVersion").as[PackagingVersion].flatMap {
+        case V2PackagingVersion => hc.as[V2Bundle]
+        case V3PackagingVersion => hc.as[V3Bundle]
+      }
+    }
+  }
```
larioj: These definitions need to be moved to universe. Since universe now owns Bundle.

cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Encoders.scala
```diff
@@ -8,6 +8,14 @@ import io.circe.syntax._
 
 object Encoders {
 
+  implicit val encodePackageBundle: Encoder[BundleDefinition] =
+    Encoder.instance {
+      case v2: V2Bundle => v2.asJson
+      case v3: V3Bundle => v3.asJson
+    }
+  implicit val encodeV2PackageBundle: Encoder[V2Bundle] = deriveFor[V2Bundle].encoder
+  implicit val encodeV3PackageBundle: Encoder[V3Bundle] = deriveFor[V3Bundle].encoder
```
larioj: These too have to move to universe.

larioj:
@cruhland @BenWhitehead It's ready. Hopefully this is the last one.

cruhland:
LGTM! @BenWhitehead ?

PR #477, opened 2016-08-26
Update to circe 0.5.0-M3 and finch 0.10.0

BenWhitehead:
* Replace manual map encode/decode to use new KeyEncoder/KeyDecoder functionality in circe
* Ordering of properties changed between versions of circe and some tests had to be updated due to the new ordering
* There is a new flag in HistoryOp so some tests had to be updated for new error message format
* All finch request readers have been replaced with applicative endpoint validation


.gitignore
```diff
@@ -77,3 +77,4 @@ project/boot/
 .cache
 
 dcos-image/target/
+tmp/
```
cruhland: When is this created?

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -75,7 +103,7 @@ object Encoders {
    * message, which could contain information that we don't want to surface to users in ordinary
    * error responses. It's better to make a conscious decision to include error data when needed.
    */
-  implicit def dropThrowableFromEncodedObjects[K, H <: Throwable, T <: HList](implicit
+  implicit def dropThrowableFromEncodedObjects[K <: Symbol, H <: Throwable, T <: HList](implicit
```
cruhland: Ah, this is leftover from my "guessing" earlier. Should be safe to revert.

cruhland:
:ship: it

Issue #478, opened 2016-08-29
Mechanism for notifying storage change


jsancio:
We need to implement a mechanism for notifying all the Cosmos that the local repository has change. We are going to use this to determine if Cosmos should perform a read against the S3 storage.

```
trait DistributedVar {
  def get: Future[String]
  def put(value: String): Future[Unit]
}
```

Curator recipe: https://curator.apache.org/curator-recipes/node-cache.html

larioj:
We have decided to make distributed var generic.

We are changing the interface to be:
```scala
trait DistributedVar[T] {
  def get: Future[T]
  def put(value: T): Future[T]
}
```

In order to implement distributed var we need some infrastructure
1. We need a way to decode and encode the type T that is stored
2. We need to keep any state that we externalize versioned
3. We need a ZookeeperDriver that will provide us with zookeeper CRUD operations

All of the above infrastructure exists in Cosmos already, but it is tied to ZkRepositoryList. This suggests that the first steps towards the implementation of distributed var is to refactor the functionality so that it can be used to satisfy 1, 2 and 3 above.

This also suggests a decomposition into 3 PRs.
1. Refactor MediaTypedDecoder to support the need to associate just an Circe.Decoder and a MediaType. Currently it associates a finch.DecodeRequest with a mediaType, which is inconsistent with it's name. See Issue #500 for more info
2. ZkRepositoryList keeps elements versioned using an envelope. This functionality should be factored out and generalized to accommodate types beyond the repo list. See Issue #501 for more info.
3. ZkRepositoryList has some zookeeper operations that can be factored. These operation can be leveraged in Distributed var. See issue #502.

After the above refactoring has been completed, implementation of distributed var can begin.

Note that the PRs outlined above will happen against master while the distributed var PR will occur against distributed-publish. The reason for the master PRs is because PR 1 2 and 3 are changing existing components. While the distributed var PR will be adding new functionality. Making the PRs to master will make it easier to merge down the line.

jsancio:
Migrated.

PR #482, opened 2016-08-31
Add implementation/tests for distributed var

larioj:


cruhland:
@larioj did you include tests?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,67 @@
...
+import com.netaporter.uri.Uri
+import com.twitter.util.{Future, Promise}
+import org.apache.curator.framework.CuratorFramework
+import org.apache.curator.framework.api.{BackgroundCallback, CuratorEvent, CuratorEventType}
+import org.apache.zookeeper.KeeperException
+
+class DistributedVar private(zkClient: CuratorFramework, path: Uri) {
+  import DistributedVar._
+
+  def get: Future[String] = {
```
BenWhitehead: Can we not use `String` as the type here, I doubt we're going to only be wanting to store Strings.

ZooKeeperStorage alludes to how this can be done using encoders and decoders to serialize objects into an envelope.

See https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/ZooKeeperStorage.scala#L192-L228

This will allow us to have much more robust handling of types.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,67 @@
...
+      promiseHandler(
+        promise,
+        CuratorEventType.CREATE,
+        Set(KeeperException.Code.OK, KeeperException.Code.NODEEXISTS)
+      )(_ => new DistributedVar(zkClient, path))
+    ).forPath(path.toString(), "".getBytes())
+    promise
+  }
+
+  private def promiseHandler[T](
```
BenWhitehead: Could this code leverage the existing Create/Read/Write handlers in ZooKeeperStorage? Or, if this is accomplishing the same thing can those handlers be cleaned up to have less code?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,67 @@
...
+    val promise = Promise[String]()
+    zkClient.setData().inBackground(
+      promiseHandler(
+        promise,
+        CuratorEventType.SET_DATA
+      )(_ => value)
+    ).forPath(path.toString(), value.getBytes)
+    promise
+  }
+}
```
BenWhitehead: What is the design rationale behind having this class have a direct reference to the zk client rather than there being something (perhaps ZooKeeperStorage) that knows how to manage variables?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,67 @@
...
+
+object DistributedVar {
+  def apply(zkClient: CuratorFramework, path: Uri): Future[DistributedVar] = {
```
jsancio: This method should return a `DistributedVar` not a `Future[DistributedVar]`. Take a look at `ZooKeeperStorage` to see how we achieve this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
larioj: @BenWhitehead Yes, we had a conversation about this and where we decided to make it Strings and later on generalize it. @jsancio Is now the time to generalize?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
larioj: @jsancio I did look into that. It's a lot more complicated, and I didn't see the advantages. What are the advantages of having it not return a future?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
larioj: @BenWhitehead This did not come up while we were doing the design. My post-design rationale for this is that it's simpler. What are the advantages of having a single thing manage all the variables?

larioj:
@cruhland Yes, I do have tests. The problem is that they depend on zookeeper (they would fail the PR). Jose and I had a conversation about this a couple of days ago where we decided to test at integration time (this was one of the reasons why distributed publish branch was created). The tests are here https://github.com/larioj/cosmos/blob/distvar/cosmos-server/src/test/scala/com/mesosphere/cosmos/DistributedVarSpec.scala

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
jsancio: We talked about it offline. @larioj is going to look into returning a `DistributedVar` instead of a `Future[...]`.

cruhland:
@larioj ok, let's include the tests but mark them as ignored, with a TODO comment or issue so that we remember to enable them later. See the "Ignored Tests" heading in the [documentation](http://doc.scalatest.org/3.0.0/index.html#org.scalatest.FreeSpec).

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,67 @@
+package com.mesosphere.cosmos.storage
+
+import com.netaporter.uri.Uri
+import com.twitter.util.{Future, Promise}
+import org.apache.curator.framework.CuratorFramework
+import org.apache.curator.framework.api.{BackgroundCallback, CuratorEvent, CuratorEventType}
+import org.apache.zookeeper.KeeperException
+
+class DistributedVar private(zkClient: CuratorFramework, path: Uri) {
```
cruhland: Always default to making a class `final` unless there's a good reason to allow subclassing.

cruhland: Is `Uri` the right type to use for the ZK path? It could potentially have components like the scheme which ZK won't understand.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,67 @@
...
+  import DistributedVar._
+
+  def get: Future[String] = {
+    val promise = Promise[String]()
+    zkClient.getData.inBackground(
+      promiseHandler(
+        promise,
+        CuratorEventType.GET_DATA
+      )(event => new String(event.getData))
```
cruhland: This `String` constructor decodes the `byte[]` from `getData` using the JVM's default charset, which isn't guaranteed to be UTF-8. It's always better to pass an explicit `Charset` to the constructor:

```scala

new String(event.getData, StandardCharsets.UTF_8)
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  def set(value: String): Future[String] = {
+    val promise = Promise[String]()
+    zkClient.setData().inBackground(
+      promiseHandler(
+        promise,
+        CuratorEventType.SET_DATA
+      )(_ => value)
+    ).forPath(path.toString(), value.getBytes)
```
cruhland: As per my comment on charsets above, this should be `value.getBytes(StandardCharsets.UTF_8)`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
cruhland: @larioj I think what Ben is referring to is that currently, `ZooKeeperStorage` has two responsibilities:

1. Define business-logic operations on the repo list (`read`, `readCache`, `add`, `delete`). These take in specific, meaningful arguments and return `Future[List[PackageRepository]]`.
1. Define low-level CRUD operations on znodes (`create`, `readFromZooKeeper`, `readFromCache`, `write`). These arguably should receive generic data as arguments and return similarly generic data, although right now they are specialized to `List[PackageRepository]`.

If we factor out the repo list stuff into another class, say `ZkRepoList`, and then have `ZooKeeperStorage` perform only generic data operations, then both `ZkRepoList` and your `DistributedVar` could be implemented on top of `ZooKeeperStorage`. This would simplify all three classes and make it easier to store new things in ZooKeeper.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,67 @@
...
+object DistributedVar {
+  def apply(zkClient: CuratorFramework, path: Uri): Future[DistributedVar] = {
+    val promise = Promise[DistributedVar]()
+    zkClient.create().inBackground(
+      promiseHandler(
+        promise,
+        CuratorEventType.CREATE,
+        Set(KeeperException.Code.OK, KeeperException.Code.NODEEXISTS)
+      )(_ => new DistributedVar(zkClient, path))
+    ).forPath(path.toString(), "".getBytes())
```
cruhland: Another `getBytes` call that should have a `Charset` (for consistency).

larioj:
#478

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/Etag.scala
```diff
@@ -0,0 +1,3 @@
+package com.mesosphere.cosmos.storage
+
+sealed case class Etag(value: String)
```
jsancio: You want `final case class` instead of `sealed ...`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+
+}
+
+object DistributedVar {
+  def apply[T](
+    client: CuratorFramework, path: String, zero: T
+  )( implicit mediaTypedEncoder: MediaTypedEncoder[T],
+     mediaTypedDecoder: MediaTypedDecoder[T] )
+  : DistributedVar[T] = {
+    new DistributedVar[T](client, path, zero)
```
jsancio: You should be able to just do `new DistributedVar(client, path, zero)`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/MediaTypedDecoder.scala
```diff
@@ -0,0 +1,13 @@
+package com.mesosphere.cosmos.storage
+
+import com.mesosphere.cosmos.http.MediaType
+import io.circe.Decoder
+
+/** Associates a media type with an [[io.circe.Decoder]] instance. */
+final class MediaTypedDecoder[A] private(val decoder: Decoder[A], val mediaType: MediaType)
```
jsancio: We have a similar class in `com.mesosphere.cosmos.circe` the main different is that this takes a `Decoder` while the other one takes a `DecodeRequest`. Is there a way we can unify the two? cc @cruhland @BenWhitehead 

Maybe what we want is:

1. MediaTypedDecoder which just associates a `circe.Decoder` with a `MediaType`
1. MediaTypedRequestDecoder which associates a `finch.DecodeRequest` with a `MediaType`
1. MediaTypedEnvelopeDecoder which associates an decoder for `model.ZooKeeperStorageEnvelope` (we should rename this to `storage.Envelope` we can call this type `storage.EnvelopeDecoder` to a `MediaType`.

We maybe able to construct 2. and 3. from 1.. What do we think?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/MediaTypes.scala
```diff
@@ -0,0 +1,14 @@
+package com.mesosphere.cosmos.storage
+
+import com.mesosphere.cosmos.http.{MediaType, MediaTypeSubType}
+
+object MediaTypes {
+  val EtagMediaType = MediaType(
```
jsancio: Let's keep this in `com.mesosphere.cosmos.http.MediaTypes` for now. Just add a section through comments.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/MediaTypedDecoder.scala
larioj: @jsancio I agree with this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/MediaTypedEncoder.scala
```diff
@@ -7,9 +7,7 @@ import io.circe.Encoder
 final class MediaTypedEncoder[A] private(val encoder: Encoder[A], val mediaType: MediaType)
 
 object MediaTypedEncoder {
-
```
BenWhitehead: Please remove this file from the commits, it's only whitespace change.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/MediaTypedEncoder.scala
```diff
@@ -7,9 +7,7 @@ import io.circe.Encoder
...
   def apply[A](encoder: Encoder[A], mediaType: MediaType): MediaTypedEncoder[A] = {
     new MediaTypedEncoder(encoder, mediaType)
   }
-
```
BenWhitehead: Please remove this file from the commits, it's only whitespace change.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+        case code => promise.setException(KeeperException.create(code, event.getPath))
+      }
+    }).forPath(path, encodeData[T](zero))
+    promise
+  }
+
+  private def getIfNone(vm: Option[T]): Future[T] = vm match {
+    case Some(value) => Future.value(value)
+    case None => getUnsafe
+  }
```
jsancio: This is same as `vm.map(Future.value(_)).getOrElse(getMaybe)`. We may be able to remove this method since we only call it once.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/Decoders.scala
```diff
@@ -0,0 +1,8 @@
+package com.mesosphere.cosmos.storage
```
BenWhitehead: please move into package `com.mesosphere.cosmos.storage.circe`

All circe/json specific things should be in their own package.  See `com.mesosphere.universe` or `com.mesosphere.cosmos.rpc` for an example of this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/Encoders.scala
```diff
@@ -0,0 +1,8 @@
+package com.mesosphere.cosmos.storage
```
BenWhitehead: please move into package `com.mesosphere.cosmos.storage.circe`

All circe/json specific things should be in their own package.  See `com.mesosphere.universe` or `com.mesosphere.cosmos.rpc` for an example of this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/Etag.scala
```diff
@@ -0,0 +1,3 @@
+package com.mesosphere.cosmos.storage
```
BenWhitehead: please move into package `com.mesosphere.cosmos.storage.model`

All model objects should be in their own package.  See `com.mesosphere.universe` or `com.mesosphere.cosmos.rpc` for an example of this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/MediaTypedEncoder.scala
larioj: OK

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/MediaTypedEncoder.scala
larioj: OK

cosmos-server/src/test/scala/com/mesosphere/cosmos/ZookeeperStorageSpec.scala
```diff
@@ -0,0 +1,15 @@
+package com.mesosphere.cosmos
```
BenWhitehead: The spec file should be in the same package as the source it is testing.

please move into package `com.mesosphere.cosmos.storage`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ZookeeperStorage.scala
```diff
@@ -0,0 +1,56 @@
...
+import com.mesosphere.cosmos.circe.Encoders._
+import com.mesosphere.cosmos.circe.Decoders._
+import com.mesosphere.cosmos.circe.MediaTypedEncoder
+import com.mesosphere.cosmos.http.{MediaType, MediaTypeOps}
+import com.mesosphere.cosmos.model.ZooKeeperStorageEnvelope
+import com.mesosphere.universe.common.ByteBuffers
+import io.circe.jawn.decode
+import io.circe.syntax._
+
+object ZookeeperStorage {
```
BenWhitehead: `s/Zookeeper/ZooKeeper/g`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+      valueMaybe <- getMaybe
+      _ <- createIfNone(valueMaybe)
+      v <- getIfNone(valueMaybe)
+    } yield v
+
+  def set(value: T): Future[T] = for {
+    valueMaybe <- setMaybe(value)
+    _ <- createIfNone(valueMaybe)
+    v <- setIfNone(value, valueMaybe)
+  } yield v
```
jsancio: Let's talk about these two methods (`set` and `get`) in person. Please schedule 30 min meeting.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+      private[this] val logger = org.slf4j.LoggerFactory.getLogger(getClass)
+      override def processResult(client: CuratorFramework, event: CuratorEvent): Unit = {
+        if (event.getType == eventType) {
+          handle(event)
+        } else {
+          logger.error(s"Handler of type $eventType called for incorrect event type ${event.getType}")
+        }
+      }
+    }
+  }
```
BenWhitehead: Proposal, instead use the following def for `handler`
```scala
  private def handler(handle: PartialFunction[CuratorEvent, Unit]) = {
    new BackgroundCallback {
      override def processResult(client: CuratorFramework, event: CuratorEvent): Unit = {
        handle(event)
      }
    }
  }
```

This gives the ability to the user to more interestingly handle the different event results rather than just swallowing and logging something that doesn't match the one defined.

for example `getMaybe` would be re-written as:
```scala
  private def getMaybe: Future[Option[T]] = {
    val promise = Promise[Option[T]]()
    client.getData.inBackground(handler { case event: CuratorEvent if event.getType == CuratorEventType.GET_DATA =>
      KeeperException.Code.get(event.getResultCode) match {
        case KeeperException.Code.OK => promise.setValue(Some(decodeData(event.getData)))
        case KeeperException.Code.NONODE => promise.setValue(None)
        case code => promise.setException(KeeperException.create(code, event.getPath))
      }
    }).forPath(path)
    promise
  }
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+
+import com.mesosphere.cosmos.circe.MediaTypedEncoder
+import com.mesosphere.cosmos.storage.ZookeeperStorage._
+import com.twitter.util.{Future, Promise}
+import org.apache.curator.framework.CuratorFramework
+import org.apache.curator.framework.api.{BackgroundCallback, CuratorEvent, CuratorEventType}
+import org.apache.zookeeper.KeeperException
+
+final class DistributedVar[T] private(
+  client: CuratorFramework, path: String, zero: T
```
BenWhitehead: Please rename `zero` to `default` to be more clear about what it is used for.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+    case None => getUnsafe
+  }
+
+  private def setIfNone(value: T, vm: Option[T]): Future[T] = vm match {
+    case Some(v) => Future.value(v)
+    case None => setUnsafe(value)
+  }
+
+  private def createIfNone(vm: Option[Any]): Future[Unit] = vm match {
+    case Some(_) => Future.value(())
```
BenWhitehead: `s/Future.value(())/Future.Unit/`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+    case Some(value) => Future.value(value)
+    case None => getUnsafe
+  }
+
+  private def setIfNone(value: T, vm: Option[T]): Future[T] = vm match {
+    case Some(v) => Future.value(v)
+    case None => setUnsafe(value)
+  }
+
+  private def createIfNone(vm: Option[Any]): Future[Unit] = vm match {
```
BenWhitehead: `s/Any/T/`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/Etag.scala
```diff
@@ -0,0 +1,3 @@
...
+
+final case class Etag(value: String)
```
BenWhitehead: An ETag is more than just a string see: https://tools.ietf.org/html/rfc7232#section-2.3

cosmos-server/src/test/scala/com/mesosphere/cosmos/DistributedVarSpec.scala
```diff
@@ -0,0 +1,94 @@
+package com.mesosphere.cosmos
```
BenWhitehead: The spec file should be in the same package as the source it is testing.

please move into package `com.mesosphere.cosmos.storage`

cosmos-server/src/test/scala/com/mesosphere/cosmos/DistributedVarSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+    val zk = CuratorFrameworkFactory.newClient(
+      zkPath,
+      new ExponentialBackoffRetry(1000, Integer.MAX_VALUE)
+    )
+    zk.start()
+    zk.getZookeeperClient.blockUntilConnectedOrTimedOut()
+    zk
+  }
+
+  "An unset variable should get the empty string" ignore {
```
BenWhitehead: `s/An unset variable should get the empty string/When variable path does not exist default value should be written and returned/`

cosmos-server/src/test/scala/com/mesosphere/cosmos/DistributedVarSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+    val path = "/unset"
+    val dvar = DistributedVar[Etag](zk, path, zero)
+    val exp = zero
+    val act = Await.result(dvar.get)
+    assertResult(exp)(act)
+    zk.delete().forPath(path)
+    ()
+  }
+
+  "The user should be able to create a variable with a path that already exists in zookeeper" ignore {
```
BenWhitehead: `s/The user should be able to create a variable with a path that already exists in zookeeper/When variable path does exist the value should be decoded and returned/`

cosmos-server/src/test/scala/com/mesosphere/cosmos/DistributedVarSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+    val zk = startZk()
+    val path = "/exists"
+    val exp = Etag("foo")
+    zk.create().forPath(path, encodeData(exp))
+    val dvar = DistributedVar[Etag](zk, path, zero)
+    val act = Await.result {
+      for {
+        r <- dvar.get
+      } yield r
```
BenWhitehead: Personal style pref, avoid for comprehensions when only one map is being performed. Instead just map 
```scala
val act = Await.result(dvar.map(_.get))
```

cosmos-server/src/test/scala/com/mesosphere/cosmos/DistributedVarSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+    assertResult(exp)(act)
+    zk.delete().forPath(path)
+    ()
+  }
+
+  "The user should be able to create multiple variables with the same path" ignore {
```
BenWhitehead: Do we actually want this to happen? 
This sounds like something that would lead to coordination issues between the instances of the var and would force syncing with zk for every read (which can be very costly.)

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/MediaTypedDecoder.scala
BenWhitehead: Refactoring to remove the duplication would be preferable.

BenWhitehead: I'm not sure that we need number 3, because for the envelope shouldn't need to escape any of the storage stuff.

And `DistributedVar` already takes in an encoder and decoder.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/MediaTypes.scala
larioj: OK

BenWhitehead:
@larioj Is this PR superseded by #494 #495 #496 and #497? If so, please close this PR.

larioj:
@BenWhitehead There is no reason why this PR can't happen here. Now that all of the refactoring has I can get this merged. I'd like to keep the history.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/DistributedVarSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+    val exp = TestClass("foo")
+    zk.create().forPath(path, Envelope.encodeData(exp))
+    val dvar = DistributedVar(zk, path, default)
+    val act = Await.result(dvar.get)
+    assertResult(exp)(act)
+    zk.delete().forPath(path)
+    ()
+  }
+
+  "Multiple variables can share the same path, i.e. be aliased to the same value" ignore {
```
BenWhitehead: Do we actually want this to happen? 
This sounds like something that would lead to coordination issues between the instances of the var and would force syncing with zk for every read (which can be very costly.)

larioj: @BenWhitehead Every cosmos needs a copy of the var, so we need aliasing, even if it is just across different processes.

PR #484, opened 2016-09-01
Update cosmos to support content negotiation

BenWhitehead:
The `Accept` header now respects [Quality Values](https://tools.ietf.org/html/rfc7231#section-5.3.1) and along with it the ability to specify multiple MediaTypes in the header value. When multiple values are specified the highest quality value supported by the server will then be used.

Details:
* Add CompoundMediaType
* Add QualityValue
* Create functionality to calculate the intersection between a `Set[MediaType]` and a `CompoundMediaType`
* Update RequestValidators to use CompoundMediaType for `Accept` header
* Update error type when `Accept` header validation fails to have data properties
* Add integration tests to ensure applicative validation is taking place


cosmos-server/src/main/scala/com/mesosphere/cosmos/http/CompoundMediaTypeOps.scala
```diff
@@ -0,0 +1,31 @@
+package com.mesosphere.cosmos.http
+
+import com.mesosphere.cosmos.http.MediaTypeOps._
+
+class CompoundMediaTypeOps(val cmt: CompoundMediaType) extends AnyVal {
```
jsancio: Did you try making this an implicit value class? http://docs.scala-lang.org/overviews/core/value-classes.html#summary_of_limitations

jsancio: implicit value classes should allow you to remove the `compoundMediaTypeToCompoundMediaTypeOps` method below.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/CompoundMediaTypeParser.scala
```diff
@@ -0,0 +1,37 @@
...
+          .map { mts => CompoundMediaType(mts.toSet) }
+    }
+  }
+
+  private[this] def backfillParams(mts: List[MediaType]): (Map[String, String], List[MediaType]) = {
+    mts match {
+      case Nil => (Map.empty, Nil)
+      case x :: xs =>
+        // walk down the list of media types
+        backfillParams(xs) match {
```
jsancio: Okay but we should avoid unbounded stack based recursions.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/CompoundMediaTypeParser.scala
```diff
@@ -0,0 +1,37 @@
...
+    s.split(',').toList.filterNot(_.trim.isEmpty) match {
+      case Nil => Return(new CompoundMediaType(Set.empty))
+      case mts =>
+        Try.collect(mts.map(MediaType.parse))
+          .map { x => backfillParams(x.toList)._2 }
+          .map { mts => CompoundMediaType(mts.toSet) }
+    }
+  }
+
+  private[this] def backfillParams(mts: List[MediaType]): (Map[String, String], List[MediaType]) = {
```
jsancio: Can you explain this in a comment? Looking at the code, it looks like we are picking the parameters of the first MediaType that has them. It looks like we are also copying the parameters for the first one found to all of the medial types that came before it. Is this correct?

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/CompoundMediaTypeParser.scala
```diff
@@ -0,0 +1,37 @@
...
+
+object CompoundMediaTypeParser {
+
+  def parse(s: String): Try[CompoundMediaType] = {
+    s.split(',').toList.filterNot(_.trim.isEmpty) match {
+      case Nil => Return(new CompoundMediaType(Set.empty))
+      case mts =>
+        Try.collect(mts.map(MediaType.parse))
+          .map { x => backfillParams(x.toList)._2 }
+          .map { mts => CompoundMediaType(mts.toSet) }
```
jsancio: I would write this as one `map`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypeOps.scala
```diff
@@ -19,10 +28,60 @@ object MediaTypeOps {
...
+  }
+
+  /**
+    * Behavior note: right now this function does not differentiate between
+    *  * `mediaType` does not have a q parameter defined and
+    *  * `mediaType` does have a q parameter defined but parsing failed
+    */
+  def qValue(mediaType: MediaType): QualityValue = {
+    QualityValue.getFromMediaType(mediaType).toOption.flatten.getOrElse(QualityValue.default)
+  }
```
jsancio: We can move this to the class `MediaTypeOps` if we turn `class MediaTypeOps` into an `implicit class MediaTypeOps`. http://docs.scala-lang.org/overviews/core/value-classes.html#summary_of_limitations

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/QualityValue.scala
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  implicit val qualityValueOrdering: Ordering[QualityValue] = Ordering.by(_.quality)
+
+  private[this] def validateString(s: String): String = {
+    if (
+      (s.charAt(0) == '0' ||s.charAt(0) == '1')
+        && s.length <= 5
+        && "0.0" <= s && s <= "1.000"
```
jsancio: Did you mean to do this string comparison? If so, can you explain why it works?

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/CompoundMediaTypeOps.scala
```diff
@@ -0,0 +1,31 @@
...
+
+object CompoundMediaTypeOps {
+  import scala.language.implicitConversions
+  import MediaTypeOps.mediaTypeOrdering
+
+  implicit def compoundMediaTypeToCompoundMediaTypeOps(cmt: CompoundMediaType): CompoundMediaTypeOps = {
+    new CompoundMediaTypeOps(cmt)
+  }
+
+  def calculateIntersectionAndOrder(cmt: CompoundMediaType, mts: Set[MediaType]): List[MediaType] = {
```
jsancio: Can you write a comment explaining what this method is doing?

jsancio: I am most concern with the `filter` and `filterNot`.

jsancio:
@BenWhitehead I left you some comments. Please ping me when you want me to look at it again.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/CompoundMediaTypeOps.scala
BenWhitehead: In order for this to be just an implicit class rather than a value class + implicit function `CompoundMediaTypeOps` would have to be nested under some other class/object. But since it's at the top level we get this error from the compiler if we try and make it an `implicit class`:

```
[error] .../cosmos-server/src/main/scala/com/mesosphere/cosmos/http/CompoundMediaTypeOps.scala:5: CompoundMediaTypeOps is already defined as (compiler-generated) method CompoundMediaTypeOps
[error] implicit class CompoundMediaTypeOps(val cmt: CompoundMediaType) extends AnyVal {
[error]                ^
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/CompoundMediaTypeParser.scala
BenWhitehead: Do we want to put a hard cap on it and error if the list is larger than that value?

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypeOps.scala
BenWhitehead: See [this comment](https://github.com/dcos/cosmos/pull/484#discussion_r77395388) for reasoning of why MediaTypeOps isn't an `implicit class`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/QualityValue.scala
BenWhitehead: Yup the string comparison is intentional, I'll add an explanation comment in the code.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/CompoundMediaTypeOps.scala
BenWhitehead: I'll add a comment in the code, but the gist of it is. That if the Accept header specifies a Media Type with a qvalue of `0.0` then it is explicitly stating that it does not ever want that type returned.

So when we're calculating the intersection of valid media types we first filter out all types with a qvalue of zero from the left hand side so that even if they exist on the right had side they won't be returned.

BenWhitehead:
@jsancio Added code comments and replied to questions in review.

jsancio:
LGTM! :shipit:

PR #486, opened 2016-09-06
Update default value for zookeeperUri to be 127.0.0.1

BenWhitehead:
Due to localhost not being available for all cloud providers we've been asked to remove DNS names that aren't provided by dc/os itself.


jsancio:
LGTM. :shipit: Do we need this for 1.8? cc @lingmann

BenWhitehead:
Nope, we don't need it for 1.8 in 1.8 the flag is explicitly set. This is just to update the default.

PR #487, opened 2016-09-07
Add specifc encoder logic for circe errors

BenWhitehead:


cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -12,11 +12,11 @@ import com.mesosphere.universe.v2.circe.Encoders._
 import com.mesosphere.universe.v3.circe.Encoders._
 import com.mesosphere.universe.v3.model._
 import com.twitter.finagle.http.Status
-import io.circe.DecodingFailure
+import io.circe.HistoryOp.opsToPath
```
cruhland: Well, that makes it easy...

cruhland:
LGTM :+1:

PR #488, opened 2016-09-08
Fix compile warnings

BenWhitehead:


cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
```diff
@@ -151,7 +151,7 @@ final class PackageListIntegrationSpec
         assert(packages == packages.sorted)
         assert(names.sorted == resultNames.sorted)
       } finally {
-        installResponses.map(ir => Try(packageUninstall(ir)))
```
cruhland: Wasn't the `Try` there to make sure all of the packages get uninstalled, even if `packageUninstall` throws an exception for one of them?

cruhland:
@BenWhitehead one question, otherwise LGTM

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
BenWhitehead: It would make sure that an attempt was made at all of them, but not verify that the uninstall was successful. My personal opinion is that a test that leaves cruft around is a failed test which is why I removed the `Try`.

cruhland: Ah, good point about not having cruft silently left around. Looks good.

PR #494, opened 2016-09-13
Refactor MediaTypedEncoder/MediaTypedDecoder

larioj:
* The refactor will make MediaTypedEncoder and
  MediaTypedDecoder consistent with each other
  * We want to leverage these types to encode and decode things that are not http requests.
  * Move MediaTypedDecoder to
    MediaTypedRequest
  * Create MediaTypedDecoder that is the
    Decoder version of MediaTypedEncoder


cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/MediaTypedDecoders.scala
```diff
@@ -1,44 +1,44 @@
 package com.mesosphere.cosmos.rpc.v1.circe
 
-import com.mesosphere.cosmos.circe.MediaTypedDecoder
+import com.mesosphere.cosmos.circe.MediaTypedRequestDecoder
 import com.mesosphere.cosmos.http.MediaTypes
 import com.mesosphere.cosmos.rpc.v1.circe.Decoders._
 import com.mesosphere.cosmos.rpc.v1.model._
 import io.finch.circe._
 
 object MediaTypedDecoders {
```
jsancio: Please rename the file and object to `MediaTypedRequestDecoder`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/MediaTypedRequestDecoder.scala
```diff
@@ -0,0 +1,32 @@
+package com.mesosphere.cosmos.circe
```
jsancio: `DecodeRequest` comes from `io.finch` we should change this filename and package to `com.mesosphere.cosmos.finch`.

We talked about not needing this class at all and instead using `MediaTypedDecoder`. What made you change from that plan? Were you not able to do this? Please explain why. Thanks!

larioj:
@jsancio The change that we discussed did not work. It broke six unit tests. I didn't see much value into delving any further given that the difference in quality between the solutions is small

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/MediaTypedEncoder.scala
```diff
@@ -8,6 +8,12 @@ final class MediaTypedEncoder[A] private(val encoder: Encoder[A], val mediaType:
 
 object MediaTypedEncoder {
 
+  def apply[A](mediaType: MediaType)(implicit 
+    encoder: Encoder[A]
+  ): MediaTypedEncoder[A] = {
+    MediaTypedEncoder(encoder, mediaType)
+  }
```
BenWhitehead: This method doesn't appear to be used, why was it added?

BenWhitehead:
Based on [this](https://github.com/dcos/cosmos/pull/482#r78452261) comment in #482 I expected something more like:

```scala
case class MediaTypedDecoder[A](decoder: io.circe.Decoder[A], mediaType: MediaType)
case class MediaTypedResponseDecoder[A](mtd: MediaTypedDecoder[A]) {
  val finchDecoder: io.finch.DecoderRequest[A] = ???
}
```

Why did you end up with the duplication between MediaTypedDecoder and MediaTypedResponseDecoder?

larioj:
* The purpose of this refactoring is to utilize
  the MediaTypedCoders in the serialization of
  classes into storage. Currently they are being
  used only to serialize classes for http requests
  and responses
* Move MediaTypedDecoder to MediaTypedRequestDecoder
  * MediaTypedRequestDecoder is a distinct type from
    MediaTypedDecoder. MediaTypedDecoder associates
    a circe.Decoder with a mediaType while
    MediaTypedRequestDecoder associates a
    finch.DecodeRequest with other metadata.
  * MediaTypedRequestDecoder will use MediaTypedDecoder
    to avoid duplication of work.
* Create MediaTypedDecoder that is the
    Decoder version of MediaTypedEncoder
* The refactor will make MediaTypedEncoder and
  MediaTypedDecoder consistent with each other
  * MediaTypedEncoder will get a method that takes
    parameters implicitly. This is not currently used
    in the code base, but will make the encoder usage
    similar to the decoder usage moving forward.
* Will fix unit tests that currently assume we decoding
  of things other the Json. This change will make it so
  that the MediaTypedCoders can only handle Json

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/MediaTypedEncoder.scala
larioj: It makes this class symmetric with MediaTypedDecoder

larioj:
Closing this PR. This PR should be made to master rather than distributed publish. See the new PR #499

PR #495, opened 2016-09-14
Add LocalRepositoryHash to models

larioj:
* This etag will be used by distributed publish
  to keep track of changes

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
@@ -8,6 +8,10 @@ object MediaTypes {
       Map("charset" -> "utf-8", "version" -> ("v" + version))
     )
 
+  // Storage Media Types
+  val ETag = vnd("repository.etag")
+
+  // RPR Media Types
   val any = MediaType("*", MediaTypeSubType("*"))
   val applicationJson = MediaType("application", MediaTypeSubType("json"), Map("charset" -> "utf-8"))
   val applicationZip = MediaType("application", MediaTypeSubType("zip"))
--- a/cosmos-json/src/main/scala/com/mesosphere/cosmos/storage/v1/circe/Decoders.scala
+++ b/cosmos-json/src/main/scala/com/mesosphere/cosmos/storage/v1/circe/Decoders.scala
@@ -1,0 +1,9 @@
+package com.mesosphere.cosmos.storage.v1.circe
+
+import com.mesosphere.cosmos.storage.v1.model.ETag
+import io.circe.Decoder
+import io.circe.generic.semiauto._
+
+object Decoders {
+  implicit val decodeEtag: Decoder[ETag] = deriveDecoder[ETag]
+}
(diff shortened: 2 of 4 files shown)


BenWhitehead:
An ETag already has a formal spec for how it is/should be serialized in the RFC https://tools.ietf.org/html/rfc7232#section-2.3

Rather than just encoding the case class as json, the ETag should be converted to a string based on the RFC and then that string can be included in a json document.

This will require writing manual encoders/decoders. (uri can be used as an example if you are wanting to see the pattern: [encoder](https://github.com/dcos/cosmos/blob/40923a44a2ea1db9d18252c0569a48e05c3c4563/cosmos-json/src/main/scala/com/mesosphere/universe/common/circe/Encoders.scala#L13-L13)/[decoder](https://github.com/dcos/cosmos/blob/7e6c28a70c0d370555d8c92caf61bd1b3478c385/cosmos-json/src/main/scala/com/mesosphere/universe/common/circe/Decoders.scala#L11-L11).

After you have created the code to parse the header according to the RFC several unit tests should be written to prove various cases and to prevent regressions in the parsing code.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
```diff
@@ -8,6 +8,10 @@ object MediaTypes {
       Map("charset" -> "utf-8", "version" -> ("v" + version))
     )
 
+  // Storage Media Types
+  val ETag = vnd("repository.etag")
```
BenWhitehead: I don't think an ETag should have a distinct media type, I think an etag only make sense in the context of the thing it applies to, and should only ever be written or persisted in another document never on it's own.

Please remove this val.

BenWhitehead: Also, as a general rule, when we're adding new variables or methods to an existing class or object, the new items are usually added at the end of the contextual block (public method kept together, private methods kept together etc) rather than adding them at the beginning.

In the future please append instead of prepend.

larioj: Noted. Will do that from now on.

larioj: @BenWhitehead Our use case requires us to store the etag on its own. Since we are storing it, we need this media type for the envelope.

larioj:
See #478 for why this change is needed

BenWhitehead:
@larioj I stand by my previous comment, having an ETag type is great but I don't think an ETag should be persisted alone. An ETag only makes sense in relation to a specific resource, on its own it makes no sense.

I'm also skeptical that an ETag is an appropriate thing to try and track the state of an entire bucket. ETags are intended to specific resources.

larioj:
#506 has a discussion on this class

larioj:
@BenWhitehead Jose and I decided that what we need here is not actually an ETag, but rather a hash of the local repository. See #506 for a more detailed discussion

PR #499, opened 2016-09-16
Refactor MediaTypedEncoder/MediaTypedDecoder

larioj:
* The purpose of this refactoring is to utilize
  the MediaTypedCoders in the serialization of
  classes into storage. Currently they are being
  used only to serialize classes for http requests
  and responses
* Move MediaTypedDecoder to MediaTypedRequestDecoder
  * MediaTypedRequestDecoder is a distinct type from
    MediaTypedDecoder. MediaTypedDecoder associates
    a circe.Decoder with a mediaType while
    MediaTypedRequestDecoder associates a
    finch.DecodeRequest with other metadata.
  * MediaTypedRequestDecoder will use MediaTypedDecoder
    to avoid duplication of work.
* Create MediaTypedDecoder that is the
    Decoder version of MediaTypedEncoder
* The refactor will make MediaTypedEncoder and
  MediaTypedDecoder consistent with each other
  * MediaTypedEncoder will get a method that takes
    parameters implicitly. This is not currently used
    in the code base, but will make the encoder usage
    similar to the decoder usage moving forward.
* Will fix unit tests that currently assume we decoding
  of things other the Json. This change will make it so
  that the MediaTypedCoders can only handle Json


larioj:
This is a PR to master that is taking over PR #494 to distributed-publish. It is better to make this PR to master since it's changing things rather than adding new functionality.

jsancio:
LGTM. :rocket:

cosmos-server/src/test/scala/com/mesosphere/cosmos/handler/RequestValidatorsSpec.scala
```diff
@@ -121,7 +122,7 @@ final class RequestValidatorsSpec extends FreeSpec {
         .setHeader("Accept", accept.toSeq)
         .setHeader("Authorization", authorization.toSeq)
         .setHeader("Content-Type", MediaTypes.applicationJson.show)
-        .buildPost(Buf.Utf8(Json.Null.noSpaces))
+        .buildPost(Buf.Utf8("\"null\""))
```
jsancio: This is a slight difference between the old code and the new code. The code has the JSON `null` while the new code has the JSON String `"null"`. Is this intentional?

larioj: Json.Null.noSpaces is a string. It's just not formatted correctly since it does not include the double quotes. I looked for something like Json.Null.withDoubleQuotes that I could use instead of the raw string, but I did not find anything.

cosmos-server/src/test/scala/com/mesosphere/cosmos/handler/RequestValidatorsSpec.scala
```diff
@@ -10,6 +10,7 @@ import io.circe.{Encoder, Json}
 import io.finch.{Endpoint, Input, Output}
 import org.scalatest.FreeSpec
 import cats.Eval
+import com.mesosphere.cosmos.finch.MediaTypedRequestDecoder
```
BenWhitehead: In general we try and keep imports sorted lexicographically, can you move this import up with the other mesosphere imports?

larioj: Yes

larioj:
See Issue for details #478

larioj:
@BenWhitehead I have a script that does my PRs. I thought just leaving the commit that addressed the feedback would work for the concerns that you pointed out in the meeting this morning, but apparently it doesn't work with the new review system. Sorry for this time, will try to fix for next time.

cosmos-server/src/test/scala/com/mesosphere/cosmos/handler/RequestValidatorsSpec.scala
larioj: @jsancio So I looked into this some more after our conversation, and I'm convinced that it doesn't matter. The reason why ```Json.Null.noSpaces``` doesn't work is because null is not formatted as a string, and there is no representation for a "null pointer of a string" that the circe decoders can understand i.e. they can understand a string, not the absence of one. Furthermore, the concrete types of the ```Req``` really don't matter because the tests never actually look at the body of the request. The tests only inspect the metadata associated with the request. I still changed the code to expect the type of the request to be Json, that way we get to keep ```Json.Null.noSpaces``` instead of some weirdly quoted string. I do think that the tests are a little less than readable, but I don't want to get stuck in an infinite refactor :)

Issue #501, opened 2016-09-19
Refactor ZkRepositoryList Enveloping


larioj:
Pull out and generalize Enveloping code so that it can be used with distributed var see #478 for high level view of what other changes are happening.

PR #497, opened 2016-09-14
Pulling Out Envelope Functionality

larioj:
Pulling out the enveloping functionality from ZooKeeperStorage

PR #504, opened 2016-09-20
Refactor ZkRepositoryList envelope (see #501)

larioj:
* Extract and generalize enveloping code from ZkRepositoryList
* Create EnvelopeError to mark eveloping errors
* Create media types and coders for repository list

See issue #501

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/Envelope.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/Envelope.scala
@@ -9,7 +9,7 @@ import com.mesosphere.cosmos.http.{MediaType, MediaTypeOps}
 import com.mesosphere.universe.common.ByteBuffers
 import com.mesosphere.cosmos.circe.Decoders._
 import com.mesosphere.cosmos.circe.Encoders._
-import com.mesosphere.cosmos.model.StorageEnvelope
+import com.mesosphere.cosmos.model.ZooKeeperStorageEnvelope
 import io.circe.jawn.decode
 import io.circe.syntax._
 
@@ -19,7 +19,7 @@ object Envelope {
     implicit val encoder = mediaTypedEncoder.encoder
     val mediaType = mediaTypedEncoder.mediaType
     val bytes = ByteBuffer.wrap(data.asJson.noSpaces.getBytes(StandardCharsets.UTF_8))
-    StorageEnvelope(
+    ZooKeeperStorageEnvelope(
       metadata = Map("Content-Type" -> mediaType.show),
       data = bytes
     ).asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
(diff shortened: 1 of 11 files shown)


larioj:
This PR is taking over PR #497. The other PR was to distributed publish branch while this PR is to master. This is more appropriate since the PR is refactoring code, rather than adding new functionality. See Issue #478 for more information

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/Envelope.scala
```diff
@@ -0,0 +1,62 @@
...
+import io.circe.{Decoder, Encoder}
+import io.circe.generic.semiauto._
+import io.circe.jawn.decode
+import io.circe.syntax._
+
+object Envelope {
+
+  private[this] case class StorageEnvelope(metadata: Map[String, String], data: ByteBuffer)
+  private[this] implicit val decodeStorageEnvelope: Decoder[StorageEnvelope] = deriveDecoder
+  private[this] implicit val encodeStorageEnvelope: Encoder[StorageEnvelope] = deriveEncoder
```
BenWhitehead: Please remove these three lines once they are restored in the original files.

larioj: @BenWhitehead Why do you want the envelope to be external?

BenWhitehead: It was already in the external/public location, it can't just be removed/renamed. Everything in the `comsos-model` module should be treated as immutable and append only.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/EnvelopeSpec.scala
```diff
@@ -0,0 +1,59 @@
...
+    val act = decodeData[TestClass](encodeData(exp))
+    assertResult(exp)(act)
+  }
+
+  "Decoding the result of an encoding as a different type should throw an exception" in {
+    val input = TestClass2("fooooo")
+    intercept[EnvelopeError] {
+      decodeData[TestClass](encodeData(input))
+    }
+    ()
```
jsancio: Why do you need this?

larioj: Without it the compiler gives me a warning that I'm ignoring a non-unit value. This makes the warning go away.

jsancio: Okay.

BenWhitehead: Instead of manually returning unit you can essentially tell scala to ignore the result of a function call by doing
```scala
val _ = decodeData[TestClass](encodeData(input))
```

However this trick can only be used once per scope.

larioj: Yeah, I know. I personally like this way better since it's more extendable. I can add a line at any point in the test and not change the val _ = foo. It also communicates intent better.

Issue #490, opened 2016-09-12
Installing queue backed by ZK


jsancio:
#We need to figure out the details here but here are some of the operations we need to support:

This is the API used during install call:
```
trait PackageQueueAdder {
  def add(
    pkg: PackageCoordinate,
    content: PackageQueueContents
  ): Future[PackageAddResult]
}
```

This is the API used by anyone interested in the install queue state. E.g. installer, statue API call:
```
trait PackageQueueReader {
  def readAll(): Future[Option[List[(PackageCoordinate, PackageQueueContents)]]]
  def read(package: PackageCoordinate): Future[Option[PackageQueueContents]]
}
```

This is the API used by the installer:

```
trait PackageQueueRemover {
  def remove(package: PackageCoordinate): Future[PackageRemoveResult]
}
```

```
final class ZooKeeperPackageQueue implements PackageQueueAdder with PackageQueueReader with PackageQueueRemover {
  override def add(...): ...
  override def read(...): ...
  ...
}
```

jsancio:
Migrated.

PR #507, opened 2016-09-20
Install queue backed by zk

tamarrow:
install queue interface: #490. Once #504 is merged I will use it to store znode data in envelopes.


cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedAtomicCounter.scala
```diff
@@ -0,0 +1,50 @@
...
+    Stat.timeFuture(stats.stat("call")) {
+      pool {
+        val newValue = counter.increment()
+
+        val newStats = newValue.getStats()
+        stats.stat("optimisticTries").add(newStats.getOptimisticTries().asInstanceOf[Float])
+        stats.stat("optimisticTriesMs").add(newStats.getOptimisticTimeMs().asInstanceOf[Float])
+
+        if (newValue.succeeded()) {
+          newValue.postValue()
```
BenWhitehead: can we add a stat to track success and failures in this `if-else`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
```diff
@@ -0,0 +1,77 @@
...
+
+class UniverseInstallQueueReader(zkClient: CuratorFramework) extends InstallQueueReader[PackageDefinition] {
+
+  val installQueue = InstallQueueHelpers.universeInstallQueue
+
+  override def getData(bytes: Array[Byte]): PackageDefinition = {
+    val str = new String(bytes, StandardCharsets.UTF_8)
+    decode[PackageDefinition](str) match {
+      case Xor.Right(pkgDef) => pkgDef
+      case Xor.Left(failure) => throw ZooKeeperStorageError("Couldn't parse data read from Zookeeper")
```
BenWhitehead: I think we need a more informative error here.

Also how do you imagine cosmos would recover from this error if this happened?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
```diff
@@ -0,0 +1,77 @@
...
+  val installQueue = InstallQueueHelpers.localInstallQueue
+
+  override def getData(bytes: Array[Byte]): Uri = {
+    new String(bytes, StandardCharsets.UTF_8).as[Uri]
+  }
+}
+
+trait InstallQueueReader[T] {
+
+  val installQueue: String
```
BenWhitehead: Can we rename `installQueue` to `zkPath` or something like that? I had no idea what `installQueue` meant until I looked at the values in the concrete classes.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
```diff
@@ -0,0 +1,77 @@
...
+import io.circe.jawn.decode
+
+import org.apache.curator.framework.CuratorFramework
+
+import java.nio.charset.StandardCharsets
+import java.util.Base64
+
+import scala.collection.JavaConversions._
+
+class UniverseInstallQueueReader(zkClient: CuratorFramework) extends InstallQueueReader[PackageDefinition] {
```
BenWhitehead: please make class `final` and `private`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
```diff
@@ -0,0 +1,77 @@
...
+    val str = new String(bytes, StandardCharsets.UTF_8)
+    decode[PackageDefinition](str) match {
+      case Xor.Right(pkgDef) => pkgDef
+      case Xor.Left(failure) => throw ZooKeeperStorageError("Couldn't parse data read from Zookeeper")
+    }
+  }
+
+}
+
+class LocalInstallQueueReader(zkClient: CuratorFramework) extends InstallQueueReader[Uri] {
```
BenWhitehead: please make class `final` and `private`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
```diff
@@ -0,0 +1,77 @@
...
+      }
+    }
+  }
+
+  def queue(zkClient: CuratorFramework): Future[List[PackageCoordinate]] = {
+    Future {
+      zkClient.getChildren().forPath(installQueue).toList.map(_.as[PackageCoordinate])
+    }
+  }
+}
```
BenWhitehead: please create a companion object to `InstallQueueReader` with factory methods for each of the concrete implementations.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
```diff
@@ -0,0 +1,77 @@
...
+class LocalInstallQueueReader(zkClient: CuratorFramework) extends InstallQueueReader[Uri] {
+
+  val installQueue = InstallQueueHelpers.localInstallQueue
+
+  override def getData(bytes: Array[Byte]): Uri = {
+    new String(bytes, StandardCharsets.UTF_8).as[Uri]
+  }
+}
+
+trait InstallQueueReader[T] {
```
BenWhitehead: please make this trait `sealed`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueWriter.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import io.circe.Encoder
+import io.circe.jawn.decode
+import io.circe.syntax._
+
+import org.apache.curator.framework.CuratorFramework
+
+import java.nio.charset.StandardCharsets
+
+class UniverseInstallQueueWriter(zkClient: CuratorFramework) extends InstallQueueWriter[PackageDefinition] {
```
BenWhitehead: please make this class `final` and `private`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueWriter.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  val installQueue = InstallQueueHelpers.universeInstallQueue
+
+  override def getBytes(pkgDef: PackageDefinition) =
+    pkgDef.asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
+}
+
+class LocalInstallQueueWriter(zkClient: CuratorFramework) extends InstallQueueWriter[Uri] {
```
BenWhitehead: please make this class `final` and `private`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueWriter.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  val installQueue = InstallQueueHelpers.localInstallQueue
+
+  override def getBytes(uri: Uri) = {
+    uri.toString.getBytes(StandardCharsets.UTF_8)
+  }
+}
+
+trait InstallQueueWriter[T] {
```
BenWhitehead: please make this trait `sealed`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueWriter.scala
```diff
@@ -0,0 +1,70 @@
...
+  ): Future[Unit] = {
+
+    val pkgPath = s"${installQueue}/${pkg.as[String]}"
+    Future {
+      if (zkClient.checkExists().forPath(pkgPath) != null) {
+        zkClient.delete().deletingChildrenIfNeeded().forPath(pkgPath)
+      }
+    }
+  }
+}
```
BenWhitehead: please create a companion object to `InstallQueueWriter` with factory methods for each of the concrete implementations.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/DistributedAtomicCounter.scala
```diff
@@ -0,0 +1,50 @@
...
+        }
+      }
+    }
+  }
+}
+
+private object DistributedAtomicCounter {
+
+  private val counterPath: String = "/package/counter"
+  private val retryPolicy = new ExponentialBackoffRetry(1000, 3)
```
BenWhitehead: Please remove this `retryPolicy` and instead have it passed in the constructor to the class. 

The fact this has to be passed is annoying since the curator client we've instantiated already has a backup configured for it.

https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala#L29

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
BenWhitehead: Should we change the naming scheme here? This reader isn't really for the whole Install Queue, it's for a specific entry in the queue. Perhaps something like `InstallQueueItemReader`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
tamarrow: If we can't read from zookeeper, we can either
1. try to fetch package manifest again using PackageCoordinate or
2. remove the zknode so the user can re-install

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
tamarrow: It reads from the queue though. In `readPackage` it only reads one item, but in `queue()` it reads all items.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
tamarrow: sure thing

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
BenWhitehead: If we can't read from zk at all we probably need to communicate that to the user somehow (maybe a 503?)

Regarding point 2, do we currently have a plan to allow a user to cancel an install?  This cancel code could be used to remove the unreadable node.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
BenWhitehead: If i'm reading it correctly `queue` just lists the nested znode names (which happen to be coordinates) whereas the actual items in the queue are the contents of those znodes.  Is this correct?

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Decoders.scala
```diff
@@ -13,4 +14,6 @@ object Decoders {
   implicit val decodeZooKeeperStorageEnvelope: Decoder[ZooKeeperStorageEnvelope] =
     deriveDecoder[ZooKeeperStorageEnvelope]
 
+  implicit val decodePackageCoordinate: Decoder[PackageCoordinate]=
```
cruhland: Nitpick: space before `=`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueAdder.scala
```diff
@@ -0,0 +1,57 @@
...
+
+private[cosmos] final class UniverseInstallQueueAdder(zkClient: CuratorFramework) extends InstallQueueAdder[PackageDefinition] {
+
+  val zkPath = InstallQueueHelpers.universeInstallQueue
+
+  override def getBytes(pkgDef: PackageDefinition) =
+    pkgDef.asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
+}
+
+private[cosmos] final class LocalInstallQueueAdder(zkClient: CuratorFramework) extends InstallQueueAdder[Uri] {
```
cruhland: Why do these subclasses have `zkClient` as constructor parameters when they aren't used?

cruhland: If you'd rather have the `zkClient` be an instance variable instead of a parameter of `InstallQueueAdder.addPackage`, you could make `InstallQueueAdder` an abstract class instead of a trait.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+}
+
+private[cosmos] final class LocalInstallQueueReader(zkClient: CuratorFramework) extends InstallQueueReader[Uri] {
+
+  val zkPath = InstallQueueHelpers.localInstallQueue
+
+  override def getData(bytes: Array[Byte]): Uri = {
+    new String(bytes, StandardCharsets.UTF_8).as[Uri]
```
BenWhitehead: you don't need the `.as[Uri]` here. The implicit conversion from string to Uri comes from `com.netaporter.uri.dsl._`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
```diff
@@ -0,0 +1,77 @@
...
+    val str = new String(bytes, StandardCharsets.UTF_8)
+    decode[PackageDefinition](str) match {
+      case Xor.Right(pkgDef) => pkgDef
+      case Xor.Left(failure) => throw ZooKeeperStorageError(s"Couldn't parse package definition from Zookeeper: $str")
+    }
+  }
+
+}
+
+private[cosmos] final class LocalInstallQueueReader(zkClient: CuratorFramework) extends InstallQueueReader[Uri] {
```
cruhland: Same here, the `zkClient` doesn't appear to be used. In fact, could the conversion between `T` and `Array[Byte]` be implemented as a `Bijection` or similar? Then you might be able to eliminate the subclasses altogether by having:

```scala
final class InstallQueueReader[T](zkClient: CuratorFramework, zkPath: String, dataToBytes: Bijection[T, Array[Byte]])
```

(maybe `Injection` or `Conversion` would be better, but it's the same idea)

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueAdder.scala
```diff
@@ -0,0 +1,57 @@
...
+    zkClient: CuratorFramework,
+    pkg: PackageCoordinate,
+    data: T
+  ): Future[Unit] = {
+
+    val pkgPath = s"${zkPath}/${pkg.as[String]}"
+    Future {
+      if (zkClient.checkExists().forPath(pkgPath) == null) {
+          zkClient.create.creatingParentsIfNeeded.forPath(pkgPath)
+          zkClient.setData().forPath(pkgPath, getBytes(data))
```
cruhland: Do we need to protect against races here? If two simultaneous installs are requested for a package with the same name and version, but different contents, only one will be added to the queue.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueAdder.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  def addPackage(
+    zkClient: CuratorFramework,
+    pkg: PackageCoordinate,
+    data: T
+  ): Future[Unit] = {
+
+    val pkgPath = s"${zkPath}/${pkg.as[String]}"
+    Future {
+      if (zkClient.checkExists().forPath(pkgPath) == null) {
```
cruhland: Potentially more idiomatic Scala (avoiding `null`):

```scala
Option(zkClient.checkExists().forPath(pkgPath)).getOrElse {
  // ...
}
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueReader.scala
```diff
@@ -0,0 +1,77 @@
...
+    pkg: PackageCoordinate
+  ): Future[T] = {
+
+    val pkgPath = s"${zkPath}/${pkg.as[String]}"
+    Future {
+      if (zkClient.checkExists().forPath(pkgPath) != null) {
+        getData(zkClient.getData().forPath(pkgPath))
+      } else {
+        throw PackageNotFound(pkg.name)
+      }
```
cruhland: Scalafied:

```scala
Option(zkClient.checkExists().forPath(pkgPath)) match {
  case Some(_) => getData(zkClient.getData().forPath(pkgPath))
  case _ => throw PackageNotFound(pkg.name)
}
```

Although, wouldn't it be more robust to omit the existence check? If the node gets deleted in between the `checkExists()` and the `getData()`, you'll still need to handle the error that results.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/InstallQueueUpdater.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  def deletePackage(
+    zkClient: CuratorFramework,
+    pkg: PackageCoordinate
+  ): Future[Unit] = {
+
+    val pkgPath = s"${zkPath}/${pkg.as[String]}"
+    Future {
+      if (zkClient.checkExists().forPath(pkgPath) != null) {
+        zkClient.delete().deletingChildrenIfNeeded().forPath(pkgPath)
```
cruhland: Similar concurrency issue here? What happens if the node is deleted in between the call to `checkExists()` and the `delete()`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Storage.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import java.nio.charset.StandardCharsets
+import java.util.Base64
+
+object Storage {
+
+  implicit val packageCoordinateToBase64String
+  : Bijection[PackageCoordinate, String] = {
+    def fwd(coord: PackageCoordinate): String = {
+      Base64.getEncoder.encodeToString(
```
jsancio: Please double check with what ZooKeeper allows for the znode name but you may want to make this URL safe base64 encoding.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueueAdder.scala
```diff
@@ -0,0 +1,88 @@
...
+
+sealed abstract class PackageQueueAdder[T](zkClient: CuratorFramework) {
+
+  val zkPath: String
+
+  def getBytes(a: T): Array[Byte]
+
+  def addPackage(
+    pkg: PackageCoordinate,
+    data: T
```
jsancio: If we make `T` a typeclass that implements the enveloper that @larioj created then we should be able to remove the `getBytes` methods above move them to the typeclass.

@tamarrow and @larioj we may need to make some changes to the class you created for enveloping to be able to use it here. The main difference with what we have now is that now we need to support decode variable types.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueueHelpers.scala
```diff
@@ -0,0 +1,17 @@
...
+  val nodeAlreadyExists = false
+
+  val packageQueueBase = "/package/state"
+  val localPackageQueue = s"$packageQueueBase/local"
+  val universePackageQueue = s"$packageQueueBase/universe"
+
+  sealed trait EnumVal
+  case object Install extends EnumVal
+  case object Uninstall extends EnumVal
+  val PackageState = Seq(Install, Uninstall)
```
jsancio: We never use these values.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueueHelpers.scala
```diff
@@ -0,0 +1,17 @@
+package com.mesosphere.cosmos.storage
+
+object PackageQueueHelpers {
+
+  val newNode = true
+  val nodeAlreadyExists = false
```
jsancio: We should use a similar pattern to the one for PackageState.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueueHelpers.scala
jsancio: `EnumVal` is not a great name. How about `PackageOperation` with `PackageOperations` being `Seq(Install, Uninstall)`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueueReader.scala
```diff
@@ -0,0 +1,136 @@
...
+  ) extends  BackgroundCallback {
+
+    private[this] val logger = org.slf4j.LoggerFactory.getLogger(getClass)
+
+    override def processResult(client: CuratorFramework, event: CuratorEvent): Unit = {
+      if (event.getType == CuratorEventType.GET_DATA) {
+	val code = KeeperException.Code.get(event.getResultCode)
+	code match {
+	  case KeeperException.Code.OK =>
+	    promise.setValue(Some(getData(event.getData)))
```
jsancio: How do we know which T to call?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueueReader.scala
```diff
@@ -0,0 +1,136 @@
...
+    zkClient.getData().inBackground(
+      new ReadPackage(promise)
+    ).forPath(
+      pkgPath
+    )
+
+    promise
+  }
+
+  private final class ReadPackage(
```
jsancio: Not a big fan of implicit `this` pointer that get pass around threads. I.e. the thread executing `readPackage` above is different from the thread executing `processResult`. You can get around this by making this class a top level class and not a nested class. When you do that you will lose access to the `getData` method above but we need to fix that to take into account that we are going to store both `PackageDefinition` and `Uri`. Let's you, @larioj and I sit down to figure out a solution.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueueReader.scala
```diff
@@ -0,0 +1,136 @@
...
+    zkClient.getChildren().inBackground(
+      new ReadQueue(promise)
+    ).forPath(
+      zkPath
+    )
+
+    promise
+  }
+
+  private final class ReadQueue(
```
jsancio: See my previous comment regarding implicit `this` and top level classes.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueueHelpers.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  val packageQueueBase = "/package/state"
+  val localPackageQueue = s"$packageQueueBase/local"
+  val universePackageQueue = s"$packageQueueBase/universe"
```
jsancio: Unless there is a specific reason for doing this I would like these two znodes to be the same. We should be able to do that if we use the same envelope we use for ZooKeeper storage class for repo list.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueueUpdater.scala
```diff
@@ -0,0 +1,81 @@
...
+    zkClient.delete.deletingChildrenIfNeeded.inBackground(
+      new DeletePackage(promise)
+    ).forPath(
+      pkgPath
+    )
+
+    promise
+  }
+
+  private final class DeletePackage(
```
jsancio: See my comment above regarding implicit `this` and top level classes.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueue.scala
```diff
@@ -0,0 +1,234 @@
...
+    pkg: PackageCoordinate,
+    error: Option[ErrorResponse]
+  ): Future[Unit] = {
+
+    val pkgPath = s"$packageQueueBase/${pkg.as[String]}"
+    val errPath = s"$errorQueueBase/${pkg.as[String]}"
+
+    val transaction = zkClient.inTransaction().delete().forPath(pkgPath).and()
+    error match {
+      case Some(e) => transaction.create().forPath(pkgPath, Envelope.encodeData(e)).and()
```
tamarrow: s/errPath/pkgPath

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueue.scala
```diff
@@ -0,0 +1,234 @@
...
+      case None => transaction.delete().forPath(errPath).and()
+    }
+
+    try {
+      // TODO: verify results from commit are what we expect
+      transaction.commit()
+      Future.Unit
+    } catch {
+      case e: Exception => throw ZooKeeperStorageError(e.getMessage)
```
tamarrow: Future.exception

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageQueue.scala
```diff
@@ -0,0 +1,234 @@
...
+
+    override def processResult(client: CuratorFramework, event: CuratorEvent): Unit = {
+      if (event.getType == CuratorEventType.GET_DATA) {
+	val code = KeeperException.Code.get(event.getResultCode)
+	code match {
+	  case KeeperException.Code.OK =>
+	    promise.setValue(Some(Envelope.decodeData[ErrorResponse](event.getData)))
+	  case KeeperException.Code.NONODE =>
+	    promise.setValue(None)
+          case _ =>
```
tamarrow: add logger

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/Envelope.scala
```diff
@@ -5,7 +5,7 @@ import java.nio.charset.StandardCharsets
 
 import com.mesosphere.cosmos.{CirceError, EnvelopeError}
 import com.mesosphere.cosmos.circe.{MediaTypedDecoder, MediaTypedEncoder}
-import com.mesosphere.cosmos.http.{MediaType, MediaTypeOps}
+import com.mesosphere.cosmos.http.{MediaType, MediaTypeOps, MediaTypes}
```
jsancio: Remove the import for `MediaTypes`. Probably not used in this file.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Storage.scala
```diff
@@ -0,0 +1,46 @@
...
+import io.circe.jawn.decode
+import io.circe.syntax._
+
+import java.nio.charset.StandardCharsets
+import java.util.Base64
+
+object Storage {
+
+  implicit val packageCoordinateToBase64String
+  : Bijection[PackageCoordinate, String] = {
```
jsancio: This should be an `Injection`. We can always go from a `PackageCoordinate` to a `String` but we can't always go from a `String` to a `PackageCoordinate`.

cruhland:
@larioj I deleted the `distributed-publish` branch because it's been merged. If you want to merge these changes, please open a new PR to `master`.

PR #513, opened 2016-09-29
Cleanup sbt config in preparation of artifact publishing

BenWhitehead:
* Flush out circe exclusion on finch for correct pom creation
* Add `publishTo` and set it to ossrh
* Add `pomExtra` with license, scm urls, and developer list (alphabetical order)

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -59,10 +59,13 @@ object CosmosBuild extends Build {
     ))
 
     val finch = Seq(
-      "com.github.finagle" %% "finch-core" % V.finch,
+      "com.github.finagle" %% "finch-core" % V.finch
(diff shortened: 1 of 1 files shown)


project/build.scala
```diff
@@ -199,7 +202,43 @@ object CosmosBuild extends Build {
...
+          <connection>scm:git:https://github.com/dcos/cosmos.git</connection>
+        </scm>
+        <developers>
+          <developer>
+            <name>Ben Whitehead</name>
+          </developer>
+          <developer>
+            <name>Charles Ruhland</name>
+          </developer>
+          <developer>
```
cruhland: Include Jesus?

PR #516, opened 2016-10-11
Create new modules for http, finch, bijection and render

BenWhitehead:
#### New Modules
* `cosmos-http`: Code related to parsing, representing and evaluating http headers
* `cosmos-finch`: Extensions to finch that facilitate creating EndpointHandlers and performing request content negotiation
* `cosmos-bijection`: Set of bijections for classes in cosmos-model
* `cosmos-jsonschema`: Code related to validation json objects against json schemas represented as cirice json object.
* `cosmos-render`: Code related to rendering the runtime manifest for packages

#### Modifications made while creating new modules
* Remove `com.mesosphere.cosmos.internal.PackageDefinition` and `com.mesosphere.cosmos.internal.CosmosInternalRepository` and update `PackageCollection` to use universe model directly
  * This modification was made due to the fact that the new render module operates on the universe model since that is the only model that is publicly exposed
* Move "Bundle" related objects into `com.mesosphere.cosmos.internal` since they are not part of the public API of cosmos or universe


cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/BijectionUtils.scala
```diff
@@ -0,0 +1,12 @@
+package com.mesosphere.universe.bijection
+
+import scala.util.{Failure, Success}
+
+object BijectionUtils {
+  def twitterTryToScalaTry[A, B](a: (A) => com.twitter.util.Try[B]): (A) => scala.util.Try[B] = {
```
cruhland: This signature is a bit odd, although I understand why it's written this way -- you're "decorating" a function to give it a different result type. Would it be clearer to have this operate on the `Try`s directly, and use `andThen` to compose it with conversion functions? For example:

```scala
def twitterTryToScalaTry[A](a: com.twitter.util.Try[A]): scala.util.Try[A] = { ... }

val rev = (universe.v3.model.PackagingVersion.apply _)
              .andThen(BijectionUtils.twitterTryToScalaTry)
```

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -0,0 +1,158 @@
...
+    pkgDef.marathon match {
+      case None => Xor.Left(MissingMarathonV2AppTemplate)
+      case Some(m) =>
+        val defaultOptions = pkgDef.config.map(JsonSchema.extractDefaultsFromSchema)
+        val defaultOptionsAndUserOptions = Seq(defaultOptions, options).flatten.foldLeft(JsonObject.empty)(merge)
+        validateOptionsAgainstSchema(pkgDef, defaultOptionsAndUserOptions).flatMap { _ =>
+          // now that we know the users options are valid for the schema, we build up a composite json object
+          // to send into the mustache context for rendering. The following seq prepares for the merge
+          // of all the options, documents at later indices have higher priority than lower index objects
+          // order here is important, DO NOT carelessly re-order.
```
cruhland: Are there tests that will fail if the order is changed?

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -0,0 +1,158 @@
...
+      jsonArray = _.map(jsonToJava).asJava,
+      jsonObject = _.toMap.mapValues(jsonToJava).asJava
+    )
+  }
+
+  private[this] def extractLabels(obj: Json): Xor[PackageDefinitionRenderError, JsonObject] = {
+    val hasLabels = obj.cursor.fieldSet.exists(_.contains("labels"))
+    if (hasLabels) {
+      // this is a bit of a sketchy check since marathon could change underneath us, but we want to try and
+      // surface this error the user as soon as possible. The check that is being performed here is to ensure
```
cruhland: "surface this error _**to**_ the user..."

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -0,0 +1,158 @@
...
+      // that `.labels` of `obj` is a Map[String, String]. Marathon enforces this and produces a very cryptic
+      // message if not adhered to, so we try and let the user know here were we can craft a more informational
```
cruhland: "...let the user know here _**where**_ we can..."

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -0,0 +1,158 @@
...
+      // error message.
+      // If marathon ever changes it's schema for labels then this code will most likely need a new version with
+      // this version left in tact for backward compatibility reasons.
```
cruhland: "in tact" -> intact

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -0,0 +1,158 @@
...
+      obj.cursor.get[Map[String, String]]("labels") match {
+        case Xor.Left(err) => Xor.Left(InvalidLabelSchema(err))
+        case Xor.Right(m) => Xor.Right(JsonObject.fromMap(m.mapValues(_.asJson)))
+      }
```
cruhland: Are there tests covering this check?

BenWhitehead: Yup in the corresponding spec `PackageDefinitionRendererSpec` under `if .labels from .marathon.v2AppMustacheTemplate`

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -0,0 +1,158 @@
...
+    pkgDef: PackageDefinition,
+    options: Option[JsonObject],
+    marathonAppId: Option[AppId]
+  ): Xor[PackageDefinitionRenderError, Json] = {
+    pkgDef.marathon match {
+      case None => Xor.Left(MissingMarathonV2AppTemplate)
+      case Some(m) =>
+        val defaultOptions = pkgDef.config.map(JsonSchema.extractDefaultsFromSchema)
+        val defaultOptionsAndUserOptions = Seq(defaultOptions, options).flatten.foldLeft(JsonObject.empty)(merge)
+        validateOptionsAgainstSchema(pkgDef, defaultOptionsAndUserOptions).flatMap { _ =>
```
BenWhitehead: I feel like the nesting here is justified and should be kept in place. This function is orchestrating all the other functions to result in a rendered template and I think it should all be kept together. I found it a bit confusing when extracting this code to have to cross the multiple methods to get the total result.

cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/BijectionUtils.scala
BenWhitehead: will update

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
BenWhitehead: will update

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  private[this] def extractLabels(obj: Json): Xor[PackageDefinitionRenderError, JsonObject] = {
+    val hasLabels = obj.cursor.fieldSet.exists(_.contains("labels"))
+    if (hasLabels) {
+      // this is a bit of a sketchy check since marathon could change underneath us, but we want to try and
+      // surface this error the user as soon as possible. The check that is being performed here is to ensure
+      // that `.labels` of `obj` is a Map[String, String]. Marathon enforces this and produces a very cryptic
+      // message if not adhered to, so we try and let the user know here were we can craft a more informational
+      // error message.
+      // If marathon ever changes it's schema for labels then this code will most likely need a new version with
```
BenWhitehead: will update

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
BenWhitehead: will update

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageSearchSpec.scala
```diff
@@ -158,8 +158,8 @@ private object PackageSearchSpec extends TableDrivenPropertyChecks {
       "database",
       "riak",
       "NoSql"
-    ).map(universe.v3.model.PackageDefinition.Tag(_)),
-    selected = Some(false),
+    ).map(universe.v3.model.PackageDefinition.Tag(_).get),
+    selected = None,
```
cruhland: Any significant reason this was changed? I know it doesn't change the meaning but I'm just curious.

BenWhitehead: All of the value objects were updated to have private constructors with the validation performed in the `apply` of the companion object and corresponding `Try` as the return type.

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
cruhland: I'd like to think that calls to well-named methods in the right places would be less confusing, since it would keep the code at a uniform level of abstraction. But I'll defer to your judgment since you just spent a few days untangling this code. :+1:

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageSearchSpec.scala
cruhland: Sorry, I should have been more specific. I was asking about why `selected = Some(false)` changed to `selected = None`.

BenWhitehead: ah, side effect of ripping out the internal PackageDefinition that has non-option `selected`

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
BenWhitehead: test added

PR #522, opened 2016-10-20
Finagle streaming requests POC test

cruhland:


cosmos-server/src/it/scala/com/mesosphere/cosmos/finagle/FinagleStreamingSpec.scala
```diff
@@ -0,0 +1,87 @@
...
+import org.scalatest.FreeSpec
+
+final class FinagleStreamingSpec extends FreeSpec {
+
+  import FinagleStreamingSpec._
+
+  "A Finagle service can read a streamed request body that cannot fit on the heap" in {
+    val host = "127.0.0.1"
+    val port = ":10101"
+    val packageSizeHeader = "Dcos-Package-Content-Length"
```
jsancio: I know this is just a test. Our code should probably use something like `X-DCOS-CONTENT-LENGTH`.

PR #523, opened 2016-10-21
Add Package Coordinate

larioj:


cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Storage.scala
```diff
@@ -0,0 +1,40 @@
...
+import io.circe.syntax._
+import java.nio.charset.StandardCharsets
+import java.util.Base64
+
+import com.mesosphere.cosmos.rpc.v1.model.PackageCoordinate
+
+object Storage {
+
+  implicit val packageCoordinateToBase64String
+  : Bijection[PackageCoordinate, String] = {
```
larioj: This should be an Injection. We can always go from a PackageCoordinate to a String but we can't always go from a String to a PackageCoordinate.

larioj:
I'd appreciate some feedback on the location of things.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Storage.scala
jsancio: You can put this `Injection` here https://github.com/dcos/cosmos/blob/local-package-collection/cosmos-bijection/src/main/scala/com/mesosphere/cosmos/bijection/CosmosConversions.scala

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Storage.scala
```diff
@@ -0,0 +1,40 @@
...
+    def fwd(coordinate: PackageCoordinate): String = {
+      Base64.getEncoder.encodeToString(
```
jsancio: https://github.com/dcos/cosmos/blob/local-package-collection/cosmos-bijection/src/main/scala/com/mesosphere/cosmos/bijection/CosmosConversions.scala

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PackageCoordinate.scala
```diff
@@ -0,0 +1,5 @@
+package com.mesosphere.cosmos.rpc.v1.model
+
+import com.mesosphere.universe
+
+final case class PackageCoordinate(name: String, version: universe.v3.model.PackageDefinition.Version)
```
cruhland: Do you have an example request or response object that this will appear in?

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Storage.scala
cruhland: Also, you might want to have a small unit test with a few examples of conversions in both directions.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PackageCoordinate.scala
jsancio: Eventually `/package/describe` will return this: https://github.com/dcos/cosmos/blob/local-package-collection/cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/LocalPackage.scala#L27-L30

cruhland:
@larioj looks good, just one comment about testing.

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Storage.scala
larioj: @jsancio @cruhland  Cosmos-Bijection does not depend on circe. So moving this there is difficult. Cosmos-Bijection also does not depend on Cosmos-Json so I can't use the coders defined there. What should I do?

jsancio: @larioj I suggest we implement this in `cosmos-server`: .../cosmos/converter/Common.scala

cruhland:
:ship: it

PR #524, opened 2016-10-21
Rename 'install' and 'uninstall' endpoints to 'run' and 'kill', respectively

cruhland:
- This is to "make room" for the new 'install' and 'uninstall' endpoints.
- All code, comments, and text related to the old endpoints has been updated.
- Caveat: kept 'preInstallNotes' and related fields from the Universe spec

Other things we might want to address:
- The `framework` parameter in `package-schema.json` is documented with the following description: `True if this package installs a new Mesos framework`. The wording could be updated in Universe.
- There's a `/package/publish` endpoint and related classes from @larioj's earlier work. Is there a reason to keep it?


jsancio:
@cruhland This looks good to me! :shipit: 

>The framework parameter in package-schema.json is documented with the following description: True if this package installs a new Mesos framework. The wording could be updated in Universe.

Let's file an issue under milestone 1. We need to change this in both Universe and Cosmos, no?

> There's a /package/publish endpoint and related classes from @larioj's earlier work. Is there a reason to keep it?

Let's review this code and decide if there is anything there that we should keep.

PR #526, opened 2016-10-24
Add Install Queue API

larioj:
The purpose of this PR is to discuss the Install Queue API and whether we want to change some names.

PR #530, opened 2016-10-28
Implement install queue

larioj:
This is a follow on to #526. This PR aims to implement the install queue. It satisfies the following interfaces: ProducerView, ProcessorView, and ReaderView. It has tests that depend on ZK, so they have all been ignored. They passed on my machine.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
@@ -8,6 +8,6 @@ object MediaTypes {
 
   // Storage Media Types
   val RepositoryList = MediaType.vndJson(List("dcos", "package"))("repository.repo-list", 1)
-  val OperationStatus = MediaType.vndJson(List("dcos", "package"))("package-queue.status", 1)
+  val OperationStatus = MediaType.vndJson(List("dcos", "package"))("operation.status", 1)
 
 }
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
@@ -8,5 +8,7 @@ object MediaTypes {
 
   // Storage Media Types
   val RepositoryList = MediaType.vndJson(List("dcos", "package"))("repository.repo-list", 1)
+  val Operation = MediaType.vndJson(List("dcos", "package"))("package-queue.contents", 1)
+  val OperationFailure = MediaType.vndJson(List("dcos", "package"))("package-queue.errors", 1)
 
 }
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
@@ -8,7 +8,6 @@ object MediaTypes {
 
   // Storage Media Types
   val RepositoryList = MediaType.vndJson(List("dcos", "package"))("repository.repo-list", 1)
-  val Operation = MediaType.vndJson(List("dcos", "package"))("package-queue.contents", 1)
-  val OperationFailure = MediaType.vndJson(List("dcos", "package"))("package-queue.errors", 1)
+  val OperationStatus = MediaType.vndJson(List("dcos", "package"))("package-queue.status", 1)
 
 }
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
@@ -7,6 +7,7 @@ import com.mesosphere.cosmos.rpc.v1.model.PackageCoordinate
 import com.mesosphere.cosmos.storage.Envelope
 import com.mesosphere.cosmos.storage.v1.circe.MediaTypedDecoders._
 import com.mesosphere.cosmos.storage.v1.circe.MediaTypedEncoders._
+import com.mesosphere.universe.bijection.BijectionUtils
 import com.twitter.bijection.Conversion.asMethod
 import com.twitter.util.Future
 import com.twitter.util.Promise
@@ -22,15 +23,22 @@ import org.apache.curator.framework.recipes.cache.TreeCache
 import org.apache.zookeeper.KeeperException
 import org.apache.zookeeper.data.Stat
 import scala.collection.JavaConversions._
-import scala.collection.mutable.ListBuffer
 
 final class InstallQueue private(client: CuratorFramework) extends
-  ProcessorView with ProducerView with ReaderView {
+  ProcessorView with ProducerView with ReaderView with AutoCloseable {
 
   import InstallQueue._
 
   private[this] val operationStatusCache = new TreeCache(client, installQueuePath)
-  operationStatusCache.start()
+
+  def start(): Unit = {
+    operationStatusCache.start()
+    ()
+  }
+
+  override def close(): Unit = {
+    operationStatusCache.close()
+  }
 
   override def next(): Future[Option[PendingOperation]] = {
     getAllCoordinates.flatMap { coordinates =>
@@ -39,7 +47,11 @@ final class InstallQueue private(client: CuratorFramework) extends
         if (pendingOperations.isEmpty) {
           None
         } else {
-          Some(pendingOperations.minBy(_.info.getCtime).value)
+          // We return the earliest modified pending operations as
+          // determined by the ZooKeeper mzxid. See
+          // https://zookeeper.apache.org/doc/r3.2.2/zookeeperProgrammers.html#sc_timeInZk
+          // for more information
+          Some(pendingOperations.minBy(_.stat.getMzxid).value)
         }
       }
     }
@@ -48,12 +60,12 @@ final class InstallQueue private(client: CuratorFramework) extends
   private[this] def getOperationIfPending
   (
     packageCoordinate: PackageCoordinate
-  ): Future[Option[Info[PendingOperation]]] = {
+  ): Future[Option[WithZkStat[PendingOperation]]] = {
     getOperationStatus(packageCoordinate).map { maybeOperationStatus =>
       maybeOperationStatus.flatMap {
-        case Info(info, Pending(operation, failure)) =>
-          Some(Info(
-            info,
+        case WithZkStat(stat, Pending(operation, failure)) =>
+          Some(WithZkStat(
+            stat,
             PendingOperation(packageCoordinate, operation, failure)
           ))
         case _ => None
@@ -64,16 +76,19 @@ final class InstallQueue private(client: CuratorFramework) extends
   private[this] def getAllCoordinates: Future[List[PackageCoordinate]] = {
     val promise = Promise[List[PackageCoordinate]]()
     client.getChildren.inBackground(
-      handler(promise) { case event if event.getType == CuratorEventType.CHILDREN =>
-        val code = KeeperException.Code.get(event.getResultCode)
-        code match {
-          case KeeperException.Code.OK =>
-            Try(event.getChildren.toList.map(_.as[scala.util.Try[PackageCoordinate]].get))
-          case KeeperException.Code.NONODE =>
-            Return(List())
-          case _ =>
-            Throw(KeeperException.create(code, event.getPath))
-        }
+      handler(promise, CuratorEventType.CHILDREN) {
+        case (KeeperException.Code.OK, event) =>
+          val coordinates =
+            event.getChildren.toList.map { encodedCoordinate =>
+              BijectionUtils.scalaTryToTwitterTry(
+                encodedCoordinate.as[scala.util.Try[PackageCoordinate]]
+              )
+            }
+          Try.collect(coordinates).map(_.toList)
+        case (KeeperException.Code.NONODE, _) =>
+          Return(List())
+        case (code, event) =>
+          Throw(KeeperException.create(code, event.getPath))
       }
     ).forPath(
       installQueuePath
@@ -92,16 +107,16 @@ final class InstallQueue private(client: CuratorFramework) extends
           "Attempted to signal failure on an " +
             s"operation not in the install queue: $packageCoordinate"
         Future.exception(InstallQueueError(message))
-      case Some(Info(_, Failed(_))) =>
+      case Some(WithZkStat(_, Failed(_))) =>
         val message =
           "Attempted to signal failure on an " +
             s"operation that has failed: $packageCoordinate"
         Future.exception(InstallQueueError(message))
-      case Some(Info(info, Pending(operation, _))) =>
+      case Some(WithZkStat(stat, Pending(operation, _))) =>
         setOperationStatus(
           packageCoordinate,
           Failed(OperationFailure(operation, error)),
-          info.getVersion
+          stat.getVersion
         )
     }
   }
@@ -109,22 +124,24 @@ final class InstallQueue private(client: CuratorFramework) extends
   private[this] def getOperationStatus
   (
     packageCoordinate: PackageCoordinate
-  ): Future[Option[Info[OperationStatus]]] = {
-    val promise = Promise[Option[Info[OperationStatus]]]()
+  ): Future[Option[WithZkStat[OperationStatus]]] = {
+    val promise = Promise[Option[WithZkStat[OperationStatus]]]()
     client.getData.inBackground(
-      handler(promise) { case event if event.getType == CuratorEventType.GET_DATA =>
-        val code = KeeperException.Code.get(event.getResultCode)
-        code match {
-          case KeeperException.Code.OK =>
-            Return(Some(Info(
-              event.getStat,
-              Envelope.decodeData[OperationStatus](event.getData)
-            )))
-          case KeeperException.Code.NONODE =>
-            Return(None)
-          case _ =>
-            Throw(KeeperException.create(code, event.getPath))
-        }
+      handler(promise, CuratorEventType.GET_DATA) {
+        case (KeeperException.Code.OK, event) =>
+          Try(Envelope.decodeData[OperationStatus](event.getData))
+            .map { operationStatus =>
+              Some(
+                WithZkStat(
+                  event.getStat,
+                  operationStatus
+                )
+              )
+            }
+        case (KeeperException.Code.NONODE, _) =>
+          Return(None)
+        case (code, event) =>
+          Throw(KeeperException.create(code, event.getPath))
       }
     ).forPath(
       statusPath(packageCoordinate)
@@ -140,14 +157,11 @@ final class InstallQueue private(client: CuratorFramework) extends
   ): Future[Unit] = {
     val promise = Promise[Unit]()
     client.setData().withVersion(version).inBackground(
-      handler(promise) { case event if event.getType == CuratorEventType.SET_DATA =>
-        val code = KeeperException.Code.get(event.getResultCode)
-        code match {
-          case KeeperException.Code.OK =>
-            Return(())
-          case _ =>
-            Throw(KeeperException.create(code, event.getPath))
-        }
+      handler(promise, CuratorEventType.SET_DATA) {
+        case (KeeperException.Code.OK, _) =>
+          Return(())
+        case (code, event) =>
+          Throw(KeeperException.create(code, event.getPath))
       }
     ).forPath(
       statusPath(packageCoordinate),
@@ -166,13 +180,13 @@ final class InstallQueue private(client: CuratorFramework) extends
           "Attempted to signal success on an " +
             s"operation not in the install queue: $packageCoordinate"
         Future.exception(InstallQueueError(message))
-      case Some(Info(_, Failed(_))) =>
+      case Some(WithZkStat(_, Failed(_))) =>
         val message =
           "Attempted to signal success on an " +
             s"operation that has failed: $packageCoordinate"
         Future.exception(InstallQueueError(message))
-      case Some(Info(info, Pending(operation, _))) =>
-        deleteOperationStatus(packageCoordinate, info.getVersion)
+      case Some(WithZkStat(stat, Pending(operation, _))) =>
+        deleteOperationStatus(packageCoordinate, stat.getVersion)
     }
   }
 
@@ -183,14 +197,11 @@ final class InstallQueue private(client: CuratorFramework) extends
   ): Future[Unit] = {
     val promise = Promise[Unit]()
     client.delete().withVersion(version).inBackground(
-      handler(promise) { case event if event.getType == CuratorEventType.DELETE =>
-        val code = KeeperException.Code.get(event.getResultCode)
-        code match {
-          case KeeperException.Code.OK =>
-            Return(())
-          case _ =>
-            Throw(KeeperException.create(code, event.getPath))
-        }
+      handler(promise, CuratorEventType.DELETE) {
+        case (KeeperException.Code.OK, _) =>
+          Return(())
+        case (code, event) =>
+          Throw(KeeperException.create(code, event.getPath))
       }
     ).forPath(
       statusPath(packageCoordinate)
@@ -219,14 +230,11 @@ final class InstallQueue private(client: CuratorFramework) extends
   ): Future[AddResult] = {
     val promise = Promise[AddResult]()
     client.create().creatingParentsIfNeeded().inBackground(
-      handler(promise) { case event if event.getType == CuratorEventType.CREATE =>
-        val code = KeeperException.Code.get(event.getResultCode)
-        code match {
-          case KeeperException.Code.OK =>
-            Return(Created)
-          case _ =>
-            Throw(KeeperException.create(code, event.getPath))
-        }
+      handler(promise, CuratorEventType.CREATE) {
+        case (KeeperException.Code.OK, _) =>
+          Return(Created)
+        case (code, event) =>
+          Throw(KeeperException.create(code, event.getPath))
       }
     ).forPath(
       statusPath(packageCoordinate),
@@ -234,7 +242,7 @@ final class InstallQueue private(client: CuratorFramework) extends
     )
     promise
   }
-
+  
   private[this] def setOperationInOperationStatus
   (
     packageCoordinate: PackageCoordinate,
@@ -242,50 +250,62 @@ final class InstallQueue private(client: CuratorFramework) extends
   ): Future[AddResult] = {
     getOperationStatus(packageCoordinate).flatMap {
       case None =>
-        // if we reach this statement, that means there was
-        // an operation pending when the user made the request. The operation has
-        // since completed, but it is not incorrect to return AlreadyExists, because
-        // the request overlaps the existence of a pending operation.
+        /* if we reach this statement, that means there was
+         * an operation when the user made the request. The operation has
+         * since completed which means that it was pending.
+         * It is correct to return AlreadyExists, because
+         * the request to add overlaps the existence of a pending operation.
+         *
+         * |------------|----------------------------|
+         * |add starts  |                            |
+         * |------------|----------------------------|
+         * |Create      |  Node Exists               |
+         * |            |  Contents = ?              |
+         * |            |.............................
+         * |            |  Node Exists               |
+         * |            |  Contents = Pending        |
+         * |            |                            |
+         * |            |  Note:                     |
+         * |            |  This must be Pending      |
+         * |            |  otherwise node would not  |
+         * |            |  have been deleted         |
+         * |            |                            |
+         * |------------|----------------------------|
+         * |Set         |  Node does not Exists      |
+         * |            |                            |
+         * |            |                            |
+         * |------------|----------------------------|
+         * |add ends    |                            |
+         * |------------|----------------------------|
+         */
         Future.value(AlreadyExists)
-      case Some(Info(_, Pending(_, _))) =>
+      case Some(WithZkStat(_, Pending(_, _))) =>
         Future.value(AlreadyExists)
-      case Some(Info(info, Failed(failure))) =>
+      case Some(WithZkStat(stat, Failed(failure))) =>
         setOperationStatus(
           packageCoordinate,
           Pending(operation, Some(failure)),
-          info.getVersion
+          stat.getVersion
         ).before(Future.value(Created))
     }
   }
 
   override def viewStatus(): Future[Map[PackageCoordinate, OperationStatus]] = {
-    def listTryToTryList[A](listTry: List[Try[A]]): Try[List[A]] = {
-      listTry
-        .foldLeft(
-          Try(ListBuffer[A]())
-        ) { (tryOfList, oneTry) =>
-          tryOfList.flatMap { list =>
-            oneTry.map { element =>
-              list :+ element
-            }
-          }
-        }.map(_.toList)
-    }
-
     val operationStatus =
       Option(operationStatusCache
         .getCurrentChildren(installQueuePath))
         .getOrElse(new java.util.HashMap[String, ChildData]())
         .toMap
         .map { case (encodedPackageCoordinate, childData) =>
-          Try(encodedPackageCoordinate.as[scala.util.Try[PackageCoordinate]].get).flatMap {
-            packageCoordinate =>
-              Try(packageCoordinate
-                -> Envelope.decodeData[OperationStatus](childData.getData))
+          BijectionUtils.scalaTryToTwitterTry(
+            encodedPackageCoordinate.as[scala.util.Try[PackageCoordinate]]).flatMap { coordinate =>
+            Try(Envelope.decodeData[OperationStatus](childData.getData)).map { operationStatus =>
+              coordinate -> operationStatus
+            }
           }
         }
-        .toList
-    Future.const(listTryToTryList(operationStatus).map(_.toMap))
+        .toSeq
+    Future.const(Try.collect(operationStatus).map(_.toMap))
   }
 
 }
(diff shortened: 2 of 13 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  /** Signals to the install queue that the operation on the package has
+    * been successful. This will delete the node from the install queue.
+    *
+    * @param packageCoordinate The package coordinate of the package whose
+    *                          operation has succeeded.
+    * @return a future that will complete when state changes have been completed
+    */
+  override def success(packageCoordinate: PackageCoordinate): Future[Unit] = {
```
larioj: This does not delete the associated OperationFailure.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,280 @@
...
+
+    val transaction =
+      client.inTransaction()
+        .delete().forPath(operationPath)
+      .and()
+        .create().forPath(failurePath, Envelope.encodeData(operationFailure))
+      .and()
+
+    try {
+      transaction.commit()
```
jsancio: Let's do this in a FuturePool. We do this for `S3ObjectStorage` and `LocalObjectStorage`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,280 @@
...
+import org.apache.curator.framework.api.BackgroundCallback
+import org.apache.curator.framework.api.CuratorEvent
+import org.apache.curator.framework.api.CuratorEventType
+import org.apache.curator.framework.recipes.cache.ChildData
+import org.apache.curator.framework.recipes.cache.TreeCache
+import org.apache.zookeeper.KeeperException
+import scala.collection.JavaConversions._
+import scala.util.Try
+
+final class InstallQueue(client: CuratorFramework) extends
```
jsancio: Since you have an `apply` method in the companion object we can make this constructor private: `... class InstallQueue private (client: ...`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,280 @@
...
+  ): Future[Unit] = {
+    val operationFailure = OperationFailure(operation, failure)
+    val operationPath = s"$pendingOperationsPath/${packageCoordinate.as[String]}"
+    val failurePath = s"$failedOperationsPath/${packageCoordinate.as[String]}"
+
+    val transaction =
+      client.inTransaction()
+        .delete().forPath(operationPath)
+      .and()
+        .create().forPath(failurePath, Envelope.encodeData(operationFailure))
```
larioj: There may already be an operation failure associated with this package coordinate

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/PendingOperation.scala
```diff
@@ -5,5 +5,6 @@ import com.mesosphere.cosmos.rpc
 case class PendingOperation(
   packageCoordinate: rpc.v1.model.PackageCoordinate,
   operation: Operation,
-  failure: Option[OperationFailure]
+  failure: Option[OperationFailure],
+  creationTime: Long
```
jsancio: Let's document this argument or give it a better type. I.e. what is the user suppose to do with this Long? Is it just for sorting? If it just for sorting then a different argument name is better: `creationOrder`. If the user can use this establish a wall clock create time then a different type is better: `java.time.Instant`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
cruhland: Does this mean that if a package coordinate has a failure associated with it, and then another operation on it fails, that second failure is lost?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,280 @@
...
+      .and()
+
+    try {
+      transaction.commit()
+      Future.Unit
+    } catch {
+      case e: KeeperException.NoNodeException =>
+        createFailedOperationsPath().flatMap(_ => this.failure(packageCoordinate, operation, failure))
```
cruhland: We can't do this during the transaction?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
cruhland: You're going to add that in this PR?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,280 @@
...
+    promise
+  }
+
+  /** Signals to the install queue that the operation on the package has
+    * been successful. This will delete the node from the install queue.
+    *
+    * @param packageCoordinate The package coordinate of the package whose
+    *                          operation has succeeded.
+    * @return a future that will complete when state changes have been completed
+    */
```
cruhland: Remove comment

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,280 @@
...
+  /** Adds an operation on a package to the install queue. It will return
+    * an object signaling failure when there is already an operation outstanding
+    * on the package coordinate.
+    *
+    * @param packageCoordinate the package coordinate on which the operation
+    *                          should be performed.
+    * @param operation         The operation to be performed
+    * @return Created if the operation was added to the queue, else AlreadyExists
+    *         if the operation was not added to the queue.
+    */
```
cruhland: Delete comment.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,280 @@
...
+      Envelope.encodeData(operation)
+    )
+    promise
+  }
+
+  /** Shows the status of every package in the install queue.
+    *
+    * @return A map from package coordinate to the state of any pending or
+    *         failed operations associated with that package coordinate
+    */
```
cruhland: Remove comment

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,280 @@
...
+        .toMap
+        .map { case (encodedPc, data) =>
+          val pc = encodedPc.as[Try[PackageCoordinate]].toOption.get //TODO
+          logger.debug("encoded package coordinate: {}", encodedPc)
+          logger.debug("decoded package coordinate: {}", pc)
+          pc -> Some(Envelope.decodeData[OperationFailure](data.getData))}
+        .withDefault(_ => None)
+
+    Future.value((failed.keySet ++ pending.keySet)
+      .map(pc => pc -> OperationStatus(pending(pc), failed(pc))).toMap)
```
cruhland: How coherent should the pending and failed data be for a given coordinate? Using the above approach, it's possible to see a package in a misleading state, because e.g. we could have the following operations:

1. (this thread) read pending operation for coordinate
1. (another thread) update pending and failed operations for coordinate
1. (this thread) read failed operation for coordinate

larioj:
@jsancio @cruhland One item that has not been resolved is Charles' comment:

"""
It would be nice if we could actually run these tests on TeamCity. But since they need a ZK server, they're not really unit tests. If we make this an integration test, I can see two options:

Use the ZK server created by the existing integration test harness. We would have to make sure any znodes used by this test did not interfere with the znodes used by other tests.
Start a ZK server just for this test, but on a port that doesn't conflict with the ZK for the other integration tests.
@jsancio @larioj what do you think?
"""

I'm okay with either option.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
```diff
@@ -0,0 +1,472 @@
...
+  "Reader view" - {
+    "ViewStatus should " +
+      "return all pending and failed operations in the queue" ignore { f =>
+      insertPackageStatusIntoQueue(f, coordinate1, statusWithOnlyOperation1)
+      insertPackageStatusIntoQueue(f, coordinate2, statusWithOnlyOperation1)
+      insertPackageStatusIntoQueue(f, coordinate3, statusWithOnlyFailure1)
+      insertPackageStatusIntoQueue(f, coordinate4, statusWithOnlyFailure1)
+      insertPackageStatusIntoQueue(f, coordinate5, statusWithOperation1AndFailure1)
+
+      val pollCount = 10
```
larioj: Each poll waits for a second, so 10 seconds should be enough for the cache to catch up.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+
+  def statusPath(packageCoordinate: PackageCoordinate): String = {
+    s"$installQueuePath/${packageCoordinate.as[String]}"
+  }
+
+  def apply(client: CuratorFramework): InstallQueue = new InstallQueue(client)
+
+  private case class Info[A](info: Stat, value: A)
+
+  private def handler[A](promise: Promise[A])(handle: PartialFunction[CuratorEvent, Try[A]]) = {
```
larioj: Quoting @cruhland """How do we want to deal with events that aren't in the domain of the partial function?"""

I could check whether the partial function is defined for event, and fail the future if it is not. But the function should be defined for event. Not being defined for event is a bad bug that should be catched easily with tests, and should be fixed.

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
```diff
@@ -8,5 +8,6 @@ object MediaTypes {
 
   // Storage Media Types
   val RepositoryList = MediaType.vndJson(List("dcos", "package"))("repository.repo-list", 1)
+  val OperationStatus = MediaType.vndJson(List("dcos", "package"))("package-queue.status", 1)
```
cruhland: How about `operation.status` instead? The `queue` part seems too implementation-specific.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+object InstallQueue {
+  val installQueuePath = "/packages"
+
+  def statusPath(packageCoordinate: PackageCoordinate): String = {
+    s"$installQueuePath/${packageCoordinate.as[String]}"
+  }
+
+  def apply(client: CuratorFramework): InstallQueue = new InstallQueue(client)
+
+  private case class Info[A](info: Stat, value: A)
```
cruhland: Change some names to make this more clear?

```scala
private case class WithZkStat[A](stat: Stat, value: A)
```

cruhland:
@larioj for my unresolved comment, I'm leaning towards using the same ZK as the other tests. We could just make sure to configure the ZK client with a different path prefix than Cosmos and/or other tests, and delete that subtree when the test completes.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
```diff
@@ -0,0 +1,472 @@
...
+    Stream.tabulate(attempts) { _ =>
+      val oneSecond = 1000L
+      Thread.sleep(oneSecond)
+      Await.result(f.installQueue.viewStatus())
+    }.dropWhile(_.size < size).headOption
+  }
+
+  "Producer View" - {
+    "Add an operation " - {
+      "when no parent path exists" ignore { f =>
```
cruhland: I think it would be easier to understand these fixture arguments `f` if they were more descriptive. Also, it looks like a lot of your helper functions take in the `FixtureParam` but only use one of the two fields. Why not pattern match here and then pass along only the fields that are needed to the helper functions?

```scala
"when no parent path exists" ignore { case FixtureParam(client, installQueue) =>
```

You could even use `_` when a field isn't needed in a test.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
```diff
@@ -0,0 +1,472 @@
...
+    size: Int
+  ): Option[Map[PackageCoordinate, OperationStatus]] = {
+    Stream.tabulate(attempts) { _ =>
+      val oneSecond = 1000L
+      Thread.sleep(oneSecond)
+      Await.result(f.installQueue.viewStatus())
+    }.dropWhile(_.size < size).headOption
+  }
+
+  "Producer View" - {
```
cruhland: Could you move the actual test cases to the top of the class, and the put everything else below them? We want to be able to immediately look at the test when we open a file.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+        setOperationStatus(
+          packageCoordinate,
+          Pending(operation, Some(failure)),
+          info.getVersion
+        ).before(Future.value(Created))
+    }
+  }
+
+  override def viewStatus(): Future[Map[PackageCoordinate, OperationStatus]] = {
+    def listTryToTryList[A](listTry: List[Try[A]]): Try[List[A]] = {
```
cruhland: You'll want to use `Try.collect` (companion object method) instead

cruhland: And in general the operation `F[G[A]] => G[F[A]]` is frequently useful; the Cats library implements it for many common data structures as [`sequence`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Traverse.scala#L83).

cosmos-server/src/main/scala/com/mesosphere/cosmos/http/MediaTypes.scala
larioj: Okay

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
larioj: This is a good idea.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
```diff
@@ -0,0 +1,472 @@
...
+    s"Attempted to signal success on an " +
+      s"operation that has failed: $coordinate1"
+  private val pendingOperation1Coordinate1NoFailure =
+    PendingOperation(coordinate1, operation1, None)
+  private val pendingOperation1Coordinate2NoFailure =
+    PendingOperation(coordinate2, operation1, None)
+  private val pendingOperation1Coordinate1Failure1 =
+    PendingOperation(coordinate1, operation1, Some(failure1))
+  private val pendingOperation1Coordinate5Failure1 =
+    PendingOperation(coordinate5, operation1, Some(failure1))
```
cruhland: Of the above fields, the `coordinateN`s, `errorResponse1`, and the error messages are useful. I think the rest would be more clear if they were just written inline, since there's a cost to remembering (or looking up) what a variable name corresponds to.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
```diff
@@ -0,0 +1,472 @@
...
+              f.installQueue.failure(coordinate1, errorResponse1)
+            )
+          )
+          assertResult(alreadyFailedFailureMessageCoordinate1)(error.msg)
+        }
+
+        "on a coordinate that has an operation and a failure" ignore { f =>
+          insertPackageStatusIntoQueue(f, coordinate1, statusWithOperation1AndFailure1)
+          Await.result(
+            f.installQueue.failure(coordinate1, errorResponse1)
```
cruhland: Since it's updating the failure here, we should probably use a different `ErrorResponse` value so we can tell it was changed.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
```diff
@@ -0,0 +1,472 @@
...
+  "Processor view" - {
+    "failure" - {
+      "Fail an operation " - {
+        "when no parent path exists" ignore { f =>
+          val error = intercept[InstallQueueError](
+            Await.result(
+              f.installQueue.failure(coordinate1, errorResponse1)
+            )
+          )
+          assertResult(notInQueueFailureMessageCoordinate1)(error.msg)
```
cruhland: You can also "catch" errors by using `Future.liftToTry`:

```scala
val Throw(error) = Await.result(
  f.installQueue.failure(coordinate1, errorResponse1).liftToTry
)
assertResult(notInQueueFailureMessageCoordinate1)(error.msg)
```

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
larioj: Okay

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: Okay

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
```diff
@@ -0,0 +1,472 @@
...
+      val expectedState =
+        Some(Map(
+          coordinate1 -> statusWithOnlyOperation1,
+          coordinate2 -> statusWithOnlyOperation1,
+          coordinate3 -> statusWithOnlyFailure1,
+          coordinate4 -> statusWithOnlyFailure1,
+          coordinate5 -> statusWithOperation1AndFailure1
+        ))
+      assertResult(expectedState)(actualState)
+    }
```
cruhland: Let's use the expected data to generate the initial data, to ensure they stay in sync:

```scala
"return all pending and failed operations in the queue" ignore { f =>
  val expectedState = Map(
    coordinate1 -> statusWithOnlyOperation1,
    // etc...
  )
  expectedState.foreach { case (coordinate, status) =>
    insertPackageStatusIntoQueue(f, coordinate, status)
  }

  val pollCount = 10
  // etc...

  assertResult(Some(expectedState))(actualState)
}
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  private[this] def getAllCoordinates: Future[List[PackageCoordinate]] = {
+    val promise = Promise[List[PackageCoordinate]]()
+    client.getChildren.inBackground(
+      handler(promise) { case event if event.getType == CuratorEventType.CHILDREN =>
+        val code = KeeperException.Code.get(event.getResultCode)
+        code match {
+          case KeeperException.Code.OK =>
+            Try(event.getChildren.toList.map(_.as[scala.util.Try[PackageCoordinate]].get))
```
cruhland: I think `Try.collect` would work here also? Then you don't need the `.get`

larioj: I agree. I'd still need the get though since I'm using it to convert a scala Try to a twitter Try. I looked around a lot, and I did not find a way to convert between the two directly. Do you know of any?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+        if (pendingOperations.isEmpty) {
+          None
+        } else {
+          Some(pendingOperations.minBy(_.info.getCtime).value)
+        }
+      }
+    }
+  }
+
+  private[this] def getOperationIfPending
```
cruhland: I think you can get rid of this method: in `next()` just use `getOperationStatus` directly, and then perform a `filter` after calling `flatten`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+    packageCoordinate: PackageCoordinate,
+    operation: Operation
+  ): Future[AddResult] = {
+    getOperationStatus(packageCoordinate).flatMap {
+      case None =>
+        // if we reach this statement, that means there was
+        // an operation pending when the user made the request. The operation has
+        // since completed, but it is not incorrect to return AlreadyExists, because
+        // the request overlaps the existence of a pending operation.
+        Future.value(AlreadyExists)
```
cruhland: What if we reversed the order of operations for `add`? That is, we first try to set the operation in an existing node; if it doesn't exist, we reach this `case None` here, and then call `createOperationStatus`, which can be very simple -- it just needs a `case NODEEXISTS` where we would return `AlreadyExists`.

We wouldn't need this comment if we did it that way, what do you think?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+  (
+    packageCoordinate: PackageCoordinate,
+    operation: Operation
+  ): Future[AddResult] = {
+    createOperationStatus(
+      packageCoordinate,
+      Pending(operation, None)
+    ).rescue {
+      case e: KeeperException.NodeExistsException =>
+        setOperationInOperationStatus(packageCoordinate, operation)
```
larioj: In my opinion it's more readable to catch it here, rather than nesting a call to a future inside the background handler. Also the structure of the method conveys that it's attempting to create, and then if the node exists, it is attempting to set. If it were nested you'd have to follow that logic through the background handler, which has a lot of other noise. This is a nicer separation of concerns, with the added benefit that it makes createOperationStatus reusable by other methods, even though we don't use it anywhere else.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+      Option(operationStatusCache
+        .getCurrentChildren(installQueuePath))
+        .getOrElse(new java.util.HashMap[String, ChildData]())
+        .toMap
+        .map { case (encodedPackageCoordinate, childData) =>
+          Try(encodedPackageCoordinate.as[scala.util.Try[PackageCoordinate]].get).flatMap {
+            packageCoordinate =>
+              Try(packageCoordinate
+                -> Envelope.decodeData[OperationStatus](childData.getData))
+          }
```
cruhland: Looks like you might have run into problems with Twitter's `Try` vs. Scala's `Try`. In general I think we want to be using Twitter's `Try` as often as possible; it's only for the Bijection stuff that we have to use Scala's `Try`. Ben wrote a function in `BijectionUtils` to do the conversion one way; you might need to write the reverse direction?

Assuming that conversion is written, then this block could be:

```scala
scalaTryToTwitterTry(encodedPackageCoordinate.as[scala.util.Try[PackageCoordinate]]).map {
  packageCoordinate =>
    packageCoordinate -> Envelope.decodeData[OperationStatus](childData.getData))
}
```

cruhland: (the `Try` will catch any exceptions that are thrown from the function passed to `map`)

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
cruhland: Ah yes, I just made a comment about this. See `BijectionUtils`, but I think you'll have to write the function you need.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
cruhland: Good point about not making calls inside `createOperationStatus`. But we do have a way to indicate to `add` that the node exists, without catching an exception: what if `createOperationStatus` returns `AlreadyExists` when it sees a `NODEEXISTS` code?

Then out here, we could just pattern match on the return value and call `setOperationInOperationStatus` if we get an `AlreadyExists`. Mainly I'm trying to use our own constructs as much as possible.

cruhland: Of course, there's also my other comment about switching the order of operations, but that's mostly orthogonal to this one.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
larioj: Thanks for the tip. I'll stick with the intercept since it's already written, and it gives an appropriate error when it fails to throw, and it's part of the testing library.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
```diff
@@ -0,0 +1,472 @@
...
+
+        "there are multiple pending operations some of which have failed" ignore { f =>
+          insertPackageStatusIntoQueue(f, coordinate1, statusWithOnlyOperation1)
+          insertPackageStatusIntoQueue(f, coordinate2, statusWithOnlyOperation1)
+          insertPackageStatusIntoQueue(f, coordinate3, statusWithOnlyFailure1)
+          insertPackageStatusIntoQueue(f, coordinate4, statusWithOnlyFailure1)
+          insertPackageStatusIntoQueue(f, coordinate5, statusWithOperation1AndFailure1)
+
+          val n1 = Await.result(f.installQueue.next())
+          removePackageStatusFromQueue(f, coordinate1)
```
larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+  operationStatusCache.start()
+
+  override def next(): Future[Option[PendingOperation]] = {
+    getAllCoordinates.flatMap { coordinates =>
+      Future.collect(coordinates.map(getOperationIfPending)).map { operationsMaybePending =>
+        val pendingOperations = operationsMaybePending.flatten
+        if (pendingOperations.isEmpty) {
+          None
+        } else {
+          Some(pendingOperations.minBy(_.info.getCtime).value)
```
jsancio: I think we should use `czxid` and we should write a comment noting what it meas with a link to https://zookeeper.apache.org/doc/r3.2.2/zookeeperProgrammers.html#sc_timeInZk

E.g. "The zxid of the change that caused this znode to be created."

What do you think?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+  ): Future[Option[Info[OperationStatus]]] = {
+    val promise = Promise[Option[Info[OperationStatus]]]()
+    client.getData.inBackground(
+      handler(promise) { case event if event.getType == CuratorEventType.GET_DATA =>
+        val code = KeeperException.Code.get(event.getResultCode)
+        code match {
+          case KeeperException.Code.OK =>
+            Return(Some(Info(
+              event.getStat,
+              Envelope.decodeData[OperationStatus](event.getData)
```
larioj: Yes

larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: I'd like to keep the try of the decoding explicit.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: The problem is that we need to propagate in the package coordinate. Which means that if we use getOperationStatus we need to map that to an optional pair of PackageCoordinate and WithZkStat[OperationStatus]. It's doable, but not very readable.

```scala
Future.collect(
  coordinates.map(coordinate => getOperationStatus(coordinate).map(_.map(coordinate -> _)))
)
```

The purpose of the getOperationIfPending method is to add readability, since we have so much nesting.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: I agree. It's easier to understand in this order.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: Yeah, this is  waaayy better.

larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -0,0 +1,314 @@
...
+import scala.collection.JavaConversions._
+import scala.collection.mutable.ListBuffer
+
+final class InstallQueue private(client: CuratorFramework) extends
+  ProcessorView with ProducerView with ReaderView {
+
+  import InstallQueue._
+
+  private[this] val operationStatusCache = new TreeCache(client, installQueuePath)
+  operationStatusCache.start()
```
larioj: Okay

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: Let's talk about this during standup.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
cruhland: Ah, got it. Could you add a comment to `getOperationIfPending` so that we know why it's useful to have?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
cruhland: That makes sense, but to really make it explicit let's wrap the `Try` *only* around the call to `decodeData`.

It would be even better if `decodeData` already returned a `Try`, but that's a change for another time...

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: Actually using created time is an error. We need to use modified time, since we want to sort based upon the time that an operationStatus became pending. In the original two node implementation this was exactly at created time, but in this implementation it is not.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: I was thinking the same thing :)

larioj:
@cruhland Using the same ZK would be easiest. Let's do it in another PR though.

larioj:
In hindsight, I should have made a commit per change. Sorry about that.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
larioj: Sure

cruhland:
Awesome work @larioj! :ship: it!

PR #531, opened 2016-10-30
Implementation for processing leader

jsancio:


cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/package.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  type Installer = (Uri, universe.v3.model.PackageDefinition) => Future[Unit]
+
+  type UniverseInstaller = universe.v3.model.PackageDefinition => Future[Unit]
+
+  type Uninstaller =
+    Either[
+      rpc.v1.model.PackageCoordinate,
+      universe.v3.model.PackageDefinition
+    ] => Future[Unit]
```
cruhland: Is there a meaningful distinction between the two cases of this `Either`? Can't we obtain a `PackageCoordinate` from a `PackageDefinition`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/OperationProcessorLeader.scala
```diff
@@ -0,0 +1,234 @@
...
+                pending.operation,
+                exceptionErrorResponse(err)
+              )
+          }
+        }
+      }
+    } catch {
+      case cancel: InterruptedException =>
+        // Just log that we lost leadership
+        logger.warn("Lost leadership while processing operation...", cancel)
```
cruhland: Is there a reason why we do this for `InterruptedException` only?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/OperationProcessorLeader.scala
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  private[this] def nextOperation(): Future[PendingOperation] = {
+    processorView.next().flatMap {
+      case Some(pending) => Future.value(pending)
+      case None =>
+        val duration = Duration.fromSeconds(10)
+        // TODO: Log at info
+        logger.warn(s"Nothing to do in the processor waiting for $duration")
+        Future.sleep(duration).before(nextOperation())
```
cruhland: Hmm, maybe instead of waiting for a fixed time period, the `Future` returned by `ProcessorView.next()` should not produce a result until another operation is available on the queue. That would make this code simpler to implement and get rid of the arbitrary polling period.

cruhland: So basically, change `ProcessorView.next()` to return `Future[PendingOperation]`, and implement it so that `Await.result(processorView.next())` would block until an operation was added to the queue.

cosmos-server/src/main/scala/com/mesosphere/cosmos/package.scala
```diff
@@ -0,0 +1,8 @@
+package com.mesosphere
+
+import com.twitter.util.ScheduledThreadPoolTimer
+import com.twitter.util.Timer
+
+package object cosmos {
+  implicit val globalTimer: Timer = new ScheduledThreadPoolTimer()
```
cruhland: Is this only here for the tests?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/OperationProcessorLeader.scala
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    def success(p: PackageCoordinate): Future[Unit] = {
+      Future(println("recording a success..."))
+    }
+  }
+
+  object TestInstaller extends Installer {
+    def apply(uri: Uri, pkg: PackageDefinition): Future[Unit] = {
+      val duration = Duration.fromSeconds(30)
```
cruhland: I know this is just a mock, but in case you still need mocks in the final test implementation, you might want to use a `Timer` that can be controlled by the test. Then you can simulate waiting for 30 seconds without actually using that amount of "wall clock" time.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/OperationProcessorLeader.scala
jsancio: Yes. This thread should not throw any exception except for that one. We should be handling all error conditions in the `flatMap` above. The only one that we don't handle is `InterruptedException` which can be thrown by the `Await.result` method because `stateChanged` sometime throws `CancelLeadershipException`. I schedule a quick meeting to talk about this in person. It is a very important point.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/OperationProcessorLeader.scala
jsancio: Yes. We can do that. @larioj what do you think?

jsancio: Actually, I thought about this some more. We should keep the `ProcessorView`'s `next` as is. We can talk about it in our meeting later today.

cosmos-server/src/main/scala/com/mesosphere/cosmos/package.scala
cruhland: :+1:

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/OperationProcessorLeader.scala
cruhland: Ah, yes, the `Future` takes care of the other exceptions. Maybe add a comment that the `InterruptedException` is expected because it can be triggered by the `CancelLeadershipException`? The connection isn't obvious unless you look at the docs for `CancelLeadershipException`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/package.scala
cruhland: Yeah I like that a little better actually. More clearly shows that we'll always know the package being uninstalled, but we won't always have the full definition.

jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/OperationProcessorLeader.scala
jsancio: We discussed this offline. Added some comments.

jsancio:
@cruhland @larioj PR is ready for review.

jsancio:
FYI, I rebased and squashed on top of the latest master.

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
```diff
@@ -0,0 +1,339 @@
...
+            Install(
+              Uri.parse("http://localhost/some.dcos"),
+              universe.v3.model.V3Package(
+                universe.v3.model.V3PackagingVersion,
+                "test",
+                universe.v3.model.PackageDefinition.Version("1.2.3"),
+                universe.v3.model.PackageDefinition.ReleaseVersion(0).get,
+                "does@not.matter",
+                "doesn't matter"
+              )
```
cruhland: Having the whole package definition inline is somewhat distracting. What if we factored it out into a method that has parameters for the fields that are referenced by the tests, such as `name` and `version`? Then the test code would still show the field values that mattered, while hiding the boilerplate.

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
```diff
@@ -0,0 +1,339 @@
...
+
+  val uninstallerError = new IllegalArgumentException("Uninstall failed")
+  object FailingUninstaller extends Uninstaller {
+    def apply(
+      pc: rpc.v1.model.PackageCoordinate,
+      pkg: Option[universe.v3.model.PackageDefinition]
+    ): Future[Unit] = Future.exception(
+      uninstallerError
+    )
+  }
```
cruhland: Why not make these mocks simple `val`s that store lambda functions? There'd be less boilerplate to wade through:

```scala
val uninstallerError = new IllegalArgumentException("Uninstall failed")
val FailingUninstaller: Uninstaller = _ => Future.exception(uninstallerError)
```

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
```diff
@@ -0,0 +1,339 @@
...
+    when(processorViewMock.success(packageCoordinate)).thenReturn(Future.Done)
+
+    val processor = OperationProcessor(
+      processorViewMock,
+      FailingInstaller,
+      SucceedingUniverseInstaller,
+      FailingUninstaller
+    )
+
+    Await.result(processor())
```
cruhland: Maybe add a comment saying that no assertions are needed because we just want no exceptions to be thrown?

cruhland: Ah, we should also assert that the mock methods were actually called.

cruhland:
@jsancio back to you! Had some comments about the tests...they should be extrapolated to all test cases/fixtures.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
```diff
@@ -0,0 +1,70 @@
...
+  processor: () => Future[Unit]
+) extends LeaderSelectorListener with AutoCloseable {
+  private[this] val logger = org.slf4j.LoggerFactory.getLogger(getClass)
+  private[this] val leaderSelector = new LeaderSelector(
+    curatorClient,
+    "/processor/leader",
+    this
+  )
+
+  override def stateChanged(
```
larioj: Why not extend LeaderSelectorListenerAdapter? Doesn't that do essentially this?

larioj: You wouldn't need to override this.

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
```diff
@@ -0,0 +1,339 @@
+package com.mesosphere.cosmos.repository
+
+import com.mesosphere.cosmos.circe.Encoders.exceptionErrorResponse
+import com.mesosphere.cosmos.model.ZooKeeperUri
```
larioj: This is not used

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
```diff
@@ -0,0 +1,339 @@
...
+import com.mesosphere.cosmos.storage.installqueue.PendingOperation
+import com.mesosphere.cosmos.storage.installqueue.ProcessorView
+import com.mesosphere.cosmos.storage.installqueue.Uninstall
+import com.mesosphere.cosmos.storage.installqueue.UniverseInstall
+import com.mesosphere.cosmos.zookeeper.Clients
+import com.mesosphere.universe
+import com.netaporter.uri.Uri
+import com.twitter.util.Await
+import com.twitter.util.Future
+import org.apache.curator.framework.CuratorFramework
```
larioj: Not used

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
```diff
@@ -0,0 +1,339 @@
...
+import org.apache.curator.test.TestingServer
```
larioj: Not used

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
```diff
@@ -0,0 +1,339 @@
...
+import org.mockito.Mockito.mock
+import org.mockito.Mockito.timeout
+import org.mockito.Mockito.verify
```
larioj: Not used

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
```diff
@@ -0,0 +1,339 @@
...
+
+import com.mesosphere.cosmos.circe.Encoders.exceptionErrorResponse
+import com.mesosphere.cosmos.model.ZooKeeperUri
+import com.mesosphere.cosmos.rpc
+import com.mesosphere.cosmos.storage.installqueue.Install
+import com.mesosphere.cosmos.storage.installqueue.PendingOperation
+import com.mesosphere.cosmos.storage.installqueue.ProcessorView
+import com.mesosphere.cosmos.storage.installqueue.Uninstall
+import com.mesosphere.cosmos.storage.installqueue.UniverseInstall
+import com.mesosphere.cosmos.zookeeper.Clients
```
larioj: not used

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
```diff
@@ -0,0 +1,339 @@
...
+
+  "Test success is called on install" in {
+    val packageCoordinate = rpc.v1.model.PackageCoordinate(
+      "test",
+      universe.v3.model.PackageDefinition.Version("1.2.3")
+    )
+    val processorViewMock = mock(classOf[ProcessorView])
+    when(processorViewMock.next()).thenReturn(
+      Future.value(
+        Some(
```
larioj: I wonder in general how careful we should be about keeping the package coordinate we pass to methods in synch with the package coordinate of the package definition in the operation. Install Queue does not check for this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  override def takeLeadership(client: CuratorFramework): Unit = {
+    try {
+      Await.result {
+        /* NOTE: To reduce the probability of two different Cosmos performing
+         * the same operation, it is very important that this function doesn't
+         * block forever. The return Future should try to terminate as soon as
+         * possible.
```
jsancio: @cruhland We can join the two `Future`s. `Future.join(Future.sleep(...), processor())` won't return until they both finish guaranteeing that we wait at least the sleep time. What do you think?

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
jsancio: Great idea! Will do.

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
jsancio: We discussed this offline. The tests fail if we call methods that were not sub out. Leaving as is.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
jsancio: Trying to avoid inheritance. I prefer composition over inheritance.

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
jsancio: That is true. At a minimum I would add this logic to the `/package/add` and/or `Installer`. They need to verify that their inputs are valid. cc @cruhland.

We may be able to remove data duplication in some places but not all. For example, we won't be able to remove duplication in the Install operation be cause the mapping is `PackageCoordinate => Uri => Zip(PackageDefinition, Resources)`. There is  duplication between `PackageCoordinate` and `PackageDefinition` but the issue is that one part of the system `InstallOps` cares about `PackageCoordinate => Uri` while another part of the system cares about `Uri => Zip(PackageDefinition, ...)`.

jsancio: @larioj I agree with you that we should have a method in `PackageDefintionOps` for `PackageDefinition => PackageCoordinate` do you want to add it? The only issue may be that they are in independent modules.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
jsancio: Done.

cosmos-server/src/test/scala/com/mesosphere/cosmos/repository/OperationProcessorSpec.scala
jsancio: Done. Good catch.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
larioj: Yes, but this already extends LeaderSelectorListener. You would instead extend LeaderSelectorListenerAdapter.

``` scala
final class SyncFutureLeader private (
  curatorClient: CuratorFramework,
  processor: () => Future[Unit]
) extends LeaderSelectorListenerAdapter with AutoCloseable
```

jsancio: We talked offline. Even though both solution use the same keyword hey have very different implementations. Decided to keep it as is.

jsancio:
@cruhland @larioj Ready for review. Thanks!

jsancio:
Thanks for the review @cruhland and @larioj

PR #533, opened 2016-11-04
Implement conversion from Reader to InputStream

jsancio:


cruhland:
@jsancio when looking at the docs for `InputStream`, I discovered [`PipedInputStream`](https://docs.oracle.com/javase/8/docs/api/java/io/PipedInputStream.html) and [`PipedOutputStream`](https://docs.oracle.com/javase/8/docs/api/java/io/PipedOutputStream.html). They seem to fit our use case pretty well...maybe we should just use those?

jsancio:
@cruhland and I talked about `PipedInputStream` offline and decided against it. Instead we are going to add bulk operation(s) for doing read.

jsancio:
All yours @cruhland

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorage.scala
```diff
@@ -85,4 +89,98 @@ object ObjectStorage {
     val directories: List[String]
     val listToken: Option[ListToken]
   }
+
+  final class ReaderInputStream private (reader: Reader) extends InputStream {
```
cruhland: Let's put this in its own file, since the implementation is not tied to `ObjectStorage`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorage.scala
```diff
@@ -85,4 +89,98 @@ object ObjectStorage {
...
+    import ReaderInputStream._
+
+    private[this] var maybeBuffer = Option.empty[Try[Option[Array[Byte]]]]
```
cruhland: The nested data structures of this value make some of the code harder to understand. I'm wondering why we need the `Try` and the outer `Option`?
- For the `Try`, can't we just let exceptions propagate when they are thrown? (all of the methods are invoked by the thread doing the reading, so it will be able to catch the exceptions as you wanted).
- For the outer `Option`, it looks like the only time when it's `None` is before any bytes have been read. Why not have the initial value be an empty array instead? That should have the same effect.

Summarizing:
```scala
private[this] var maybeBuffer: Option[Array[Byte]] = Some(Array.empty)
```

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/ObjectStorageSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+import com.twitter.concurrent.AsyncStream
+import com.twitter.io.Buf
+import com.twitter.io.Reader
+import com.twitter.io.StreamIO
+import com.twitter.util.Future
+import org.scalatest.FreeSpec
+import org.scalatest.Matchers
+import org.scalatest.prop.PropertyChecks
+
+final class ObjectStorageSpec extends FreeSpec with Matchers with PropertyChecks {
```
cruhland: Rename to `ReaderInputStreamSpec`?

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/ObjectStorageSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+    "for many buffer reader" in {
+      val expected = "hello world".getBytes()
+      val buffers = List(
+        "he".getBytes(),
+        "ll".getBytes(),
+        "o ".getBytes(),
+        "wo".getBytes(),
+        "rl".getBytes(),
+        "d".getBytes()
+      )
```
cruhland: `val buffers = "hello world".grouped(2).map(_.getBytes()).toList` ?

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/ObjectStorageSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+      StreamIO.buffer(
+        ObjectStorage.ReaderInputStream(
+          reader(buffers.map(array => new Buf.ByteArray(array, 0, array.length)))
+        )
+      ).toByteArray() shouldBe expected
+    }
+
+    "for failed reader" in {
```
cruhland: How about another one of these, but with a list of buffers, where some middle buffer in the list throws an `IOException` when `read()` is called on it? Then we could test that we received all the bytes up to the exception.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorage.scala
jsancio: Well, the `InputStream` can be in four states:

1. Haven't performed a read on the reader. This is represented by `maybeBuffer == None`
1. The last read on the reader returned a failure. This is represented by `maybeBuffer == Some(Throw(...))`
1. The last read on the reader returned a buffer. This is represented by `maybeBuffer == Some(Return(Some(...)))`
1. The last read on the reader was an EOF. This is represented by `maybeBuffer == Some(Return(None))`

We cannot simplify the state or else we information loss. We have two options. We can write a comment that explains the states above. Or we can create our own ADT.

I added some comments let me know if that works for you.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/ObjectStorageSpec.scala
jsancio: Done.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/ObjectStorageSpec.scala
jsancio: Nice. Done.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/ObjectStorageSpec.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/io/ReaderInputStream.scala
```diff
@@ -0,0 +1,128 @@
+package com.mesosphere.cosmos.io
```
cruhland: I like having an `io` package, but I don't like all the diff noise caused by it. Could we give it a different name to prevent all the name clashes? How about `ioutil` or `util.io`?

cruhland:
@jsancio a few more comments. Let me know if you want to discuss offline.

cosmos-server/src/main/scala/com/mesosphere/cosmos/io/ReaderInputStream.scala
jsancio: We talked offline. We are going to keep the `com.mesosphere.cosmos.io` package and try to keep the import rewrite as a separate commit.

cosmos-server/src/main/scala/com/mesosphere/cosmos/io/ReaderInputStream.scala
```diff
@@ -0,0 +1,128 @@
...
+import com.twitter.util.Try
+import java.io.InputStream
+import scala.math.min
+
+final class ReaderInputStream private (reader: Reader) extends InputStream {
+  import ReaderInputStream._
+
+  /* The InputStream can be in one of 5 general states:
+   * 1. Never read an input from the Reader: state == None
+   * 2. Last read from the Reader resulted in a failure: state == Some(Throw(_))
```
jsancio: Talked offline. I am going to change the stored state to `Option[(Array[Byte], Int)]`.

cosmos-server/src/test/scala/com/mesosphere/cosmos/io/ReaderInputStreamSpec.scala
```diff
@@ -0,0 +1,106 @@
...
+              reader(List(new Buf.ByteArray(expected, 0, expected.length))),
+              FailedReader
+            )
+          )
+        )
+      ).read(actual, 0, actual.length)
+
+      actual shouldBe expected
+      intercept[IllegalArgumentException] {
+        ReaderInputStream(FailedReader).read()
```
jsancio: Great catch. I'll fix it.

jsancio:
Check it out @cruhland. The new commit is [cbdf778](https://github.com/dcos/cosmos/pull/533/commits/cbdf778235babac2001efaaa10210dc450faf9e5).

cruhland:
Looks great @jsancio ! Ready to merge.

PR #534, opened 2016-11-04
Add stats to install queue

larioj:


cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -50,6 +50,7 @@ final class InstallQueue private(
   override def next(): Future[Option[PendingOperation]] = {
     timeFuture(stats.stat("next")) {
       getAllCoordinates.flatMap { coordinates =>
+        stats.stat("Number of Nodes").add(coordinates.length)
```
jsancio: Let's call this `numberOfNodesOnNext`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -173,6 +174,10 @@ final class InstallQueue private(
     version: Int
   ): Future[Unit] = {
     val promise = Promise[Unit]()
+
+    val data = Envelope.encodeData(operationStatus)
+    stats.stat("Set Node Size").add(data.length)
```
jsancio: Let's call this `nodeSize`. I think it is better to merge both the set and create histograms.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -250,6 +255,10 @@ final class InstallQueue private(
     operationStatus: OperationStatus
   ): Future[AddResult] = {
     val promise = Promise[AddResult]()
+
+    val data = Envelope.encodeData(operationStatus)
+    stats.stat("Create Node Size").add(data.length)
```
jsancio: Let's call this `nodeSize`. I think it is better to merge both the set and create histograms.

PR #537, opened 2016-11-08
Add service run handler

larioj:
This should do...

What do you guys think? @jsancio @cruhland

### DIFF
--- a/cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
+++ b/cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
@@ -6,6 +6,7 @@ object MediaTypes {
 
   private[this] val dcos = MediaType.vndJson(List("dcos")) _
   private[this] val dcosPackage = MediaType.vndJson(List("dcos", "package")) _
+  private[this] val dcosService = MediaType.vndJson(List("dcos", "service")) _
   private[this] def vnd(kind: String, version: Int = 1) = dcosPackage(kind, version)
 
   val PublishRequest = vnd("publish-request")
@@ -42,4 +43,6 @@ object MediaTypes {
   val V1ListResponse = vnd("list-response", 1)
   val V2ListResponse = vnd("list-response", 2)
 
+  val ServiceRunRequest = dcosService("run-request", 1)
+  val ServiceRunResponse = dcosService("run-response", 2)
 }
--- a/cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
+++ b/cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
@@ -43,6 +43,6 @@ object MediaTypes {
   val V1ListResponse = vnd("list-response", 1)
   val V2ListResponse = vnd("list-response", 2)
 
-  val ServiceRunRequest = dcosService("run-request", 1)
-  val ServiceRunResponse = dcosService("run-response", 2)
+  val ServiceStartRequest = dcosService("start-request", 1)
+  val ServiceStartResponse = dcosService("start-response", 2)
 }
--- a/cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
+++ b/cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
@@ -44,5 +44,5 @@ object MediaTypes {
   val V2ListResponse = vnd("list-response", 2)
 
   val ServiceStartRequest = dcosService("start-request", 1)
-  val ServiceStartResponse = dcosService("start-response", 2)
+  val ServiceStartResponse = dcosService("start-response", 1)
 }
--- a/cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v2/model/ServiceStartResponse.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v2/model/ServiceStartResponse.scala
@@ -3,7 +3,7 @@ package com.mesosphere.cosmos.rpc.v2.model
 import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
 import com.mesosphere.universe
 
-case class ServiceRunResponse(
+case class ServiceStartResponse(
   packageName: String,
   packageVersion: universe.v3.model.PackageDefinition.Version,
   appId: Option[AppId] = None
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
@@ -1,0 +1,74 @@
+package com.mesosphere.cosmos.handler
+
+import cats.data.Xor
+import com.mesosphere.cosmos._
+import com.mesosphere.cosmos.finch.EndpointHandler
+import com.mesosphere.cosmos.http.RequestSession
+import com.mesosphere.cosmos.render._
+import com.mesosphere.cosmos.repository.LocalPackageCollection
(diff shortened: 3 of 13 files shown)


larioj:
@jsancio We still need to figure out the URI for the local package collection.

jsancio:
Looking.

cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
```diff
@@ -42,4 +43,6 @@ object MediaTypes {
   val V1ListResponse = vnd("list-response", 1)
   val V2ListResponse = vnd("list-response", 2)
 
+  val ServiceRunRequest = dcosService("run-request", 1)
+  val ServiceRunResponse = dcosService("run-response", 2)
```
jsancio: I thought about this. Let's make this `start`.

jsancio: We may need to search and replace in a few places.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
```diff
@@ -0,0 +1,74 @@
...
+  override def apply(
+    request: rpc.v1.model.ServiceRunRequest
+  )(
+    implicit session: RequestSession
+  ): Future[rpc.v2.model.ServiceRunResponse] = {
+    localPackageCollection
+      .getInstalledPackage(
+        request.packageName,
+        request.packageVersion.as[Option[universe.v3.model.PackageDefinition.Version]])
+      .map(_.asInstanceOf[Installed].metadata -> Uri.parse("localPackageCollection"))
```
jsancio: The cast is not ideal. I think I kept the return type `LocalPackage` because we may want to return `Invalid` later on. We should either handle that here or change the return type for `getInstalledPackage` to `Installed` and handle it in that method.

jsancio: Regarding the "local" URI. We can make it `urn://cosmos.mesos/repo`. cc @cruhland

cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
cruhland: Why is `run-response` version 2 here? Shouldn't they both be version 1?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
cruhland: @jsancio I'm ok with the special "local" URI value for now, but longer-term would it be better represented with something like the following?

```scala
sealed trait PackageOrigin
case object Local extends PackageOrigin
case class ExternalRepo(uri: Uri) extends PackageOrigin
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
```diff
@@ -0,0 +1,74 @@
...
+
+        packageConfig match {
+          case Xor.Right(renderedMarathonJson) =>
+            packageRunner.launch(renderedMarathonJson)
+              .map { runnerResponse =>
+                rpc.v2.model.ServiceRunResponse(
+                  packageName = pkg.name,
+                  packageVersion = pkg.version,
+                  appId = Some(runnerResponse.id)
+                )
```
cruhland: Shouldn't we respond with the same amount of information as `/package/install`? Specifically, we should include the `postInstallNotes` and the `cli` fields here.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
```diff
@@ -0,0 +1,74 @@
...
+    request: rpc.v1.model.ServiceRunRequest
+  )(
+    implicit session: RequestSession
+  ): Future[rpc.v2.model.ServiceRunResponse] = {
+    localPackageCollection
+      .getInstalledPackage(
+        request.packageName,
+        request.packageVersion.as[Option[universe.v3.model.PackageDefinition.Version]])
+      .map(_.asInstanceOf[Installed].metadata -> Uri.parse("localPackageCollection"))
+      .flatMap { case (pkg, sourceUri) =>
```
cruhland: From here down the code is the same as for `/package/install`. Factor it out so we only have a single copy?

cruhland:
@larioj added a few comments.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
jsancio: @cruhland This is great. The only requirement is that it needs to be encodable and decodable to and from a JSON string because we persistent as a label in Marathon. Please work with @larioj on a solution.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
larioj: @jsancio had a reason for not wanting the cli field. And the postInstall notes should be shown post install. Although, I could add a postStartNotes field since install and start are quite conflated.

jsancio: The current "dcos package install" tries to give the user the illusion of being both "dcos package add" and "dcos service start". I argue that in the long term we want to make the command available as part of "dcos package add" not as part of "dcos service start". This opens the possibility of really cool flows. We can talk offline.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
larioj: Except that they are not, because they return different types. We could factor out the error conditions, but that would inhibit the check that we have covered all cases, and the code savings are not much. @jsancio and I had this conversation, as I was originally wrapping LocalPackageCollection, so that It would have a PackageCollection interface, and I could use the old installHandler unchanged. But the reasoning for doing it this way was to add new request/response types, and since we are eventually getting rid of the other handler, we could just delete it.

cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
larioj: Yes

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
cruhland: Ok, sounds like this was intentional :+1:

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceRunHandler.scala
cruhland: :+1:

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v2/model/ServiceStartResponse.scala
```diff
@@ -0,0 +1,10 @@
+package com.mesosphere.cosmos.rpc.v2.model
```
cruhland: This should be in the `v1` package.

cruhland:
@larioj one more change, and then LGTM. I'll approve in advance but please make the change before merging.

larioj:
@cruhland Thanks

PR #540, opened 2016-11-10
Flags for staged package object storage

larioj:


jsancio:
@larioj I'll take a look tomorrow morning.

larioj:
@jsancio and I had a conversation about the flags and we decided that the flags should be an S3 uri or a file URI. For example the packageStagingUri should be either s3://bucket/path or file://path. Similarly for the final storage location

project/CosmosIntegrationTestServer.scala
```diff
@@ -49,10 +51,13 @@ class CosmosIntegrationTestServer(javaHome: Option[String], itResourceDirs: Seq[
       .getOrElse("java")
 
     val dcosUri = systemProperty("com.mesosphere.cosmos.dcosUri").get
-    System.setProperty(
-      "com.mesosphere.cosmos.test.CosmosIntegrationTestClient.CosmosClient.uri",
-      "http://localhost:7070"
-    )
+    val stagedPackageBucket = systemProperty("com.mesosphere.cosmos.stagedPackageBucket").get
+    val stagedPackagePath = systemProperty("com.mesosphere.cosmos.stagedPackagePath").get
```
cruhland: These lines are causing the integration tests to fail. Since I have a test that uses them in my PR, I can update the build steps in TeamCity to make sure these properties get set.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceStartNotConfiguredHandler.scala
```diff
@@ -0,0 +1,16 @@
...
+import com.mesosphere.cosmos.http.RequestSession
+import com.mesosphere.cosmos.rpc.v1.model.{ServiceStartRequest, ServiceStartResponse}
+import com.twitter.util.Future
+
+private[cosmos] final class ServiceStartNotConfiguredHandler
+  extends EndpointHandler[rpc.v1.model.ServiceStartRequest, rpc.v1.model.ServiceStartResponse] {
+  override def apply(
+    request: ServiceStartRequest
+  )(implicit session: RequestSession): Future[ServiceStartResponse] = {
+    Future.exception(NotImplemented("Service Start has not been configured correctly"))
```
cruhland: Since there will be several endpoints that use this, let's make it generic? It should be possible to have `final class NotConfiguredHandler[Req, Res] extends EndpointHandler[Req, Res]`.

project/CosmosIntegrationTestServer.scala
```diff
@@ -95,6 +102,7 @@ class CosmosIntegrationTestServer(javaHome: Option[String], itResourceDirs: Seq[
         throw t
     }
   }
+  //scalastyle:on method.length
```
cruhland: Any way to pull out some code into helper methods? We shouldn't turn off style checks unless all other options are worse.

larioj: I'll give it a shot.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -4,24 +4,32 @@ import _root_.io.circe.Json
...
 import com.mesosphere.cosmos.repository.UniverseClient
 import com.mesosphere.cosmos.repository.ZkRepositoryList
 import com.mesosphere.cosmos.rpc.MediaTypes
 import com.mesosphere.cosmos.rpc.v1.circe.MediaTypedEncoders._
 import com.mesosphere.cosmos.rpc.v1.circe.MediaTypedRequestDecoders._
+import com.mesosphere.cosmos.rpc.v1.model.ServiceStartRequest
+import com.mesosphere.cosmos.rpc.v1.model.ServiceStartResponse
 import com.mesosphere.cosmos.rpc.v2.circe.MediaTypedEncoders._
-import com.mesosphere.cosmos.storage.{InMemoryPackageStorage, PackageStorage}
+import com.mesosphere.cosmos.storage._
```
jsancio: Let's avoid general wildcard imports. We are allowing for implicit but we should limit their use outside of implicit.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -154,6 +163,13 @@ object Cosmos extends FinchServer {
       val zkUri = zookeeperUri()
       logger.info("Using {} for the ZooKeeper connection", zkUri)
 
+      logger.info("Using {} for the package storage uri", packageStorageUri())
+      val objectStorage = packageStorageUri() match {
+        case Some(S3Uri(uri)) => Some(S3ObjectStorage(new AmazonS3Client(), uri))
+        case Some(FileUri(path)) => Some(LocalObjectStorage(path))
+        case None => None
+      }
```
jsancio: :+1:

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -167,7 +183,15 @@ object Cosmos extends FinchServer {
       val sourcesStorage = new ZkRepositoryList(zkClient)()
       val packageStorage = new InMemoryPackageStorage()
 
-      val cosmos = Cosmos(adminRouter, marathonPackageRunner, sourcesStorage, packageStorage, UniverseClient(adminRouter))
+      val cosmos =
+        Cosmos(
+          adminRouter,
+          marathonPackageRunner,
+          sourcesStorage,
+          packageStorage,
```
jsancio: Looks like we need to get rid of this `InMemoryPackageStorage`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -216,14 +240,22 @@ object Cosmos extends FinchServer {
...
       sourcesStorage,
       universeClient
     )
 
+    val serviceStartHandler = objectStorage match {
+      case Some(objStore) =>
+        new ServiceStartHandler(LocalPackageCollection(PackageObjectStorage(objStore)), packageRunner)
+      case None =>
+        new NotConfiguredHandler[ServiceStartRequest, ServiceStartResponse]
+    }
```
jsancio: Okay for now. We are going to be doing this for a lot of handlers in the coming days. We should figure out a better way of managing this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/ObjectStorageUri.scala
```diff
@@ -0,0 +1,27 @@
...
+import com.twitter.util.Try
+import com.twitter.util.Throw
+import java.nio.file.{Path, Paths}
+
+sealed trait ObjectStorageUri
+
+object ObjectStorageUri {
+  def parse(s: String): Try[ObjectStorageUri] = {
+    val s3UriStart = "s3://"
+    val fileUriStart = "file://"
```
jsancio: We can be even more permissive and make the strings: `s3:` and `file:` respectively and call the variables something like `s3UriScheme`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/ObjectStorageUri.scala
```diff
@@ -0,0 +1,27 @@
...
+    s match {
+      case uri if uri.toLowerCase.startsWith(s3UriStart) =>
+        Try(S3Uri(new AmazonS3URI(uri)))
+      case uri if uri.toLowerCase.startsWith(fileUriStart) =>
+        Try(FileUri(Paths.get(uri.drop(fileUriStart.length))))
+      case _ =>
+        Throw(NotValidObjectStorageUri())
```
jsancio: We use `CosmosError`s for things that we would like to serialize to the user using JSON. In this case we want Cosmos to fail to start. We can just throw an `IllegalArgumentException` and include a message with the argument. In this case `s`.

Also, let's user better variable names. We should avoid single letter variables at all costs. They are nice now but they are not nice a month later when you are trying to read this code.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -350,6 +350,8 @@ object Encoders extends LowPriorityImplicits {
         s"Package: [$name] version: [$version] does not have a Marathon template defined and can not be rendered"
       case EnvelopeError(msg) => msg
       case InstallQueueError(msg) => msg
+      case NotImplemented(msg) => msg
+      case NotValidObjectStorageUri() => "The given Uri was not a valid storage Uri"
```
jsancio: I left you a comment in another place but we don't need this type. This is a misconfiguration of Cosmos. We don't have serialize misconfiguration.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/NotConfiguredHandler.scala
```diff
@@ -0,0 +1,15 @@
...
+import com.mesosphere.cosmos.finch.EndpointHandler
+import com.mesosphere.cosmos.http.RequestSession
+import com.twitter.util.Future
+
+private[cosmos] final class NotConfiguredHandler[Req, Res]
+  extends EndpointHandler[Req, Res] {
+  override def apply(
+    unused: Req
+  )(implicit session: RequestSession): Future[Res] = {
+    Future.exception(NotImplemented("Service Start has not been configured correctly"))
```
jsancio: Difficult to say if the configuration is correct or not. We could just say: "Cosmos was not configured to support this operation: $operation". Later on we can provide information on how to configure it. That means that $operation should be passed in.

cosmos-server/src/it/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
```diff
@@ -204,4 +202,28 @@ object CosmosIntegrationTestClient extends Matchers {
...
+  }
+
+  object StagedPackageStorageClient {
+    val configuration: ObjectStorageUri = {
+      getClientProperty("StagedPackageStorageClient", "stagedPackageUri")
+        .flatMap(ObjectStorageUri.parse).get()
+    }
+  }
+
+  private[this] def getClientProperty(clientName: String, key: String): Try[String] = {
```
jsancio: We could just return `getClientProperty(...): String` and throw. Looks like all of the call sites do a `.get`

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -216,14 +240,22 @@ object Cosmos extends FinchServer {
     packageRunner: PackageRunner,
     sourcesStorage: PackageSourcesStorage,
     packageStorage: PackageStorage,
-    universeClient: UniverseClient
+    universeClient: UniverseClient,
+    objectStorage: Option[ObjectStorage]
```
jsancio: We want this type to be `PackageObjectStorage`. This part of the code should never access the `ObjectStorage` directly. We are going to do the same this for the `StagedObjectStorage`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
jsancio: Actually, see my other comment but I think the type here should be `Option[PackageObjectStorage]`.

cosmos-server/src/it/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
cruhland: Yeah, I used `Try` to be explicit that the method can throw (I wrote this part of the code). I guess since it's just test code, we don't plan to recover from the failure, so getting rid of the `Try` makes sense.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: You can configure Intellij's "Organize Imports" feature to never introduce wildcard imports, let me know if you want be to show you where that setting lives.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: You could use `map` to eliminate some pattern-matching noise:

```scala
val objectStorage = packageStorageUri().map {
  case S3Uri(uri) => S3ObjectStorage(new AmazonS3Client(), uri)
  case FileUri(path) => LocalObjectStorage(path)
}
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/ObjectStorageUri.scala
```diff
@@ -0,0 +1,27 @@
...
+
+object ObjectStorageUri {
+  def parse(s: String): Try[ObjectStorageUri] = {
+    val s3UriStart = "s3://"
+    val fileUriStart = "file://"
+    s match {
+      case uri if uri.toLowerCase.startsWith(s3UriStart) =>
+        Try(S3Uri(new AmazonS3URI(uri)))
+      case uri if uri.toLowerCase.startsWith(fileUriStart) =>
+        Try(FileUri(Paths.get(uri.drop(fileUriStart.length))))
```
cruhland: Instead of using `startsWith` and `drop`, you might first parse the input string as a `com.netaporter.uri.Uri` (there's an implicit conversion for it, so you could just make the argument type for this method `Uri` and the compiler should do the rest). Then you could reference fields like `scheme` and `path` directly.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/NotConfiguredHandler.scala
cruhland: Agree, the `Service Start` in the error message should come from a parameter.

cosmos-server/src/main/scala/com/mesosphere/cosmos/ObjectStorageUri.scala
larioj: @cruhland and I discussed the parse implementation, and we're delegating this to the Uri.parse, and whatever it accepts for the scheme.

PR #542, opened 2016-11-15
Close connections to universe

jsancio:
We need to close the InputStream and disconnect the connection after
every request to universe.


jsancio:
I am in the process of verifying this change. I am going to let it run for a bit. I will run the same test against master and report my findings.

jsancio:
@cruhland This doesn't fix the running file descriptor problem a user encounter. That suspect that has to do with the AdminRouter creating connection against Cosmos and not closing them.

PR #544, opened 2016-11-18
Wire the installer into the processor

jsancio:


jsancio:
Not ready for review. Just testing it in DC/OS.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/LocalPackage.scala
```diff
@@ -94,6 +94,7 @@ object LocalPackage {
   }
 }
 
+// TODO: Replace every instance to PackageDefinition with V3Package
```
jsancio: I am going to do this right after this PR.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -178,50 +178,9 @@ private[cosmos] final class Cosmos(
...
-        marathonPackageRunner,
-        sourcesStorage,
-        UniverseClient(adminRouter),
-        installQueue,
-        objectStorages
-      )
-      cosmos.service
-    }
-    boot.get
+    startCosmos(statsReceiver.scope("cosmos"))
```
jsancio: Re-organized this to meet our scalastyle. Method was too long.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -328,12 +328,12 @@ final class InstallQueue private(
...
-              Try(Envelope.decodeData[OperationStatus](childData.getData)).map { operationStatus =>
-                coordinate -> operationStatus
-              }
-            }
+            for {
+              coordinate <- BijectionUtils.scalaTryToTwitterTry(
+                encodedPackageCoordinate.as[scala.util.Try[PackageCoordinate]]
+              )
+              operationStatus <- Try(Envelope.decodeData[OperationStatus](childData.getData))
+            } yield coordinate -> operationStatus
```
jsancio: Reformatted this to be less than 100 chars per line.

jsancio:
@cruhland @larioj Ready for review.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/LocalPackage.scala
cruhland: Ah, are we enforcing the `V3Package` restriction even for Universe packages? In my PR (#545) I use `PackageDefinition` for those.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
```diff
@@ -49,38 +51,59 @@ final class PackageAddSpec extends FreeSpec with BeforeAndAfterAll with BeforeAn
...
+        newV3Package,
+        decodeAndValidateResponse(
+          CosmosClient.submit(packageAddRequest(buildPackage(metadata)))
+        )
+      )
+
+      // Wait until the install queue is empty
+      Await.result(eventualFutureNone(installQueue.next))
+
+      // Assert that the externalized state doesn't change
```
cruhland: This is because the processor will not install a package that's already installed?

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
```diff
@@ -91,61 +114,58 @@ final class PackageAddSpec extends FreeSpec with BeforeAndAfterAll with BeforeAn
...
-          installQueue.success(operation.packageCoordinate)
-          Future.value(operation)
-        case _ => Future.exception(new AssertionError("Pending operation expected"))
-      }
+  private[this] def eventualFuture[T](
+    future: () => Future[Option[T]]
+  ): Future[T] = {
+    future().flatMap {
+      case Some(value) => Future.value(value)
+      case None => eventualFuture(future)
```
cruhland: Could you add `@annotation.tailrec` to this method and the one above? Also, shouldn't there be a short delay between calls to the passed `future` thunk, to avoid polling too rapidly?

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/LocalPackage.scala
jsancio: Yes, we should only support V3Package even if they come from Universe. We can talk about it offline on Monday.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
jsancio: Yep. We need to do a better job of communicating this in the future but currently we don't have a mechanism for doing this.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
jsancio: re the polling frequency. I thought about throttling but decided against it since this only happens in the tests and we should try to run tests as fast as possible. What do you think?

jsancio:
Thanks for the review!

cruhland:
@jsancio merge when you're ready.

PR #545, opened 2016-11-24
Enable POST /package/add to accept requests for Universe packages

cruhland:
Request info:
- `POST /package/add`
- `Accept: application/vnd.dcos.package.add-response+json;charset=utf-8;version=v1`
- `Content-Type: application/vnd.dcos.package.add-request+json;charset=utf-8;version=v1`
- Body is a JSON object with `packageName` and optional `packageVersion` fields

Response is a JSON encoded `universe.v3.model.PackageDefinition`.


cosmos-finch/src/test/scala/com/mesosphere/cosmos/finch/RequestValidatorsSpec.scala
```diff
@@ -1,85 +1,273 @@
...
+import io.finch.items.HeaderItem
+import org.scalacheck.Gen
 import org.scalatest.FreeSpec
-
+import org.scalatest.Matchers
+import org.scalatest.prop.PropertyChecks
 import scala.util.Random
 
-final class RequestValidatorsSpec extends FreeSpec {
+final class RequestValidatorsSpec extends FreeSpec with Matchers with PropertyChecks {
```
cruhland: This is the test class I was trying to clean up, while adding test cases for the chunk-encoded request. Since we're not using that approach anymore, the only changes here are additional tests for existing validators.

cosmos-finch/src/test/scala/com/mesosphere/cosmos/http/CosmosRequest.scala
```diff
@@ -0,0 +1,93 @@
...
+import com.twitter.finagle.http.Method
+import com.twitter.finagle.http.Request
+import com.twitter.finagle.http.Version.Http11
+import com.twitter.io.Buf
+import com.twitter.io.Reader
+import io.circe.Encoder
+import io.circe.syntax._
+import java.io.InputStream
+
+case class CosmosRequest(
```
cruhland: Moved this from `cosmos-server` so I could use it in `RequestValidatorsSpec`.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
```diff
@@ -168,15 +216,35 @@ final class PackageAddSpec extends FreeSpec with BeforeAndAfterAll with BeforeAn
...
+  def packageAddRequest(requestBody: rpc.v1.model.UniverseAddRequest): CosmosRequest = {
     CosmosRequest.post(
       path = "package/add",
-      body = packageBytes,
-      contentType = MediaTypes.PackageZip,
-      accept = MediaTypes.universeV3Package,
-      customHeaders = Map("X-Dcos-Content-Length" -> packageData.length.toString)
+      body = requestBody,
+      contentType = MediaTypes.AddRequest,
+      accept = MediaTypes.AddResponse
```
cruhland: @larioj good news for your CLI changes :)

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -32,18 +35,27 @@ final class PackageAddHandler(
...
+          v3Package <- readPackageDefinitionFromStorage(stagedPackageId)
+        } yield (v3Package, Install(stagedPackageId, v3Package))
+    }
 
-      packageCoordinate = rpc.v1.model.PackageCoordinate(v3Package.name, v3Package.version)
-      operation = Install(stagedPackageId, v3Package)
+    for {
+      (packageDef, operation) <- packageAndOperation
+      packageCoordinate = rpc.v1.model.PackageCoordinate(packageDef.name, packageDef.version)
       _ <- producerView.add(packageCoordinate, operation)
```
cruhland: Oops, forgot to look at the result of this call. Will update.

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/IncompatibleContentTypeHeader.scala
```diff
@@ -0,0 +1,28 @@
...
+import com.mesosphere.cosmos.http.MediaType
+import com.twitter.finagle.http.Fields
+import com.twitter.finagle.http.Status
+import io.circe.JsonObject
+import io.circe.syntax._
+
+// TODO package-add: unify with with IncompatibleAcceptHeader
+case class IncompatibleContentTypeHeader(available: Set[MediaType], specified: MediaType)
+  extends RequestError {
+  val errType: String = "not_valid"
```
jsancio: How about `val errType = getClass.getSimpleName`? That is what we do with the other `CosmosError` or `RequestError`. Before, this was enough to guarantee uniqueness but that is not the case anymore because now we have errors in different namespaces. Should we file an issue for this?

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/IncompatibleContentTypeHeader.scala
```diff
@@ -0,0 +1,28 @@
...
+      "type" -> "header".asJson,
+      "name" -> Fields.ContentType.asJson
+    )).asJson,
+    "specified" -> specified.show.asJson,
+    "available" -> available.map(_.show).asJson
+  )))
+
+  val getHeaders: Map[String, String] = Map.empty
+
+  override def getMessage: String = s"${getClass.getSimpleName}($available, $specified)"
```
jsancio: I don't remember our `RequestError => ErrorResponse` code but do we need `getMessage` to be human readable? If so, how about `"Incompatible content type header. Specified \"{}\" but only "...", "...\" are supported"`?

cosmos-finch/src/test/scala/com/mesosphere/cosmos/finch/RequestValidatorsSpec.scala
```diff
@@ -107,73 +299,123 @@ final class RequestValidatorsSpec extends FreeSpec {
...
+      headers <- genHeaders
+      body <- genBody
+    } yield CosmosRequest(method, path, headers, body)
+  }
+
+  def genMethod: Gen[Method] = Gen.alphaStr.map(Method(_))
+
+  def genPath: Gen[String] = {
+    Gen.listOf(Gen.frequency((10, Gen.alphaNumChar), Gen.freqTuple((1, '/')))).map(_.mkString)
+  }
```
jsancio: Love generator. 👍 . In the other places I have been doing `${type}Gen`. Here we are doing `gen$type`. Let's pick one or the other and run with it.

jsancio: Also, these should be `val`s except for `genRequest`.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/DescribeRequest.scala
```diff
@@ -4,5 +4,5 @@ import com.mesosphere.universe.v2.model.PackageDetailsVersion
 
 case class DescribeRequest(
   packageName: String,
-  packageVersion: Option[PackageDetailsVersion]
+  packageVersion: Option[PackageDetailsVersion] = None
```
jsancio: What's the reason for this change?

jsancio:
@cruhland Can you resolve the merge conflict? I want to create a DC/OS cluster that includes this change and `master`.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/AddResponse.scala
```diff
@@ -4,4 +4,5 @@ import com.mesosphere.universe
 
 // This class is only a wrapper for the underlying package
 // We need a unique type so that the media type encoder can be resolved implicitly
-final class AddResponse(val v3Package: universe.v3.model.V3Package) extends AnyVal
+// TODO package-add: Change argument to a V3Package?
+final class AddResponse(val packageDefinition: universe.v3.model.PackageDefinition) extends AnyVal
```
jsancio: I just realized that this kind of patterns will make it difficult for us to automatically auto-generate API documentation.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
```diff
@@ -87,6 +88,39 @@ final class PackageAddSpec extends FreeSpec with BeforeAndAfterAll with BeforeAn
...
+      val addRequest = rpc.v1.model.UniverseAddRequest("cassandra", packageVersion = None)
+      assertSuccessfulUniverseAdd(addRequest)
+    }
+
+    // TODO package-add: Fix this to not use a v2Package
+    "by name and version" ignore {
+      val version = universe.v3.model.PackageDefinition.Version("0.2.0-2")
+      val addRequest = rpc.v1.model.UniverseAddRequest("cassandra", Some(version))
+      assertSuccessfulUniverseAdd(addRequest)
+    }
```
jsancio: Should we document/test what happens if you install the same package twice. E.g. it is a noop or first package to install wins. Never mind, this is hard to test with Universe packages since they are already unique.

cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/MediaTypedBodyParsers.scala
```diff
@@ -0,0 +1,20 @@
...
+import com.twitter.util.Return
+
+object MediaTypedBodyParsers {
+
+  implicit val packageAddBodyParser: DispatchingMediaTypedBodyParser[AddRequest] = {
+    DispatchingMediaTypedBodyParser(
+      MediaTypes.AddRequest -> DispatchingMediaTypedBodyParser.parserFromDecoder[AddRequest],
+      UMediaTypes.PackageZip -> (bodyBytes => Return(UploadAddRequest(bodyBytes)))
+    )
+  }
```
jsancio: This looks like a type class to me! Let's have a conversation about using companion objects in general and using this in all of our http handlers in specific.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
```diff
@@ -31,6 +32,10 @@ final class LocalObjectStorage private (
 
   private[this] val stats = statsReceiver.scope(s"LocalObjectStorage($path)")
 
+  // TODO package-add: Make this a sibling of the storage directory
+  private[this] val scratchDir = path.resolve(".scratch")
```
jsancio: We can call it `.tmp`. It seems to be a standard name for temporary directories.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
```diff
@@ -129,19 +134,20 @@ final class LocalObjectStorage private (
...
-        StandardOpenOption.WRITE,
-        StandardOpenOption.TRUNCATE_EXISTING
+        scratchPath,
+        StandardOpenOption.CREATE_NEW,
+        StandardOpenOption.WRITE
       )
     )
 
-    try {
+    val metadataBytes = metadata.asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
```
jsancio: This can throw, no? We should do this before opening the file stream or move it into the `try`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
```diff
@@ -153,6 +159,10 @@ final class LocalObjectStorage private (
     } finally {
       outputStream.close()
     }
+
+    Files.move(scratchPath, absolutePath, StandardCopyOption.ATOMIC_MOVE)
```
jsancio: Also, if the move fails we leave the old file around. We should do this inside a try and handle the exception accordingly.

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/IncompatibleContentTypeHeader.scala
cruhland: This was just copied from `IncompatibleAcceptHeader`'s definition. I don't really know where `errType` is used, but I'm fine with changing it if it's harmless. I'll look into it.

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/RequestValidators.scala
```diff
@@ -26,27 +25,40 @@ object RequestValidators {
...
+  def selectedBody[Req, Res](implicit
+    accepts: DispatchingMediaTypedBodyParser[Req],
+    produces: DispatchingMediaTypedEncoder[Res]
   ): Endpoint[EndpointContext[Req, Res]] = {
-    val sessionValidator = header("Content-Type").as[MediaType].map { contentType =>
-      RequestSession(authorization = None, contentType = Some(contentType))
-    }
+    val contentTypeValidator = header(Fields.ContentType)
+      .as[MediaType]
+      .mapAsync { contentType =>
```
cruhland: I'm not sure, since the `.mapAsync` method is on `Endpoint` and I don't think the corresponding `.map` method will capture the exception correctly. I can try it out in the REPL and see...

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/RequestValidators.scala
```diff
@@ -26,27 +25,40 @@ object RequestValidators {
...
-    validators.map { case bufStream :: bodySize :: session :: HNil =>
-      // TODO package-add: Better error handling for request data extraction (e.g. toLong)
-      EndpointContext(toReq(bufStream, bodySize.toLong), session, resEncoder)
+    val allValidators = baseValidator(produces) :: contentTypeValidator :: binaryBody
+    allValidators.mapAsync {
+      case authorization :: responseEncoder :: contentType :: bodyParser :: bodyBytes :: HNil =>
+        Future.const(bodyParser(bodyBytes)).map { requestBody =>
+          val session = RequestSession(authorization, Some(contentType))
+          EndpointContext(requestBody, session, responseEncoder)
+        }
```
cruhland: Ah yeah, good catch!

cruhland: Hmm, actually the reason I'm using `Future.const` here is because `bodyParser` has type `Array[Byte] => Try[Req]`. So the easiest thing to do is wrap that `Try` with a `Future`. If I can determine that `Endpoint.map` captures exceptions, I'll update this to call `.get` on the `Try` instead.

cosmos-finch/src/test/scala/com/mesosphere/cosmos/finch/RequestValidatorsSpec.scala
```diff
@@ -1,85 +1,273 @@
...
+    ): Endpoint[EndpointContext[Json, Json]] = {
+      val mediaTypedDecoder = MediaTypedDecoder[Json](expectedContentTypeHeader)
+      val accepts = MediaTypedRequestDecoder(mediaTypedDecoder)
+      val produces = DispatchingMediaTypedEncoder[Json](expectedAcceptHeader)
+
+      RequestValidators.standard(accepts, produces)
+    }
+
+    def genTestData(
+      actualContentTypeHeader: Option[Option[String]] = None
```
cruhland: Yeah, I agree this is confusing. I'll see what I can do to clean it up.

cosmos-finch/src/test/scala/com/mesosphere/cosmos/finch/RequestValidatorsSpec.scala
cruhland: I don't have a strong preference for the format of the name, but I like `genType` because it reads the same as `Gen[Type]`. It's also easier to see at a glance which variables define generators, because the `gen`s in their names line up. Unless you'd rather see the other format, I'll go ahead and normalize the names.

cosmos-http/src/test/scala/com/mesosphere/cosmos/http/MediaTypeSpec.scala
```diff
@@ -78,4 +83,26 @@ class MediaTypeSpec extends FreeSpec {
...
+
+}
+
+object MediaTypeSpec {
+
+  // TODO package-add: Make this more general
+  val genMediaType: Gen[MediaType] = {
+    val genTry = for {
+      typePart <- Gen.alphaStr.map(_.toLowerCase)
+      subTypePart <- Gen.alphaStr.map(_.toLowerCase)
```
cruhland: Unfortunately it's not in the version we're using, 1.12.5. Want me to try updating it?

cosmos-finch/src/test/scala/com/mesosphere/cosmos/finch/RequestValidatorsSpec.scala
jsancio: Sounds good. I like your naming.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/DescribeRequest.scala
cruhland: Must have been leftover from an early change. I reverted it and everything still compiles.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
```diff
@@ -87,6 +74,39 @@ final class PackageAddSpec extends FreeSpec with BeforeAndAfterAll with BeforeAn
...
   }
 
+  "/package/add of a universe package" - {
+
+    "by name only" in {
+      val addRequest = rpc.v1.model.UniverseAddRequest("cassandra", packageVersion = None)
+      assertSuccessfulUniverseAdd(addRequest)
+    }
+
+    // TODO package-add: Fix this to not use a v2Package
```
cruhland: I need to address this before the PR gets merged.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
```diff
@@ -118,25 +138,29 @@ final class PackageAddSpec extends FreeSpec with BeforeAndAfterAll with BeforeAn
     cleanObjectStorage(packageObjectStorage)
   }
 
+  // TODO package-add: Is this needed anymore?
```
cruhland: Looking into this...

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -223,6 +223,7 @@ object Encoders extends LowPriorityImplicits {
     case IncompatibleAcceptHeader(available, _) =>
       "Item 'header 'Accept'' deemed invalid by rule: 'should match one of: " +
       s"${available.map(_.show).mkString(", ")}'"
+    // TODO package-add: case for IncompatibleContentTypeHeader
```
cruhland: This needs to be addressed in this PR

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -351,6 +352,7 @@ object Encoders extends LowPriorityImplicits {
       case EnvelopeError(msg) => msg
       case InstallQueueError(msg) => msg
       case NotImplemented(msg) => msg
+      case OperationInProgress(_) => ??? // TODO package-add: Ensure this gets implemented
```
cruhland: This needs to be addressed in this PR

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -32,15 +34,30 @@ final class PackageAddHandler(
...
-    val packageData = ReaderInputStream(Reader.concat(request.packageData.map(Reader.fromBuf)))
-    val Some(contentType) = session.contentType
+    val packageAndOperation = request match {
+      case rpc.v1.model.UniverseAddRequest(packageName, packageVersion) =>
+        packageCollection.getPackageByPackageVersion(packageName, packageVersion)
+          .map { case (packageDef, _) =>
+            packageDef match {
+              case v3Package: universe.v3.model.V3Package => (v3Package, UniverseInstall(v3Package))
+              // TODO package-add: Define CosmosError for this case
+              case v2Package: universe.v3.model.V2Package => ???
```
cruhland: This needs to be addressed in this PR

cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v1/circe/MediaTypedBodyParsers.scala
cruhland: Tomorrow in standup?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
cruhland: Ah, is that why it was at the top of the method? I can move it back, but I'm actually not sure if it can throw an exception -- it's just converting a `Map[String, String]` into a JSON string.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
cruhland: Good catch! Will update.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/AddResult.scala
```diff
@@ -1,5 +0,0 @@
-package com.mesosphere.cosmos.storage.installqueue
-
-sealed trait AddResult
-case object Created extends AddResult
-case object AlreadyExists extends AddResult
```
cruhland: I deleted this type because it was conflating a success state (`Created`) with a failure state (`AlreadyExists`). Since `ProducerView.add()` already returns a `Future`, we can cover the `AlreadyExists` case with an exception.

cosmos-server/src/test/scala/com/mesosphere/cosmos/handler/PackageAddHandlerSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+        val stagedPackageStorage = StagedPackageStorage(stagedObjectStorage)
+        new PackageAddHandler(packageCollection, stagedPackageStorage, producerView)
+      }
+
+      def assertErrorResponse(
+        response: Future[rpc.v1.model.AddResponse],
+        expectedCoordinate: rpc.v1.model.PackageCoordinate
+      ): Unit = {
+        val inProgress = intercept[OperationInProgress](Await.result(response))
+        assertResult(Status.Conflict)(inProgress.status)
```
cruhland: I imitated Marathon's behavior of responding with a 409.

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/RequestValidators.scala
cruhland: Ok, confirmed that `.map` works. Updating...

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/RequestValidators.scala
cruhland: Will update.

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/IncompatibleContentTypeHeader.scala
cruhland: [Here's](https://github.com/dcos/cosmos/blob/958c3a937800aeadc4e2703d830bf9e6c54bac4c/cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala#L193) the only place where `errType` is used. In that method you can see that `not_valid` is also used as the `errType` of Finch's `Error.NotValid`, which is produced by `Endpoint` validators.

Currently, `IncompatibleAcceptHeader` and `IncompatibleContentTypeHeader` are the only subclasses of `RequestError` that are not also subclasses of `CosmosError`. On the one hand, since those errors are only produced when validating requests, I can see why we might want to give them an `errType` of `not_valid`. But I think overloading the `not_valid` type in this way isn't very helpful to clients, who will now have to parse the other fields of the `ErrorResponse` in order to distinguish errors.

TL;DR: I'll change these `Incompatible*Header` types to use `getClass.getSimpleName`, and file an issue for ensuring that all of our errors have unique names.

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/IncompatibleContentTypeHeader.scala
cruhland: Confirmed that `getMessage` is not used.

cruhland: I'll make `getMessage` human-readable anyway, since it can't hurt.

cosmos-finch/src/main/scala/com/mesosphere/cosmos/finch/IncompatibleContentTypeHeader.scala
cruhland: There's actually a [test](https://github.com/dcos/cosmos/blob/8cbf16dca6a9520ef4f84fe3391589d1b1095195/cosmos-server/src/it/scala/com/mesosphere/cosmos/handler/RequestErrorsSpec.scala#L57) for the `errType` value, so I'm going to leave it alone for now.

cosmos-http/src/test/scala/com/mesosphere/cosmos/http/MediaTypeSpec.scala
cruhland: We're stuck on the 1.12.x series until ScalaTest updates the version they depend on: https://github.com/rickynils/scalacheck/issues/217#issuecomment-210070553

I'll bump our version to 1.12.6, though.

cruhland:
@jsancio all comments addressed! Back to you.

jsancio:
@cruhland Thanks. Looking.

cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
```diff
@@ -6,7 +6,6 @@ import com.twitter.bijection.Bijection
 import com.twitter.bijection.Conversion
 import com.twitter.bijection.Conversion.asMethod
 import com.twitter.bijection.Injection
-import java.time.Instant
```
jsancio: :+1: for some reason sbt is not warning for unused imports. It warns for unused wildcard imports.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
```diff
@@ -59,34 +55,48 @@ final class PackageAddSpec extends FreeSpec with BeforeAndAfterAll with BeforeAn
...
+    def assertSuccessfulUniverseAdd(addRequest: rpc.v1.model.UniverseAddRequest): Unit = {
+      val expectedPackage = describePackage(addRequest.packageName, addRequest.packageVersion)
+
+      val request = packageAddRequest(addRequest)
+      val response = CosmosClient.submit(request)
+
+      assertResult(Status.Accepted)(response.status)
+      assertResult(MediaTypes.AddResponse)(MediaType.parse(response.contentType.get).get())
+
+      val v3Package = decode[rpc.v1.model.AddResponse](response.contentString).v3Package
```
jsancio: I am more interested in testing that this de-serializes as a `universe.v3.model.V3Package`. What do you think?

cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
cruhland: Yeah scalac is not very good at it, somehow IntelliJ does it better!

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
cruhland: Good point, I'll update it and then merge.

PR #553, opened 2016-12-07
Enable InstallQueueSpec tests with suite-specific ZK test cluster

cruhland:
The problems with Curator turned out to be a red herring, although there is Curator code that I'd say contains a bug. See my comment below for more info.

### DIFF
--- a/cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
+++ b/cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
@@ -2,29 +2,46 @@ package com.mesosphere.cosmos.storage.installqueue
 
 import com.mesosphere.cosmos.InstallQueueError
 import com.mesosphere.cosmos.OperationInProgress
+import com.mesosphere.cosmos.model.ZooKeeperUri
 import com.mesosphere.cosmos.rpc.v1.model.ErrorResponse
 import com.mesosphere.cosmos.rpc.v1.model.PackageCoordinate
 import com.mesosphere.cosmos.storage.Envelope
 import com.mesosphere.cosmos.storage.v1.circe.MediaTypedDecoders._
 import com.mesosphere.cosmos.storage.v1.circe.MediaTypedEncoders._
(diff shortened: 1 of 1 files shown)


cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueueSpec.scala
```diff
@@ -531,14 +547,9 @@ class InstallQueueSpec extends fixture.FreeSpec {
...
-    val path = "/"
-    val retries = 10
-    val baseSleepTime = 1000
-    val client = CuratorFrameworkFactory.newClient(
-      path,
-      new ExponentialBackoffRetry(baseSleepTime, retries)
-    )
-    client.start()
+    val namespace = getClass.getSimpleName
+    val zkUri = ZooKeeperUri.parse(s"zk://${zkCluster.getConnectString}/$namespace").get()
```
cruhland: This line is what caused all the trouble. Our code for `ZooKeeperUri.parse` requires there to be a non-empty path component. Initially I weakened the validation so that I could use an empty path, but that is what led to the buggy behavior from Curator. So I'm now using the test name as a namespace, and I'm convinced that we should probably always use namespaces with ZK anyway, as a precaution -- I don't think the URI validation logic should be changed.

Now for the Curator bug. When Curator gets an event from ZK, it strips the namespace from the front of the path, so that the paths it reports don't show what the namespace is. The "namespace-relative" paths are still supposed to be absolute paths, and they usually are. But it turns out that if you configured Curator to use an _empty_ namespace, then the stripping logic actually removes the leading slash from the path! Here's where it happens: https://github.com/apache/curator/blob/master/curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceImpl.java#L69

cruhland: Submitted PR to the Curator project: https://github.com/apache/curator/pull/176

jsancio: :+1:

PR #558, opened 2016-12-20
Fix package ordering

jsancio:


cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -118,7 +118,13 @@ final class DefaultUniverseClient(
           }
         } map { case (contentType, bodyInputStream) =>
           try {
-            decodeUniverse(contentType, bodyInputStream, repository.uri)
+            universe.v3.model.Repository(
+              decodeUniverse(
+                contentType,
+                bodyInputStream,
+                repository.uri
+              ).packages.sorted.reverse
```
cruhland: The `decodeUniverse` method already returns a `universe.v3.model.Repository`, but you're unwrapping the packages, sorting them, and then wrapping them in `Repository` again. Should the sorting be moved into the `decodeUniverse` method instead?

jsancio:
Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: @cruhland Done. Thanks for the review!

cruhland:
LGTM when tests pass!

PR #565, opened 2017-01-26
Add getCreationTime to ObjectStorage

larioj:
This PR will add a method getCreationTime to the ObjectStorage. This will be used for the garbage collection algorithm to determine whether a staging object should be removed or not.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
@@ -92,7 +92,7 @@ final class PackageAddHandler(
         if (zipEntry.isEmpty) {
           throw new Error("metadata.json not found in zip file")
         }
-        !zipEntry.map(_.getName).contains("matadata.json")
+        !zipEntry.map(_.getName).contains("metadata.json")
       }){}
 
       val metadataBytes = StreamIO.buffer(packageZip).toByteArray
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
@@ -12,6 +12,7 @@ import com.mesosphere.cosmos.storage.v1.model.UniverseInstall
 import com.mesosphere.universe
 import com.mesosphere.universe.bijection.UniverseConversions._
 import com.mesosphere.universe.v3.circe.Decoders._
+import com.mesosphere.universe.v3.syntax.PackageDefinitionOps._
 import com.twitter.bijection.Conversion.asMethod
 import com.twitter.finagle.http.Status
 import com.twitter.io.StreamIO
@@ -34,12 +35,12 @@ final class PackageAddHandler(
   override def apply(request: rpc.v1.model.AddRequest)(implicit
     session: RequestSession
   ): Future[rpc.v1.model.AddResponse] = {
-    val packageAndOperation = request match {
+    val futureOperation = request match {
       case rpc.v1.model.UniverseAddRequest(packageName, packageVersion) =>
         packageCollection.getPackageByPackageVersion(packageName, packageVersion)
           .map {
             case (v3Package: universe.v3.model.V3Package, _) =>
-              (v3Package, UniverseInstall(v3Package))
+              UniverseInstall(v3Package)
             case _ =>
               throw new UnsupportedOperationException("Adding a package with packagingVersion 2.0")
           }
@@ -51,15 +52,16 @@ final class PackageAddHandler(
         for {
           stagedPackageId <- stagedPackageStorage.put(packageStream, packageSize, contentType)
           v3Package <- readPackageDefinitionFromStorage(stagedPackageId)
-        } yield (v3Package, Install(stagedPackageId, v3Package))
+        } yield {
+          Install(stagedPackageId, v3Package)
+        }
     }
 
     for {
-      (v3Package, operation) <- packageAndOperation
-      packageCoordinate = rpc.v1.model.PackageCoordinate(v3Package.name, v3Package.version)
-      _ <- producerView.add(packageCoordinate, operation)
+      operation <- futureOperation
+      _ <- producerView.add(operation.v3Package.packageCoordinate, operation)
     } yield {
-      new rpc.v1.model.AddResponse(v3Package)
+      new rpc.v1.model.AddResponse(operation.v3Package)
     }
   }
 
@@ -82,19 +84,21 @@ final class PackageAddHandler(
   private[this] def extractPackageMetadata(
     packageZip: ZipInputStream
   ): Try[universe.v3.model.Metadata] = {
+    // TODO package-add: Factor out common Zip-handling code into utility methods
+    // TODO package-add: Test cases for files with unexpected content
     Try {
-      // TODO package-add: Factor out common Zip-handling code into utility methods
-      Iterator
-        .continually(Option(packageZip.getNextEntry()))
-        .takeWhile(_.isDefined)
-        .flatten
-        // TODO package-add: Test cases for files with unexpected content
-        .filter(_.getName == "metadata.json")
-        .map { _ =>
-          val metadataBytes = StreamIO.buffer(packageZip).toByteArray
-          decode[universe.v3.model.Metadata](new String(metadataBytes, StandardCharsets.UTF_8))
+      while({
+        val zipEntry = Option(packageZip.getNextEntry)
+        if (zipEntry.isEmpty) {
+          throw new Error("metadata.json not found in zip file")
         }
-        .next()
+        !zipEntry.map(_.getName).contains("matadata.json")
+      }){}
+
+      val metadataBytes = StreamIO.buffer(packageZip).toByteArray
+      decode[universe.v3.model.Metadata](
+        new String(metadataBytes, StandardCharsets.UTF_8)
+      )
     } ensure packageZip.close()
   }
 
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
@@ -84,21 +84,19 @@ final class PackageAddHandler(
   private[this] def extractPackageMetadata(
     packageZip: ZipInputStream
   ): Try[universe.v3.model.Metadata] = {
-    // TODO package-add: Factor out common Zip-handling code into utility methods
-    // TODO package-add: Test cases for files with unexpected content
     Try {
-      while({
-        val zipEntry = Option(packageZip.getNextEntry)
-        if (zipEntry.isEmpty) {
-          throw new Error("metadata.json not found in zip file")
+      // TODO package-add: Factor out common Zip-handling code into utility methods
+      Iterator
+        .continually(Option(packageZip.getNextEntry()))
+        .takeWhile(_.isDefined)
+        .flatten
+        // TODO package-add: Test cases for files with unexpected content
+        .filter(_.getName == "metadata.json")
+        .map { _ =>
+          val metadataBytes = StreamIO.buffer(packageZip).toByteArray
+          decode[universe.v3.model.Metadata](new String(metadataBytes, StandardCharsets.UTF_8))
         }
-        !zipEntry.map(_.getName).contains("metadata.json")
-      }){}
-
-      val metadataBytes = StreamIO.buffer(packageZip).toByteArray
-      decode[universe.v3.model.Metadata](
-        new String(metadataBytes, StandardCharsets.UTF_8)
-      )
+        .next()
     } ensure packageZip.close()
   }
 
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
@@ -85,23 +85,27 @@ final class LocalObjectStorage private(storageDir: Path, scratchDir: Path, stats
       pool {
         val fullName = storageDir.resolve(name)
         if (Files.exists(fullName)) {
-          val pathsToDelete =
-            Iterator
-              .iterate(fullName)(_.getParent)
-              .takeWhile(_ != storageDir)
-              .toList
-
-          pathsToDelete.foldLeft(Try(())) { (status, path) =>
-            status.map{ _ =>
-              Files.deleteIfExists(path)
-              ()
-            }
-          }.handle {
-            case _: DirectoryNotEmptyException => ()
-          }.get()
+          deleteFileAndEmptyParents(fullName)
         }
       }
     }
+  }
+
+  private[this] def deleteFileAndEmptyParents(fullName: Path): Unit = {
+    val pathsToDelete =
+      Iterator
+        .iterate(fullName)(_.getParent)
+        .takeWhile(_ != storageDir)
+        .toList
+
+    pathsToDelete.foldLeft(Try(())) { (status, path) =>
+      status.map { _ =>
+        Files.deleteIfExists(path)
+        ()
+      }
+    }.handle {
+      case _: DirectoryNotEmptyException => ()
+    }.get()
   }
 
   override def list(directory: String): Future[LocalObjectStorage.ObjectList] = {
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
@@ -22,6 +22,7 @@ import java.nio.file.Path
 import java.nio.file.StandardCopyOption
 import java.nio.file.StandardOpenOption
 import java.nio.file.attribute.BasicFileAttributes
+import java.time.Instant
 import scala.collection.JavaConverters._
 import scala.collection.breakOut
 import scala.util.control.NonFatal
@@ -136,13 +137,13 @@ final class LocalObjectStorage private(storageDir: Path, scratchDir: Path, stats
 
   override def getUrl(name: String): Option[Uri] = None
 
-  override def getCreationTime(name: String): Future[Option[Long]] = {
+  override def getCreationTime(name: String): Future[Option[Instant]] = {
     pool {
       val absolutePath = storageDir.resolve(name)
       val attr = Try {
         Files.readAttributes(absolutePath, classOf[BasicFileAttributes])
       }
-      attr.toOption.map(_.creationTime().toMillis)
+      attr.toOption.map(_.creationTime().toInstant)
     }
   }
 
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
@@ -15,11 +15,13 @@ import java.io.DataInputStream
 import java.io.DataOutputStream
 import java.io.InputStream
 import java.nio.charset.StandardCharsets
+import java.nio.file.DirectoryNotEmptyException
 import java.nio.file.Files
 import java.nio.file.NoSuchFileException
 import java.nio.file.Path
 import java.nio.file.StandardCopyOption
 import java.nio.file.StandardOpenOption
+import java.nio.file.attribute.BasicFileAttributes
 import scala.collection.JavaConverters._
 import scala.collection.breakOut
 import scala.util.control.NonFatal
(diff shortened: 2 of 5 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -82,19 +84,21 @@ final class PackageAddHandler(
...
-        .map { _ =>
-          val metadataBytes = StreamIO.buffer(packageZip).toByteArray
-          decode[universe.v3.model.Metadata](new String(metadataBytes, StandardCharsets.UTF_8))
+      while({
+        val zipEntry = Option(packageZip.getNextEntry)
+        if (zipEntry.isEmpty) {
+          throw new Error("metadata.json not found in zip file")
         }
-        .next()
+        !zipEntry.map(_.getName).contains("matadata.json")
```
jsancio: `s/matadata/metadata/`

jsancio: Scala has `do...while`. I think this code is easier to read if you use `do...while` instead of this kind of `while` loop.

larioj: Thanks. This is a bad bug. But at least easy to catch.

larioj: Note that this change doesn't have to happen. I just thought that the previous version was a bit confusing, and wanted to see whether you guys thought this was an improvement. I'm fine with reverting it back to the original version.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -14,31 +17,147 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+            _ <- Future(assertResult(List())(objects.directories))
+          } yield ()
+        }
+      }
+    }
+  }
+
+  "write() must not change the case of paths" in {
+    val name1 = "B/nR/xyharPhq"
+    val name2 = "b/5nRzknggv/ilygo3oj"
```
jsancio: We don't need this write for this test, right?

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -48,20 +167,100 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+    }
+  }
+
+  def pathsConflict(paths: List[String]): Boolean = {
+    paths.combinations(2).exists { case List(a, b) =>
+      a.startsWith(b) || b.startsWith(a)
+    }
+  }
+
+  implicit class ObjectStorageNoStreams(objectStorage: LocalObjectStorage) {
```
jsancio: Try `implicit class ...(val objectStorage: ...) extends AnyVal ...`: http://docs.scala-lang.org/overviews/core/value-classes.html#extension-methods

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorage.scala
```diff
@@ -70,7 +70,13 @@ trait ObjectStorage {
   /**
    * Get the URL for an object if the backing store supports it.
    */
-  def getUrl(name: String): Option[Uri]
+  def getUrl(name: String): Future[Option[Uri]]
```
larioj: I thought it would be less restrictive for it to be a future. Otherwise we are asking every implementation of ObjectStorage to have a quick getUrl method. It also makes the api fully asynchronous.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
larioj: Yeah, you're right.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
jsancio: For now this is okay because we are only looking for one file. In the future we are going to process all the files so that algorithm will change then.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorage.scala
jsancio: We can change that when that happens. Let's not try to predict the future (no pun intended). The nice thing about being sync is that it makes it explicit that the client can compute the URL without having to ask the "server".

larioj:
@jsancio Yes, this does not include GC.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
```diff
@@ -81,14 +84,31 @@ final class LocalObjectStorage private(storageDir: Path, scratchDir: Path, stats
   override def delete(name: String): Future[Unit] = {
     Stat.timeFuture(stats.stat("delete")) {
       pool {
-        // Drop the boolean. We want to have the same semantic as S3 which succeeds in either case.
-        val _ = Files.deleteIfExists(storageDir.resolve(name))
-
-        // TODO: To have the same semantic as S3 we need to delete all empty parent directories
+        val fullName = storageDir.resolve(name)
+        if (Files.exists(fullName)) {
```
cruhland: Is this check needed? Won't `deleteFileAndEmptyParents` do the right thing if the file does not exist?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
```diff
@@ -81,14 +84,31 @@ final class LocalObjectStorage private(storageDir: Path, scratchDir: Path, stats
...
+        .toList
+
+    pathsToDelete.foldLeft(Try(())) { (status, path) =>
+      status.map { _ =>
+        Files.deleteIfExists(path)
+        ()
+      }
+    }.handle {
+      case _: DirectoryNotEmptyException => ()
+    }.get()
```
cruhland: This might be clearer if we rely on normal exception propagation:

```scala
try {
  pathsToDelete.foreach(Files.deleteIfExists)
} catch {
  case _: DirectoryNotEmptyException => // Do nothing
}
```

Or you can wrap the whole thing in `Try()` if you really want to use it:

```scala
Try(pathsToDelete.foreach(Files.deleteIfExists))
  .handle { case _: DirectoryNotEmptyException => () }
  .get()
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
```diff
@@ -81,14 +84,31 @@ final class LocalObjectStorage private(storageDir: Path, scratchDir: Path, stats
...
       }
     }
   }
 
+  private[this] def deleteFileAndEmptyParents(fullName: Path): Unit = {
+    val pathsToDelete =
+      Iterator
+        .iterate(fullName)(_.getParent)
+        .takeWhile(_ != storageDir)
+        .toList
```
cruhland: We don't really need the `.toList`; `Iterator` has pretty much all the same methods, and we're only doing a single pass.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
```diff
@@ -117,6 +137,16 @@ final class LocalObjectStorage private(storageDir: Path, scratchDir: Path, stats
 
   override def getUrl(name: String): Option[Uri] = None
 
+  override def getCreationTime(name: String): Future[Option[Instant]] = {
+    pool {
+      val absolutePath = storageDir.resolve(name)
+      val attr = Try {
+        Files.readAttributes(absolutePath, classOf[BasicFileAttributes])
+      }
```
cruhland: This `Try` will swallow all exceptions, not just the one indicating that the file doesn't exist. It might be useful for debugging purposes to let the other exceptions propagate via the `Future`. What do you think? cc @jsancio

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/S3ObjectStorage.scala
```diff
@@ -116,12 +119,30 @@ final class S3ObjectStorage(
...
 
+  override def getCreationTime(name: String): Future[Option[Instant]] = {
+    Stat.timeFuture(stats.stat("getCreationTime")) {
+      pool {
+        try {
+          Some(
+            Instant.ofEpochMilli(
+              client
+                .getObjectMetadata(new GetObjectMetadataRequest(bucket, fullPath(name)))
+                .getLastModified.getTime
```
cruhland: Why not use local variables instead of nesting here? There would be fewer lines and they would be less indented.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
larioj: True

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -48,20 +166,100 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
       segments <- Gen.listOfN(numSegments, genSegment)
       path = segments.mkString("/")
-      if isValidPath(path)
-    } yield path
+    } yield path).suchThat(isValidPath)
+  }
+
+  def isValidPath(s: String): Boolean =
+    s.nonEmpty &&
+      !s.forall(_ == '/') &&
```
cruhland: This clause is redundant with the others now.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
larioj: Haha I was overthinking this :)

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
cruhland: Really it's the `toOption` below that's hiding the exceptions...

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -14,31 +18,145 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+  }
+
+  "read() on a deleted file returns Future(None)" in {
+    forAll(genPath, arbitrary[Array[Byte]]) { (path, dataToWrite) =>
+      TestUtil.withLocalObjectStorage { localStorage =>
+        Await.result {
+          for {
+            _ <- localStorage.write(path, dataToWrite)
+            Some((_, dataFromRead)) <- localStorage.readAsArray(path)
+            _ <- Future(assertResult(dataToWrite)(dataFromRead))
```
cruhland: The above two lines are tested in the previous test; let's keep this test focused.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
larioj: That's a good idea.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
cruhland: Hmm, I guess it's not quite the same. You might want to split these lines into a separate test instead.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -14,31 +18,145 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+            _ <- localStorage.delete(path)
+            failedRead <- localStorage.readAsArray(path)
+            _ <- Future(assertResult(None)(failedRead))
```
cruhland: Let's not put test assertions inside `Future`s. There's a risk that the `AssertionError` could be eaten by the `Future` accidentally. Most tests should just have a single assertion at the very end anyway.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -14,31 +18,145 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+            _ <- localStorage.write(path, dataToWrite)
+            Some((mediaType, _)) <- localStorage.readAsArray(path)
+            _ <- Future(assertResult(MediaTypes.applicationOctetStream)(mediaType))
+          } yield ()
+        }
+      }
+    }
+  }
+
+  "delete() removes all empty parent directories" in {
```
cruhland: You might want another test that creates two objects whose names have a common prefix, deletes one, and verifies (1) that the parent directories of the deleted object that are *not* in the common prefix are deleted, and (2) that the other object still exists (and hence all of its parent directories exist).

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -14,31 +18,145 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+            _ <- Future(assertResult(List())(objects.directories))
+          } yield ()
+        }
+      }
+    }
+  }
+
+  "write() and read() must be case sensitive" in {
+    val name1 = "B/nR/xyharPhq"
+    val badName1 = "b/nR/xyharphq"
```
cruhland: Make this a generator-driven test?

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -14,31 +18,145 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+  "list() returns all objects written" in {
+    forAll(genObjectStorageState) { objectStorageState =>
+      TestUtil.withLocalObjectStorage { localStorage =>
+        Await.result {
+          val namesToWrite = objectStorageState.map(_.name)
+          for {
+            _ <- localStorage.writeAll(objectStorageState)
+            namesInStorage <- localStorage.listAll("")
+            _ <- Future(assertResult(namesToWrite.sorted)(namesInStorage.sorted))
+            _ <- localStorage.deleteAll(objectStorageState.map(_.name))
```
cruhland: This line is redundant, no?

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -14,31 +18,145 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+
+  "getCreationTime returns the time the object was created" in {
+    forAll(genPath, arbitrary[Array[Byte]]) { (path, dataToWrite) =>
+      TestUtil.withLocalObjectStorage { localStorage =>
+        Await.result {
+          val tolerance = 2L
+          for {
+            timeBefore <- Future(Instant.now().minusSeconds(tolerance))
+            _ <- localStorage.write(path, dataToWrite)
+            creationTime <- localStorage.getCreationTime(path)
```
cruhland: Here's a crazy idea: if you change `LocalObjectStorage` to use a [`FileSystemProvider`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/spi/FileSystemProvider.html) instead of the static methods on `Files`, then you could do an exact test of `getCreationTime` by initializing the `LocalObjectStorage` instance with a mock `FileSystemProvider`. Might not be worth the effort, but something to keep in mind? cc @jsancio

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -48,20 +166,100 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+      !s.endsWith("/")
+
+  val genMediaType: Gen[MediaType] = {
+    val mediaTypes = Array(
+      MediaTypes.applicationJson,
+      MediaTypes.applicationOctetStream,
+      MediaTypes.OperationStatus,
+      MediaTypes.RepositoryList
+    )
+    Gen.choose(0, mediaTypes.length - 1).map(mediaTypes)
```
cruhland: Using `Gen.oneOf()` would be simpler here.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -48,20 +166,100 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+  val genObjectStorageState: Gen[ObjectStorageState] = {
+    Gen.listOf(genObjectStorageItem).suchThat { state =>
+      !pathsConflict(state.map(_.name))
+    }
+  }
+
+  def pathsConflict(paths: List[String]): Boolean = {
+    paths.combinations(2).exists { case List(a, b) =>
+      a.startsWith(b) || b.startsWith(a)
+    }
```
cruhland: This condition is pretty strong. Won't it say that the paths `foo/bar` and `foo/bart` conflict, when they shouldn't? Splitting all paths on `/` first should fix it.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -48,20 +166,100 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+        }
+      }
+    }
+
+    def writeAll(objectStorageItems: List[ObjectStorageItem]): Future[Unit] = {
+      Future.collect(
+        objectStorageItems.map { objectStorageItem =>
+          write(objectStorageItem.name, objectStorageItem.content, objectStorageItem.mediaType)
+        }
+      ).map(_ => ())
```
cruhland: You can use the `.unit` method instead.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -48,20 +166,100 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
+    }
+
+    def deleteAll(names: List[String]): Future[Unit] = {
+      Future.collect(
+        names.map { name =>
+          objectStorage.delete(name)
+        }
+      ).map(_ => ())
```
cruhland: You can use the `.unit` method instead.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
cruhland: Actually, you could replace `Future.collect` with `Future.join`, then you won't need `.unit`.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
cruhland: Actually, you could replace `Future.collect` with `Future.join`, then you won't need `.unit`.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
cruhland: BTW, I didn't just have `.unit` and `Future.join` memorized; it's just that this is a pretty common idiom so I checked the API docs on a hunch :smile:

cruhland:
@larioj whew, finished commenting! Overall, looks good, but a few things should be updated.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
jsancio: This depends on what this test is testing. It is precise enough to test that timestamp is greater than "test start time" and less than "test end time". Why do we need `tolerance`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/S3ObjectStorage.scala
jsancio: For `LocalObjectStorage` we use `toInstant()`. Let's do the same here: `getLastModified().toInstant()`.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
larioj: It fails without tolerance. I think the filesystem has a second granularity.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
larioj: Done

cruhland:
@larioj LGTM! Nice work on the generator tests :smile:

PR #566, opened 2017-01-26
Add support of InstallQueue to LocalPackageCollection

jsancio:


cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/LocalPackageCollection.scala
```diff
@@ -5,14 +5,37 @@ import com.mesosphere.cosmos.VersionNotFound
...
-    objectStorage.list().map(_.sorted.reverse)
+    for {
+      operatingPackages <- installQueue.viewStatus()
+      storedPackages <- objectStorage.list()
+    } yield {
+      val pendingLocalPackages = operatingPackages.mapValues(
+        LocalPackageCollection.operationStatusToLocalPackage
+      )
+
+      val result = storedPackages.foldLeft(pendingLocalPackages) { (acc, pkg) =>
```
cruhland: Instead of `foldLeft`, I think this should work:

```scala
// Operation queue takes precedence over object storage in the case of conflict
val result = storedPackages.map(pkg => pkg.packageCoordinate -> pkg) ++ pendingLocalPackages
```

cruhland:
@jsancio  LGTM, one minor comment. And it looks like you missed some compilation errors in the tests.

jsancio:
@cruhland Thanks for the review. I address your comments. Back to you.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/LocalPackage.scala
```diff
@@ -115,6 +121,7 @@ final case class Uninstalling(
   data: Either[PackageCoordinate, universe.v3.model.V3Package]
 ) extends LocalPackage
 
+// TODO: We shouldn't return Operation. It contains stagedPackageId which is useless to the client
```
cruhland: Create a JIRA for this, if you're not going to fix it in this PR?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/package.scala
```diff
@@ -27,14 +26,18 @@ package object repository {
...
-    localPackageCollection.getInstalledPackage(
-      packageCoordinate.name,
-      Some(packageCoordinate.version)
-    ).transform {
+  )(
+    pkg: universe.v3.model.V3Package
+  ): Future[Unit] = {
+
+    packageObjectStorage.list().map { packages =>
+      LocalPackageCollection.installedPackage(
```
cruhland: I noticed that this method does a linear search for the package. It probably won't impact performance for a while, but since the package list is in theory sorted, we could perform a more efficient search.

jsancio: Yep. This is not the fastest algorithm. We will revisit this when we have a lot of packages and Alice starts using this.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/LocalPackage.scala
jsancio: I'll create a Jira but I suspect we will deal with this when we implement `package/list` version x.

jsancio: Added a comment in the description in the Jira for implementing `/package/list`.

jsancio:
> @jsancio did we decide how we're going to test this? If we're waiting until later then let's at least have a JIRA.

I want to test this through integration tests. I'll create a Jira for it so we don't forget.

PR #568, opened 2017-02-03
Garbage Collector Implementation

larioj:


cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/StagedPackageStorage.scala
```diff
@@ -3,18 +3,36 @@ package com.mesosphere.cosmos.storage
 import com.mesosphere.cosmos.http.MediaType
 import com.twitter.util.Future
 import java.io.InputStream
+import java.time.Instant
 import java.util.UUID
 
 final class StagedPackageStorage private(objectStorage: ObjectStorage) {
 
-  def get(packageId: UUID): Future[(MediaType, InputStream)] = {
+  import ObjectStorageOps.objectStorageOps
```
cruhland: The only import that **must** go in the body of a class is for its own companion object. Any other import should be written with a fully-qualified package and placed at the top of the file.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/StagedPackageStorage.scala
```diff
@@ -3,18 +3,36 @@ package com.mesosphere.cosmos.storage
...
+      objectList.directories.map(UUID.fromString)
+    }
+  }
+
+  def delete(id: UUID): Future[Unit] = {
+    objectStorage.delete(id.toString)
+  }
+
+  def getCreationTime(id: UUID): Future[Instant] = {
+    objectStorage.getCreationTime(id.toString).map(_.get)
```
cruhland: Let's not write any new code that uses `Option.get` -- we can just keep the `Option` around and let the caller deal with it. We should have a test to make sure the `None` and `Some` cases are handled correctly.

There's already a JIRA issue open for the other usage of `_.get()` above; it would be really nice if you fixed that in this PR too.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -5,112 +5,73 @@ import com.mesosphere.cosmos.circe.Decoders.decode
...
-import java.io.ByteArrayInputStream
 import java.nio.charset.StandardCharsets
 import scala.util.Try
 
 final class PackageStorage private(objectStorage: ObjectStorage) {
-  private[this] val pool = FuturePool.interruptibleUnboundedPool
 
-  def writePackageDefinition(
+  import ObjectStorageOps.objectStorageOps
+  import com.mesosphere.universe.v3.syntax.PackageDefinitionOps.packageDefinitionToPackageDefinitionOps
```
cruhland: Move these to the top-level imports; only the companion object should be imported here.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -5,112 +5,73 @@ import com.mesosphere.cosmos.circe.Decoders.decode
...
+
+  private[this] def getMetadataName(packageCoordinate: PackageCoordinate): String = {
+    s"${packageCoordinate.as[String]}/metadata.json"
+  }
```
cruhland: This method can be moved to the companion object, since it doesn't need the object instance.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/GarbageCollector.scala
```diff
@@ -0,0 +1,68 @@
...
+        case true => Some(id)
+        case false => None
+      }
+    }
+    Future.collect(garbage.toSeq).map(_.flatten.toSet)
+  }
+
+  private[this] def isGarbage(id: UUID): Future[Boolean] = {
+    stagedPackageStorage.getCreationTime(id).map { creationTime =>
+      val cutoffTime = Instant.now().minusSeconds(timeout)
```
cruhland: Instead of calculating the cutoff time for each package, we should calculate it once per GC "session", for consistency.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
```diff
@@ -48,9 +49,10 @@ final class SyncFutureLeader private (
          */
         Future.join(
           Future.sleep(Duration.fromSeconds(10)), // scalastyle:ignore magic.number
-          processor()
+          processor(),
+          garbageCollector()
```
cruhland: Do we need to run this every ten seconds, if the expiration threshold is one hour?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
```diff
@@ -48,9 +49,10 @@ final class SyncFutureLeader private (
...
         ).unit.onFailure { error =>
-          logger.error(s"Got a failure from the processor", error)
+          logger.error(s"Got a failure from the processor or garbageCollector", error)
```
cruhland: Let's be specific about where the error is coming from. It's a little more code, but will be worth it if we ever need to debug this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/GarbageCollector.scala
```diff
@@ -0,0 +1,68 @@
...
+import com.twitter.util.Future
+import java.time.Instant
+import java.util.UUID
+
+final class GarbageCollector private(
+  stagedPackageStorage: StagedPackageStorage,
+  installQueueReader: ReaderView
+) extends (() => Future[Unit]){
+
+  private[this] val timeout: Long = 60*60
```
cruhland: Could this be stored in a strongly-typed way, so that we know it represents time?

Also, could you add a comment explaining the reasoning for one hour being the chosen interval?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/GarbageCollector.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  override def apply(): Future[Unit] = {
+    listGarbage().flatMap { garbage =>
+      Future.join(garbage.map(stagedPackageStorage.delete).toSeq)
+    }
+  }
+
+  private[this] def listGarbage(): Future[Set[UUID]] = {
```
cruhland: I see why you're using `Set` (to compute `staged -- pending`), but we don't really need it to be in any of the return types since the set-specific properties aren't used anywhere else. Maybe we should just use `List` everywhere and call `staged.diff(pending)` below instead?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageItem.scala
```diff
@@ -0,0 +1,11 @@
+package com.mesosphere.cosmos.storage
+
+import com.mesosphere.cosmos.http.MediaType
+import java.time.Instant
+
+case class ObjectStorageItem(
```
cruhland: It looks like this isn't used outside of test code?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,120 @@
+package com.mesosphere.cosmos.storage
+
+import com.mesosphere.cosmos.http.MediaType
+import com.mesosphere.cosmos.storage.ObjectStorage.ListToken
+import com.twitter.io.StreamIO
+import com.twitter.util.Future
+import com.twitter.util.FuturePool
+import java.io.ByteArrayInputStream
+
+class ObjectStorageOps(objectStorage: ObjectStorage) {
```
cruhland: Some of the methods of this class are only used in tests. Factor out an `ObjectStorageTestOps` class that contains those methods, and put it back in `src/test`?

cruhland: The methods that remain here should have tests written for them.

cruhland: Also, to avoid runtime overhead, this should be declared as:

```scala
final class ObjectStorageOps(val objectStorage: ObjectStorage) extends AnyVal
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,120 @@
...
+        case None =>
+          Future.value(pagesSoFar)
+      }
+    }
+  }
+
+}
+
+object ObjectStorageOps {
+  import scala.language.implicitConversions
```
cruhland: Keep this in the top-level import block.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  implicit def objectStorageOps(objectStorage: ObjectStorage): ObjectStorageOps = {
```
cruhland: You might want to move this to the package object for `com.mesosphere.cosmos.storage` so it can be imported via wildcard (`_`).

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -5,112 +5,73 @@ import com.mesosphere.cosmos.circe.Decoders.decode
...
+    objectStorage.write(metadataName, data, mediaType)
   }
 
-  def readPackageDefinition(
+  def read(
     packageCoordinate: rpc.v1.model.PackageCoordinate
   ): Future[Option[universe.v3.model.V3Package]] = {
-    val path = s"${packageCoordinate.as[String]}/metadata.json"
+    val metadataName = getMetadataName(packageCoordinate)
+    objectStorage.readAsArray(metadataName).map( _.map { case (_, data) =>
```
cruhland: Formatting nitpick: even though it's more verbose, I think always using braces when passing multi-line closures as arguments is easier to read:

```scala
objectStorage.readAsArray(metadataName).map { pendingRead =>
  pendingRead.map { case (_, data) =>
    decode[universe.v3.model.V3Package](new String(data))
  }
}
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -5,112 +5,73 @@ import com.mesosphere.cosmos.circe.Decoders.decode
...
-              )
-          }
-        }
-      } map(_.toList)
-    }
+  def readAllInstalledPackages(): Future[List[universe.v3.model.V3Package]] = {
+    readAllLocalPackages().map(_.flatMap {
+      case rpc.v1.model.Installed(v3Package) => Some(v3Package)
+      case _ => None
+    })
```
cruhland: Try using `List.collect()` to simplify this.

cruhland:
@larioj left you a few comments. Looks good overall!

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/GarbageCollectorSpec.scala
```diff
@@ -0,0 +1,112 @@
...
+  val genInstant: Gen[Instant] = {
+    Gen.choose(0L, Long.MaxValue).map(Instant.ofEpochMilli)
+  }
+
+  def genInstall(v3Package: Option[V3Package] = None): Gen[Install] = {
+    for {
+      generatedV3Package <- Generators.genV3Package
+      uuid <- Gen.uuid
+    } yield Install(uuid, v3Package.getOrElse(generatedV3Package))
+  }
```
cruhland: To avoid generating data when it won't be used, you can pick the package generator based on the state of the `Option`:

```scala
def genInstall(v3Package: Option[V3Package] = None): Gen[Install] = {
  for {
    generatedV3Package <- v3Package.fold(Generators.genV3Package)(Gen.const)
    uuid <- Gen.uuid
  } yield Install(uuid, generatedV3Package)
}
```

It might be even better to go one step further:

```scala
def genInstall(genV3Package: Gen[V3Package] = Generators.genV3Package): Gen[Install] = {
  for {
    v3Package <- genV3Package
    uuid <- Gen.uuid
  } yield Install(uuid, v3Package)
}
```

At call sites where you already have `v3Package: V3Package`, you can then do `genInstall(Gen.const(v3Package))`.

cruhland: Looks like the same technique can be applied to several methods below, too.

cruhland: And actually, just remembered: because `Gen.const` is `implicit`, then in the situation where you want to use a single instance of `v3Package: V3Package`, you can just call `genInstall(v3Package)` and `Gen.const` will be applied automatically.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/GarbageCollector.scala
```diff
@@ -2,67 +2,69 @@ package com.mesosphere.cosmos.storage
...
+      }
+      Future.collect(packages).map(_.toList)
+    }
+    val operations = installQueueReader.viewStatus().map(_.values.toList)
+    val cutoffTime = Instant.now().minusSeconds(timeout)
+    val garbage = for {
+      stagedPackages <- stagedPackages
+      operations <- operations
+    } yield resolveGarbage(stagedPackages, operations, cutoffTime)
+    garbage.flatMap { garbage =>
```
cruhland: Using `flatMap` is the same as `<-` in the for-comprehension; why not combine them?

```scala
for {
  stagedPackages <- stagedPackages
  operations <- operations
  garbage <- resolveGarbage(stagedPackages, operations, cutoffTime)
  _ <- Future.join(garbage.map(stagedPackageStorage.delete).toSeq)
} yield ()
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/StagedPackageStorage.scala
larioj: Okay, I'll do that here.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
larioj: The expiration threshold is variable. It won't be ten seconds, but it could be :)

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
larioj: Package objects are a bit cumbersome. I would not be able to do the same for ObjectStorageTestOps.

cruhland: Haha yeah, I actually ran into that problem in my own code today :) This should be fine.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -70,7 +70,7 @@ final class PackageAddHandler(
   ): Future[universe.v3.model.V3Package] = {
     for {
       // TODO package-add: verify media type
-      (_, packageInputStream) <- stagedPackageStorage.get(stagedPackageId)
+      Some((_, packageInputStream)) <- stagedPackageStorage.read(stagedPackageId)
```
cruhland: If the result of the read is `None`, the `Future` will fail with a generic exception -- probably not what we want. Then again, since we just wrote the package into storage, this shouldn't fail. Maybe the best way to make that obvious is to pass the `Option` up to the caller again, in this case the `apply` method:

```scala
for {
  stagedPackageId <- stagedPackageStorage.put(packageStream, packageSize, contentType)
  Some(v3Package) <- readPackageDefinitionFromStorage(stagedPackageId)
} yield {
  Install(stagedPackageId, v3Package)
}
```

What do you think?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
```diff
@@ -48,10 +49,13 @@ final class SyncFutureLeader private (
...
           Future.sleep(Duration.fromSeconds(10)), // scalastyle:ignore magic.number
-          processor()
-        ).unit.onFailure { error =>
-          logger.error(s"Got a failure from the processor", error)
-        }
+          processor().onFailure {
+            logger.error(s"Got a failure from the processor or garbageCollector", _)
+          },
+          garbageCollector().onFailure {
+            logger.error(s"Got a failure from the processor or garbageCollector", _)
```
cruhland: Update the error messages to be specific :)

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
```diff
@@ -48,10 +49,13 @@ final class SyncFutureLeader private (
...
+          }
+        ).unit
```
cruhland: This `.unit` is unnecessary since you're using `Future.join()`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/GarbageCollector.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+}
+
+object GarbageCollector {
+
+  def apply(
+    stagedPackageStorage: StagedPackageStorage,
+    installQueueReader: ReaderView,
+    timeout: FiniteDuration = 1.hour // TODO check that this is a good default
```
cruhland: Let's get rid of this TODO. An hour should be more than fine, I'd even be comfortable with making this run once a day. Maybe we shouldn't hide this value in a default argument though; always passing it explicitly will remind us that it's configurable. What do you think @jsancio?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/GarbageCollector.scala
```diff
@@ -0,0 +1,70 @@
...
+    val possibleGarbage = staged -- pending
+    possibleGarbage
+      .filter(id => creationTimes(id).isBefore(cutoffTime))
+  }
+
+  def getStagedPackageUuids(operations: List[OperationStatus]): List[UUID] = {
+    operations.flatMap {
+      case PendingStatus(Install(stagedPackageID, _), _) => Some(stagedPackageID)
+      case _ => None
+    }
```
cruhland: You might be interested in using the `collect` method here (check out its Scaladoc):

```scala
operations.collect {
  case PendingStatus(Install(stagedPackageID, _), _) => stagedPackageID
}
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,88 @@
...
+
+final class ObjectStorageOps(val objectStorage: ObjectStorage) extends AnyVal {
+
+  import ObjectStorageOps._
+
+  def readAsArray(name: String): Future[Option[(MediaType, Array[Byte])]] = {
+    objectStorage.read(name).flatMap { item =>
+      FuturePool.interruptibleUnboundedPool { item.map { case (mediaType, readStream) =>
+        (mediaType, StreamIO.buffer(readStream).toByteArray)
+      }}.ensure(item.foreach(_._2.close()))
```
cruhland: Use a named variable instead of `_2` to make this easier to understand at a glance?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,62 @@
...
+    collectPagesReversed(firstPage.map(List(_))).map(_.reverse)
+  }
+
+
+  private[this] def collectPagesReversed
+  (
+    pages: Future[List[ObjectStorage.ObjectList]]
+  ): Future[List[ObjectStorage.ObjectList]] = {
+    pages.flatMap { pagesSoFar =>
+      val mostRecentPage = pagesSoFar.head
```
cruhland: Relying on there being at least one element in `pagesSoFar` is a bit dangerous. What if you separated the input from the output a bit, and passed the next token as a parameter as well?

```scala
private[this] def collectPagesReversed(
  nextToken: Option[ListToken],
  pages: List[ObjectStorage.ObjectList]
): Future[List[ObjectStorage.ObjectList]] = {
  nextToken match {
    case Some(token) =>
      val nextPage = objectStorage.listNext(token)
      nextPage.flatMap(page => collectPagesReversed(page.listToken, page :: pages))
    case _ =>
      Future.value(pages)
  }
}
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,88 @@
...
+    firstPage: Future[ObjectStorage.ObjectList]
+  ): Future[List[ObjectStorage.ObjectList]] = {
+    collectPagesReversed(firstPage.map(List(_))).map(_.reverse)
+  }
+
+  private[this] def collectPagesReversed(
+    pages: Future[List[ObjectStorage.ObjectList]]
+  ): Future[List[ObjectStorage.ObjectList]] = {
+    pages.flatMap { pagesSoFar =>
+      val mostRecentPage = pagesSoFar.head
```
cruhland: Relying on there being at least one element in `pagesSoFar` is a bit dangerous. What if you separated the input from the output a bit, and passed the next token as a parameter as well?

```scala
private[this] def collectPagesReversed(
  nextToken: Option[ListToken],
  pages: List[ObjectStorage.ObjectList]
): Future[List[ObjectStorage.ObjectList]] = {
  nextToken match {
    case Some(token) =>
      val nextPage = objectStorage.listNext(token)
      nextPage.flatMap(page => collectPagesReversed(page.listToken, page :: pages))
    case _ =>
      Future.value(pages)
  }
}
```

cruhland:
@larioj just a few more comments to address, then I'll approve! Looks good!

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
larioj: Haha I was sure I changed this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
larioj: The join returns a triple.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
larioj: Doing it it this way duplicates information and complicates the types. It'd be better to move the method inside collectPages, that way we know that the precondition of nonEmptyness is satisfied. I'll also add a comment.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import com.mesosphere.cosmos.http.MediaType
+import com.mesosphere.cosmos.storage.ObjectStorage.ListToken
+import com.twitter.io.StreamIO
+import com.twitter.util.Future
+import com.twitter.util.FuturePool
+import java.io.ByteArrayInputStream
+import scala.language.implicitConversions
+
+final class ObjectStorageOps(val objectStorage: ObjectStorage) extends AnyVal {
```
jsancio: Make this a companion object of `ObjectStorage`. If you do that then you don't need to import this implicitly. The Scala compiler will look in the companion object during implicit resolution.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -4,117 +4,82 @@ import com.mesosphere.cosmos.PackageFileMissing
...
-        case Some(token) =>
-          objectStorage.listNext(token).flatMap { objectList =>
-            val newPackageCoordinates = objectList.directories.flatMap { directory =>
-              directory.as[Try[rpc.v1.model.PackageCoordinate]].toOption
-            }
+  def list(): Future[List[PackageCoordinate]] = {
+    objectStorage.listWithoutPaging("").map { objectList =>
+      objectList.directories.flatMap(_.as[Try[PackageCoordinate]].toOption)
+    }
+  }
```
jsancio: This method should not be public. Given the semantic of the object stores the values returned is misleading. Why do you need this?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -4,117 +4,82 @@ import com.mesosphere.cosmos.PackageFileMissing
...
-          }
-        }
-      } map(_.toList)
+  def readAllInstalledPackages(): Future[List[universe.v3.model.V3Package]] = {
+    readAllLocalPackages().map { allLocalPackages =>
+      allLocalPackages.collect {
+        case rpc.v1.model.Installed(v3Package) => v3Package
+      }
     }
   }
```
jsancio: Why do you need this at this in this component? We should aim for our interfaces to be as small as possible.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -4,117 +4,82 @@ import com.mesosphere.cosmos.PackageFileMissing
...
 import java.nio.charset.StandardCharsets
 import scala.util.Try
 
 final class PackageStorage private(objectStorage: ObjectStorage) {
-  private[this] val pool = FuturePool.interruptibleUnboundedPool
 
-  def writePackageDefinition(
+  import PackageStorage._
+
+  def write(
```
jsancio: Let's avoid this kind of changes. Why are we renaming the method? There is reason why we call this `writePackageDefinition`. In the future we are going to add something like `writePackageResourse(packageCoordinate: ..., content: InputStream, ...): Future[...]`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -4,117 +4,82 @@ import com.mesosphere.cosmos.PackageFileMissing
...
-      new ByteArrayInputStream(data),
-      data.length.toLong,
-      universe.MediaTypes.universeV3Package
-    )
+    val mediaType = universe.MediaTypes.universeV3Package
+    objectStorage.write(metadataName, data, mediaType)
   }
 
-  def readPackageDefinition(
+  def read(
```
jsancio: Let's avoid this kind of changes. Why are we renaming the method? There is reason why we call this `readPackageDefinition`. In the future we are going to add something like `readPackageResourse(packageCoordinate: ..., id: String, ...): Future[InputStream]`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
```diff
@@ -48,10 +49,13 @@ final class SyncFutureLeader private (
...
-        ).unit.onFailure { error =>
-          logger.error(s"Got a failure from the processor", error)
-        }
+          processor().onFailure {
+            logger.error(s"Got a failure from the processor", _)
+          },
+          garbageCollector().onFailure {
+            logger.error(s"Got a failure from the garbageCollector", _)
+          }
+        ).unit
```
jsancio: This should already by a unit.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  import ObjectStorageOps._
+
+  def readAsArray(name: String): Future[Option[(MediaType, Array[Byte])]] = {
+    objectStorage.read(name).flatMap { item =>
+      FuturePool.interruptibleUnboundedPool { item.map { case (mediaType, readStream) =>
+        (mediaType, StreamIO.buffer(readStream).toByteArray)
+      }}.ensure(item.foreach { case (_ , inputStream) => inputStream.close() })
+    }
+  }
```
jsancio: Do we use this in the Cosmos server or is this for test? If this is for test please move it to a class in the test configuration.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  def write(
+    name: String,
+    content: Array[Byte],
+    mediaType: MediaType
+  ): Future[Unit] = {
+    write(name, content, Some(mediaType))
+  }
```
jsancio: Do we use this in the Cosmos server or is this for test? If this is for test please move it to a class in the test configuration.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  def write(
+    name: String,
+    content: Array[Byte],
+    mediaType: Option[MediaType] = None
+  ): Future[Unit] = {
+    val body = new ByteArrayInputStream(content)
+    val contentLength = content.length.toLong
+    objectStorage.write(name, body, contentLength, mediaType)
+  }
```
jsancio: Do we use this in the Cosmos server or is this for test? If this is for test please move it to a class in the test configuration.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  def listAllObjectNames(root: String): Future[List[String]] = {
+    objectStorage.listWithoutPaging(root).flatMap { objectList =>
+      Future.collect(
+        objectList.directories.map(listAllObjectNames)
+      ).map { rest =>
+        (objectList.objects :: rest.toList).flatten
+      }
+    }
+  }
```
jsancio: If this is for testing please move it to the test configuration.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  def listWithoutPaging(root: String): Future[ObjectStorage.ObjectList] = {
```
jsancio: Let's make this a different type that explicitly state that this is not paged and instead it is strictly evaluated. E.g. this can return `Future[ObjectStorage.ObjectTree]` where `case class ObjectTree(directories: List[String], objects: List[String])`. When we start supporting thousands of packages we can make sure that we remove this class from Cosmos.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
jsancio: To avoid noun repetition you can call it `writeDefinition`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
jsancio: To avoid noun repetition you can call it `readDefinition`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
jsancio: Grepping for this symbol it looks like it is never used.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/StagedPackageStorage.scala
```diff
@@ -1,20 +1,36 @@
...
+import java.time.Instant
 import java.util.UUID
 
 final class StagedPackageStorage private(objectStorage: ObjectStorage) {
 
-  def get(packageId: UUID): Future[(MediaType, InputStream)] = {
-    // TODO package-add: Test for failure cases
-    objectStorage.read(packageId.toString).map(_.get)
+  def read(id: UUID): Future[Option[(MediaType, InputStream)]] = {
+    objectStorage.read(id.toString)
```
jsancio: We should avoid this type of changes. If we are/were unhappy with the method name, it should have come up in the original review.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
larioj: It does not.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
larioj: Isn't that what overloading is for? I thought it was a little redundant to call it writePackageDefinition when the method signatures makes it evident that it takes a PackageDefinition.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
larioj: The explicit import makes the implicit a little less implicit. Any future developer can easily tell where all these random methods are coming from.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
larioj: Okay.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/SyncFutureLeader.scala
larioj: Join returns a tuple: `def join[A,B,C](a: Future[A],b: Future[B],c: Future[C]): Future[(A,B,C)]`

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
larioj: Removed

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
larioj: We use it in both.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
larioj: We use it in cosmos server.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
larioj: It is used in the write up above. The write up above is then used in cosmos server.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
larioj: Moved

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -65,19 +66,28 @@ final class PackageAddHandler(
     }
   }
 
+  private[this] def invert[A](a: Option[Try[A]]): Try[Option[A]] = {
+    // function to change types
+    import cats.implicits._ // Breaks universe conversions
```
cruhland: In the past we've tried to be specific with these imports, to avoid breakage. I think you'll need the following:

* `cats.instances.option._`
* `cats.instances.try_._`
* `cats.syntax.traverse._`

Let's also rename the method to `sequence` for consistency.

cruhland: Oh, and those cats imports should be placed at the top of the file.

larioj: Thanks, I'll give those a shot.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -65,19 +66,28 @@ final class PackageAddHandler(
...
+    Try.fromScala(a.map(twitterTryToScalaTry).sequenceU)
```
cruhland: You might just want to write this method yourself, instead of converting to Scala's `Try` and back. Just pattern match on the `Option`, and then the inner `Try`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
larioj: Actually, I'd rather name it readPackageDefinition.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -65,19 +66,28 @@ final class PackageAddHandler(
...
+  ): Future[Option[universe.v3.model.V3Package]] = {
     for {
       // TODO package-add: verify media type
-      Some((_, packageInputStream)) <- stagedPackageStorage.read(stagedPackageId)
-      packageZip = new ZipInputStream(packageInputStream)
-      packageMetadata <- Future.const(extractPackageMetadata(packageZip))
+      stagedPackage <- stagedPackageStorage.read(stagedPackageId)
+      packageInputStream = stagedPackage.map{ case (_, inputStream) => inputStream}
+      packageZip = packageInputStream.map(new ZipInputStream(_))
+      packageMetadata <- Future.const(invert(packageZip.map(extractPackageMetadata)))
```
cruhland: Ah, this code is more awkward now because `stagedPackage` is an `Option`. Maybe some of it could be pulled into `extractPackageMetadata`, changing its signature into `((MediaType, InputStream)) => Future[Metadata]`? Then here you could just have:

```scala
for {
  stagedPackage <- stagedPackageStorage.read(stagedPackageId)
  packageMetadata <- traverse(stagedPackage)(extractPackageMetadata)
} yield {
  // ...
}
```

Where `traverse` is inspired from Cats and has the type `Option[A] => (A => Future[B])) => Future[Option[B]]`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
larioj: They did not work.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorageOps.scala
jsancio: I am not sure I follow: http://stackoverflow.com/a/5598107/1604407

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
jsancio: We should avoid using overloading if the semantic is different.

jsancio:
@larioj Let me know if there are any other comments you would like me to reply. Thanks!

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
larioj: Is the semantic different in this case? The only difference is what we are writing.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/StagedPackageStorage.scala
larioj: The review process isn't always perfect. Should we allow things to remain inconsistent once they have become so?

jsancio: Or create another PR that does just this change.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
jsancio: I think so in the future the writing call will:

```scala
store.writeResource(...)
store.writeResource(...)
...
store.writePacakgeDefinition(...)
```

You can think of `writePackageDefinition` almost as the commit statement. From the readers point of view a package is only valid after the `writePackageDefinition` call is successful. I wanted to make it parent in the signature.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/StagedPackageStorage.scala
larioj: That's fair. I'll keep that in mind for the future. Can we marge this as is? The types of this method are already being changed, so I don't see it being a huge stretch.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
larioj: You appear to feel strongly about this so I'll make the change. For completeness though:

```scala
for {
_ <- store.writeResource(...)
_ <- store.writeResource(...)
...
packageDefinition <- extractPackageDefintion(someZip)
_ <- store.writePackageDefinition(packageDefinition)
} yield ...
```

PR #569, opened 2017-02-06
Make client response size configurable

jsancio:


cosmos-server/src/it/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
```diff
@@ -51,9 +52,9 @@ object CosmosIntegrationTestClient extends Matchers {
       }
       .flatMap { case (adminRouterUri, marathonUri, mesosMasterUri) =>
         Trys.join(
-          Services.adminRouterClient(adminRouterUri).map { adminRouterUri -> _},
-          Services.marathonClient(marathonUri).map { marathonUri -> _ },
-          Services.mesosClient(mesosMasterUri).map { mesosMasterUri -> _ }
+          Services.adminRouterClient(adminRouterUri, 5.megabytes).map { adminRouterUri -> _},
+          Services.marathonClient(marathonUri, 5.megabytes).map { marathonUri -> _ },
+          Services.mesosClient(mesosMasterUri, 5.megabytes).map { mesosMasterUri -> _ }
```
cruhland: Do we want these new arguments to have default values, so we don't need to update the code here?

jsancio: In general, I am not a big fan default of argument in larger code base. If I change the signature of a method I would like to know of all the places where it is used so I can fix them explicitly.

cruhland: Ok, makes sense!

jsancio:
Thanks for the review @cruhland!

PR #570, opened 2017-02-07
Update scala compiler flags

jsancio:
The bigger change is that we are enabling checking for unused imports

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -165,31 +165,27 @@ object CosmosBuild extends Build {
       "-Xlint:deprecation"
     ),
 
-    scalacOptions ++= Seq(
-      "-deprecation", // Emit warning and location for usages of deprecated APIs.
-      "-encoding", "UTF-8",
-      "-explaintypes",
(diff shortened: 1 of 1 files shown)


project/build.scala
```diff
@@ -165,31 +165,27 @@ object CosmosBuild extends Build {
...
-      "-Ywarn-nullary-unit",
-      "-Ywarn-numeric-widen",
-      "-Ywarn-unused",
-      "-Ywarn-unused-import",
-      "-Ywarn-value-discard"
-    ),
-
-    scalacOptions in (Compile, console) ~= (_ filterNot (_ == "-Ywarn-unused-import")),
-
-    scalacOptions in (Test, console) ~= (_ filterNot (_ == "-Ywarn-unused-import")),
```
cruhland: We want to keep these filters in place. If we don't, any time you try to import something in the REPL, an unused import error message will be printed to the screen. It's super annoying! See this issue, for example: https://issues.scala-lang.org/browse/SI-7722

jsancio: :+1:  Done.

PR #571, opened 2017-02-12
Verify universe only add v3 package

jsancio:


cosmos-http/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
```diff
@@ -25,6 +24,7 @@ object MediaTypes {
   val ListVersionsRequest = vnd("list-versions-request")
   val ListVersionsResponse = vnd("list-versions-response")
   val CapabilitiesResponse = dcos("capabilities", 1)
+  val ErrorResponse = dcos("error", 1)
```
jsancio: I found a small bug. Need to verify this is okay.

larioj:
What is the motivation for writing qualified types like `rpc.v1.model.PackageRepository` instead of `PackageRepository`?

jsancio:
> What is the motivation for writing qualified types like rpc.v1.model.PackageRepository instead of PackageRepository?

@larioj Because we have so many types with the same simple name we have decided to always import them as `{package}.{version}.model.{simpleName}. This is a decision that we made after some of the code was written. As we make changes through the code and have been updating those inconsistency.

PR #572, opened 2017-02-14
Strong types for path manipulation

cruhland:
Fixes https://jira.mesosphere.com/browse/DCOS-13634.

Here's what I have so far. I didn't have time to use the new types everywhere, but I did update `ObjectStorage` as a trial run. See PR comments for more details. Hopefully this is good enough to use for now, and we can make incremental improvements over time.


cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/LocalObjectStorage.scala
```diff
@@ -251,9 +254,17 @@ object LocalObjectStorage {
     new LocalObjectStorage(storageDir, scratchDir, stats)
   }
 
+  def resolve(filesystemPath: Path, objectPath: AbsolutePath): Path = {
+    filesystemPath.resolve(objectPath.toString.drop(1))
+  }
+
+  def relativize(basePath: Path, fullPath: Path): AbsolutePath = {
+    AbsolutePath(s"/${basePath.relativize(fullPath)}").right.get
+  }
```
cruhland: I created the above two helper methods to translate between filesystem paths and our abstract paths. In `S3ObjectStorage` we have a similar pair of methods; maybe they should be part of our `Path` implementation?

I designed the "path append" operation to only allow relative paths to be appended. But in the case of `resolve()` we actually want to treat the `objectPath` as a relative path *with respect to* the base filesystem path. It's somewhat of a unique operation, so I'm still thinking we don't want to support appends of absolute paths.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorage.scala
```diff
@@ -47,19 +48,19 @@ trait ObjectStorage {
...
    */
-  def delete(name: String): Future[Unit]
+  def delete(path: AbsolutePath): Future[Unit]
 
   /**
    * List some objects and directories in the given path. If there are too many objects or
    * directories, the result is truncated and the listToken property is set to Some value.
    */
-  def list(directory: String): Future[ObjectStorage.ObjectList]
+  def list(path: AbsolutePath): Future[ObjectStorage.ObjectList]
```
cruhland: This operation is the reason `AbsolutePath`s are used in this trait instead of `RelativePath`s: because we need to be able to list the "top-level" path. An alternative would be allowing empty `RelativePath`s, but I thought that would be confusing.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -31,7 +36,7 @@ final class PackageStorage private(objectStorage: ObjectStorage) {
     val data = packageDefinition.asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
 
     objectStorage.write(
-      s"${packageCoordinate.as[String]}/metadata.json",
+      AbsolutePath(s"/${packageCoordinate.as[String]}/$MetadataJson").right.get,
```
cruhland: I construct an entire path string before parsing it with `AbsolutePath.apply`. This is because "path append" requires a `RelativePath` argument, so we'd be forced to parse each component separately:

```scala
AbsolutePath.Root / RelativePath(packageCoordinate.as[String]).right.get / RelativePath(MetadataJson).right.get
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/S3ObjectStorage.scala
```diff
@@ -146,16 +147,18 @@ final class S3ObjectStorage(
...
+        dir => relativePath(AbsolutePath(dir).right.get)
       )(breakOut),
       listToken
     )
   }
 
-  private[this] def fullPath(name: String): String = {
-    makeStringDirectory(path) + name
+  private[this] def fullPath(path: AbsolutePath): String = {
+    (basePath / RelativePath(path.toString).right.get).toString
```
cruhland: This method, and `relativePath` below, are the equivalents of the `resolve` and `relativize` methods in `LocalObjectStorage`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -347,10 +350,10 @@ final class InstallQueue private(
 
 object InstallQueue {
 
-  val installQueuePath = "/package/task-queue"
+  val installQueuePath: AbsolutePath = abspath"/package/task-queue"
```
cruhland: String interpolation macro! This will catch an invalid path at compile time.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
```diff
@@ -347,10 +350,10 @@ final class InstallQueue private(
...
 
-  def statusPath(packageCoordinate: PackageCoordinate): String = {
-    s"$installQueuePath/${packageCoordinate.as[String]}"
+  def statusPath(packageCoordinate: PackageCoordinate): AbsolutePath = {
+    installQueuePath / RelativePath(packageCoordinate.as[String]).right.get
```
cruhland: The `RelativePath(...).right.get` idiom ends up being used a lot, because most of the time we know from context that the argument will be a valid path. It would be convenient if the `/` method (append operation) could take a `String` as an argument and return a path that's not wrapped in an `Either`, but then we'd have to either (1) throw exceptions for invalid paths, or (2) allow invalid paths to be created and hope that it doesn't happen in practice. I'd lean towards option (1) since option (2) means our abstract paths are just as "unsafe" as `String` paths.

cruhland: Just outlining my thought process above; my preferred choice is the current approach, which very explicitly calls out when we are certain that path construction will always succeed.

cosmos-server/src/main/scala/org/apache/curator/framework/api/package.scala
```diff
@@ -0,0 +1,16 @@
...
+package object api {
+
+  implicit final class PathableOps[A](val underlying: Pathable[A]) extends AnyVal {
+    def forPath(path: AbsolutePath): A = underlying.forPath(path.toString)
+  }
+
+  implicit final class PathAndBytesableOps[A](val underlying: PathAndBytesable[A]) extends AnyVal {
+    def forPath(path: AbsolutePath): A = underlying.forPath(path.toString)
+    def forPath(path: AbsolutePath, data: Array[Byte]): A = underlying.forPath(path.toString, data)
+  }
```
cruhland: These classes allow us to use our paths directly with the Curator API.

cosmos-server/src/test/scala/com/mesosphere/cosmos/storage/LocalObjectStorageSpec.scala
```diff
@@ -197,40 +200,39 @@ final class LocalObjectStorageSpec extends FreeSpec with PropertyChecks {
...
       .flatMap(Gen.buildableOfN[String, Char](_, Gen.alphaNumChar))
 
-    (for {
+    for {
       numSegments <- Gen.choose(1, maxSegments)
       segments <- Gen.listOfN(numSegments, genSegment)
-      path = segments.mkString("/")
-    } yield path).suchThat(isValidPath)
+    } yield {
+      RelativePath(segments.mkString("/")).right.get
```
cruhland: Oops, I need to filter values based on the result of `RelativePath.apply`, instead of asserting they will always succeed. Will update.

cosmos-universe/src/main/scala/com/mesosphere/util/package.scala
```diff
@@ -0,0 +1,36 @@
...
+
+    def abspathMacro(c: Context)(args: c.Expr[Any]*): c.Expr[AbsolutePath] = {
+      import c.universe._
+
+      // Examine the code of the object that abspath() was called on
+      // Should have the form: PathInterpolations(StringContext(<string literal>))
+      c.prefix.tree match {
+        case Apply(_, List(Apply(_, List(Literal(Constant(rawPath: String)))))) =>
+          AbsolutePath(rawPath) match {
+            case Right(_) => c.Expr[AbsolutePath](q"AbsolutePath($rawPath).right.get")
```
cruhland: Unfortunately we can't use a macro to call a private constructor: the code generated by macros is spliced back into the source file and recompiled, so it is treated like ordinary code in every way. That means we're forced to execute the validation code at runtime, but at least we have a static guarantee that it will work.

project/build.scala
```diff
@@ -104,6 +104,10 @@ object CosmosBuild extends Build {
       "org.scalacheck" %% "scalacheck" % V.scalaCheck % "test"
     )
 
+    val scalaReflect = Seq(
+      "org.scala-lang" % "scala-reflect" % projectScalaVersion
```
cruhland: Needed for macros.

project/build.scala
```diff
@@ -279,7 +283,16 @@ object CosmosBuild extends Build {
 
   lazy val cosmos = Project("cosmos", file("."))
     .settings(sharedSettings)
-    .aggregate(http, model, json, jsonschema, bijection, render, finch, server)
+    .aggregate(http, model, json, jsonschema, bijection, render, finch, server, universe)
+
+  lazy val universe = Project("cosmos-universe", file("cosmos-universe"))
```
cruhland: Here's the module that will be used by universe-service.

scalastyle-config.xml
```diff
@@ -72,8 +72,8 @@
  <check level="error" class="org.scalastyle.scalariform.SimplifyBooleanExpressionChecker" enabled="true"></check>
  <check level="error" class="org.scalastyle.scalariform.IfBraceChecker" enabled="true">
   <parameters>
-   <parameter name="singleLineAllowed"><![CDATA[true]]></parameter>
-   <parameter name="doubleLineAllowed"><![CDATA[false]]></parameter>
```
cruhland: Disabling warnings for two-line if statements:

```scala
if (condition) ...
else ...
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -83,16 +86,11 @@ final class PackageStorage private(objectStorage: ObjectStorage) {
       }
     }
 
-    objectStorage.list("").flatMap { objectList =>
+    objectStorage.list(AbsolutePath.Root).flatMap { objectList =>
```
jsancio: :+1:

project/build.scala
jsancio: The name doesn't feel right. This is more like a common set of functionality and it doesn't have anything to do with Universe. How about just `cosmos-common` for now?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
jsancio: I think this is a side effect of the API not matching the usage. It doesn't look like we handle the `Left` in the code. If that is the case then why not just throw in the `def /` method and simplify the call site.

cosmos-universe/src/main/scala/com/mesosphere/util/Path.scala
```diff
@@ -0,0 +1,78 @@
...
+  override def toString: String = elements.mkString(Path.Separator)
+
+}
+
+object RelativePath {
+
+  def apply(path: String): Either[Error, RelativePath] = {
+    if (path.isEmpty) Left(Empty)
+    else if (path.startsWith(Path.Separator)) Left(Absolute)
+    else Right(RelativePath(path.split(Path.Separator).toList.filter(_.nonEmpty)))
```
larioj: Does the library handle relative paths of the form `./foo/bar`?

jsancio:
Looks good in general. Left you a couple of comments.

larioj:
It seems to me like this library is quite an endeavor. It reminds me of an XKCD: https://xkcd.com/927/

Which is a little troubling because we might just end up with yet another path implementation littered throughout the codebase.

I think that paths have a rigorous foundation, so we could in theory end up with something useful. But I think we need to decide whether we want to keep on invading Russia and brave the winter, or cut our losses.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
cruhland: Yeah, thanks for calling this out. In almost all situations a `Left` would indicate a programming error and would not be recoverable. So an exception makes more sense.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/installqueue/InstallQueue.scala
cruhland: Based on your observation @jsancio, I'll change this to go with my option (1) from above.

cosmos-universe/src/main/scala/com/mesosphere/util/Path.scala
cruhland: @larioj no, and I don't plan to support them, or parent paths (`..`) either. They are usually only seen in filesystems, and the intent here is to be more generic than that.

cruhland:
@larioj I don't really view this abstract `Path` as competing with the other, more concrete path types that we use. Its main purpose is to ensure that we never use `String` to represent paths in our own code. We can convert our `Path` to `String` at the boundaries, if a library requires `String` paths. But the code we write should use the path types so that our intent is clearer, and there's less of a chance to create invalid paths.

cruhland:
@jsancio @larioj after getting rid of `Either` in the common case I decided to clean up a few other places where the interface was a bit awkward. I've gotten it to a place I'm happy with, but it required some non-trivial changes so I'd recommend taking another look. I'm done making big changes in this PR, I promise!

cosmos-common/src/main/scala/com/mesosphere/util/Path.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+trait Path {
+
+  /** The concrete type of this path. */
+  type Self <: Path
+
+  /**
+   * The path obtained by appending the given element to the end of this path.
+   *
+   * @throws IllegalArgumentException if the given element is empty or contains [[Path.Separator]].
```
cruhland: Oops, I allow empty elements now. I'll update the comment.

cosmos-common/src/main/scala/com/mesosphere/util/RelativePath.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  override def resolve(tail: RelativePath): RelativePath = RelativePath(elements ++ tail.elements)
+
+  override def toString: String = elements.mkString(Path.Separator)
+
+}
+
+object RelativePath {
+
+  val Empty: RelativePath = RelativePath(Vector.empty)
```
cruhland: I decided to allow empty relative paths because they make it much easier to use `relativize`: relativizing a path with itself will produce the empty path instead of throwing an exception.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/ObjectStorage.scala
cruhland: Technically I could change the trait to use `RelativePath` now, but I don't think it makes enough difference to bother with in this PR.

cosmos-common/src/main/scala/com/mesosphere/util/Path.scala
cruhland: Actually, I changed my mind about this. Empty relative *paths* are allowed, but I think it's still useful to prevent adding empty *elements*, because they could be the result of a programmer mistake. If you really want a no-op append you can use `resolve` to append an empty relative path.

jsancio:
@cruhland Looking.

cosmos-common/src/main/scala/com/mesosphere/util/Path.scala
```diff
@@ -0,0 +1,31 @@
+package com.mesosphere.util
+
+/**
+ * A location in a treelike structure where the nodes are named by strings.
+ *
+ * Useful for safely manipulating paths obtained from various storage systems such as the local
+ * filesystem, ZooKeeper, and S3, as well as more abstract sources like URIs.
+ */
+trait Path {
```
jsancio: Do you want this to be a sealed trait?

cruhland: I thought about this, and decided that it's actually not that useful for it to be `sealed`. In most cases I think we will either explicitly have a `RelativePath` or an `AbsolutePath`. And in the cases where we don't know, it's likely we'll only be interested in the methods on the base `Path` trait anyway.

In the future I could see having other implementations as well, for example we might have one that wraps a Java `Path` for efficiency, or uses a `String` representation to avoid the overhead of splitting and recombining the path elements.

PR #573, opened 2017-02-15
Log HTTP requests sent to Universe

jsancio:


cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -69,14 +71,25 @@ final class DefaultUniverseClient(
...
   private[this] val fetchScope = stats.scope("fetch")
 
   private[this] val cosmosVersion = BuildProperties().cosmosVersion
 
   def apply(repository: PackageRepository)(implicit session: RequestSession): Future[universe.v3.model.Repository] = {
     adminRouter.getDcosVersion().flatMap { dcosVersion =>
-      apply(repository, dcosVersion.version)
+      apply(repository, dcosVersion.version).respond {
+        case Return(_) =>
+          logger.info(
```
cruhland: What log level do we run at in prod? Do we need logs in the success case? I'm worried this will clutter the logs.

jsancio: This is our configuration: https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/resources/logback.xml#L31-L33

jsancio: I can change it to debug. We log every request to Cosmos. We should log the line less than or equal to the number of request against Cosmos.

cruhland: I guess when you put it that way, it's not that bad. Let's keep it at info level.

PR #575, opened 2017-02-17
Use scalacheck-shapeless to define generators

cruhland:
This change introduces the `Arbitrary` instances required for scalacheck-shapeless to generate `Arbitrary[universe.v3.model.Metadata]`. It also removes some explicit `Gen` definitions in favor of automatic generation.


jsancio:
LGTM.

PR #577, opened 2017-02-24
Utility methods for reading and writing DC/OS packages

cruhland:
This finishes the implementation of [DCOS-13635](https://jira.mesosphere.com/browse/DCOS-13635).

### DIFF
--- a/cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
@@ -3,18 +3,34 @@ package com.mesosphere.util
 import com.mesosphere.universe
 import java.io.InputStream
 import java.io.OutputStream
-import scala.util.Failure
+import java.util.zip.ZipInputStream
 import scala.util.Try
 
 object PackageUtil {
 
+  val MetadataPath: RelativePath = RelativePath("metadata.json")
+
   def extractMetadata(packageContent: InputStream): Try[universe.v3.model.Metadata] = {
-    Failure(EntryNotFound(AbsolutePath.Root / "metadata.json"))
+    Try {
+      val zipIn = new ZipInputStream(packageContent)
+
+      val hasMetadata = Iterator.continually(Option(zipIn.getNextEntry))
+        .takeWhile(_.isDefined)
+        .flatten
+        .exists(entry => RelativePath(entry.getName) == MetadataPath)
+
+      throw if (hasMetadata) InvalidEntry(MetadataPath) else MissingEntry(MetadataPath)
+    }
   }
 
   def buildPackage(packageContent: OutputStream, metadata: universe.v3.model.Metadata): Unit = ???
 
   sealed abstract class PackageError(override val getMessage: String) extends RuntimeException
-  case class EntryNotFound(path: AbsolutePath) extends PackageError(s"Missing package entry: $path")
+
+  case class MissingEntry(path: RelativePath)
+    extends PackageError(s"Package entry not found: $path")
+
+  case class InvalidEntry(path: RelativePath)
+    extends PackageError(s"Package entry was not JSON or did not match the expected schema: $path")
 
 }
--- a/cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
@@ -5,6 +5,7 @@ import com.mesosphere.universe
 import com.mesosphere.universe.v3.circe.Decoders._
 import com.mesosphere.universe.v3.circe.Encoders._
 import com.twitter.io.StreamIO
+import io.circe.JsonObject
 import io.circe.jawn.decode
 import io.circe.syntax._
 import java.io.InputStream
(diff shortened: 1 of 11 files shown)


cosmos-model/src/test/scala/com/mesosphere/Generators.scala
```diff
@@ -130,4 +142,22 @@ object Generators {
 
   }
 
+  /** The members of this trait cannot be implicit, to avoid resolving to themselves. */
```
cruhland: @jsancio I looked into the `Lazy` type in Shapeless, and it didn't seem to be useful for what I need here. Open to suggestions for how to improve this.

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  def buildPackage(packageContent: OutputStream, metadata: universe.v3.model.Metadata): Unit = {
+    val zipOut = new ZipOutputStream(packageContent)
+    zipOut.putNextEntry(new ZipEntry(MetadataPath.toString))
+    val metadataBytes = metadata.asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
+    zipOut.write(metadataBytes)
+    zipOut.close()
+  }
+
+  sealed abstract class PackageError(override val getMessage: String) extends RuntimeException
```
jsancio: Okay for now. We may want to consider moving CosmosError to this common module. We can do that when we schedule refactoring the CosmosError class.

cosmos-model/src/test/scala/com/mesosphere/Generators.scala
jsancio: @cruhland I am trying to understand why we need two `trait`s and one `object`. Would this work if we just one `object` and define `arbMetadataShapeless` and `arbLocalPackageShapeless` private and first?

cosmos-model/src/test/scala/com/mesosphere/Generators.scala
```diff
@@ -130,4 +142,22 @@ object Generators {
...
+  trait ShapelessImplicits extends CustomImplicits {
+
+    protected val arbMetadataShapeless: Arbitrary[universe.v3.model.Metadata] = {
+      implicitly[Arbitrary[universe.v3.model.Metadata]]
+    }
+
+    protected val arbLocalPackageShapeless: Arbitrary[rpc.v1.model.LocalPackage] = {
+      implicitly[Arbitrary[rpc.v1.model.LocalPackage]]
```
jsancio: I think you can resolve this by finding out which `implicit def/val` is being used here and calling it directly.

cosmos-model/src/test/scala/com/mesosphere/Generators.scala
cruhland: We probably don't need the separate traits, I just organized it that way so I could get things working quickly. I'll see what I can do to make it work in one object.

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
```diff
@@ -0,0 +1,63 @@
...
+import java.io.OutputStream
+import java.nio.charset.StandardCharsets
+import java.util.zip.ZipEntry
+import java.util.zip.ZipInputStream
+import java.util.zip.ZipOutputStream
+import scala.util.Try
+
+object PackageUtil {
+
+  val MetadataPath: RelativePath = RelativePath("metadata.json")
```
cruhland: Done.

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  def extractMetadata(packageContent: InputStream): Try[universe.v3.model.Metadata] = {
+    val zipIn = new ZipInputStream(packageContent)
+
+    val result = Try {
+      val foundMetadata = Iterator.continually(Option(zipIn.getNextEntry))
+        .takeWhile(_.isDefined)
+        .flatten
+        .exists(entry => RelativePath(entry.getName) == MetadataPath)
```
cruhland: Done.

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
cruhland: Oops, I actually meant to handle these errors in `PackageAddHandler` by wrapping them in `CosmosError`. I'll do that now.

But yeah, I didn't use `CosmosError` here because it's too coupled with `cosmos-server`. While we could move it to the common module, I'm wondering if it would actually be more modular to have specific `sealed trait` errors for each subsystem, and then `CosmosError` could wrap each trait in a unique, higher-level `case class`. What do you think?

cosmos-model/src/test/scala/com/mesosphere/Generators.scala
cruhland: I'll try that. Thanks!

cruhland:
@jsancio @larioj updated!

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
```diff
@@ -0,0 +1,74 @@
...
+    val zipIn = new ZipInputStream(packageContent)
+
+    try {
+      Iterator.continually(Option(zipIn.getNextEntry))
+        .takeWhile(_.isDefined)
+        .flatten
+        .find(entry => RelativePath(entry.getName) == MetadataPath)
+        .toRight[PackageError](MissingEntry(MetadataPath))
+        .flatMap { _ =>
+          val metadataBytes = StreamIO.buffer(zipIn).toByteArray
```
larioj: Isn't this blocking?

cosmos-model/src/test/scala/com/mesosphere/Generators.scala
```diff
@@ -94,7 +100,14 @@ object Generators {
     genTagString.map(universe.v3.model.PackageDefinition.Tag(_).get)
   }
 
-  private val genPathPart: Gen[PathPart] = Gen.resultOf(PathPart.apply _)
+  private val genUri: Gen[Uri] = {
+    arbitrary[String]
+      .map(s => Try(Uri.parse(s)))
+      .flatMap {
+        case Success(uri) => uri
+        case _ => Gen.fail
```
larioj: Will this not lead to too many generation failures? If we discard too many results the tests will fail.

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
larioj: I see that you wrap calls to this method in a future pool. Should we be doing something to label methods as blocking?

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
```diff
@@ -0,0 +1,74 @@
...
+
+          decode[universe.v3.model.Metadata](metadataString)
+            .leftMap(error => InvalidEntry(MetadataPath, error))
+        }
+    } finally {
+      zipIn.close()
+    }
+  }
+
+  def buildPackage(packageContent: OutputStream, metadata: universe.v3.model.Metadata): Unit = {
```
larioj: Are we using this outside of the tests?

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
cruhland: In this case, whether or not the method blocks depends on the `InputStream` argument. If it's a `ByteArrayInputStream`, which is what the unit tests use, then the method doesn't block. Since the caller hopefully knows more about what's being passed in, it can decide how to handle the possibility of blocking.

cosmos-model/src/test/scala/com/mesosphere/Generators.scala
cruhland: In general yes, this is an anti-pattern. But in this case it turns out that almost all strings are parseable as URIs. I'll add a comment.

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
cruhland: Caught me! We aren't using it right now, but eventually universe-service will need to use it. It's also, in a sense, the opposite operation to `extractMetadata` so it makes sense to keep them together.

larioj: 👍

cosmos-common/src/main/scala/com/mesosphere/util/PackageUtil.scala
larioj: Okay

PR #581, opened 2017-03-04
Update SBT version to 0.13.13

cruhland:


project/build.scala
```diff
@@ -274,8 +273,8 @@ object CosmosBuild extends Build {
   private lazy val cosmosIntegrationTestServer = settingKey[CosmosIntegrationTestServer]("cosmos-it-server")
 
   val itSettings = Defaults.itSettings ++ Seq(
-    test in IntegrationTest <<= (test in IntegrationTest).dependsOn(oneJar),
-    testOnly in IntegrationTest <<= (testOnly in IntegrationTest).dependsOn(oneJar),
+    (test in IntegrationTest).set((test in IntegrationTest).dependsOn(oneJar), NoPosition),
+    (testOnly in IntegrationTest).set((testOnly in IntegrationTest).dependsOn(oneJar), NoPosition),
```
cruhland: This PR took way longer than expected, and here's the reason. I couldn't get these settings to work correctly using the standard `:=` operator. So I've compromised by replacing `<<=` with a rough approximation of its definition (it's a macro, so I can't get file position information). This is slightly better than continuing to use `<<=` since it avoid deprecation warnings.

I've filed an issue for the follow up work: [DCOS_OSS-708](https://jira.mesosphere.com/browse/DCOS_OSS-708).

build.sbt
```diff
@@ -0,0 +1,123 @@
+import com.mesosphere.cosmos.CosmosBuild._
+
+lazy val cosmos = project.in(file("."))
+  .settings(sharedSettings)
+  .aggregate(http, model, json, common, jsonschema, bijection, render, finch, server)
+
+lazy val common = project.in(file("cosmos-common"))
```
cruhland: `project.in()` is now the recommended syntax. It obtains the project name from the variable (`common` in this case). Since we want the name to be `cosmos-common`, I set it below using the directory name.

project/build.scala
```diff
@@ -1,12 +1,11 @@
 package com.mesosphere.cosmos
 
 import com.github.retronym.SbtOneJar._
-import scoverage.ScoverageKeys._
-import sbtfilter.Plugin._
 import sbt.Keys._
 import sbt._
+import scoverage.ScoverageKeys._
 
-object CosmosBuild extends Build {
```
cruhland: The `Build` class is deprecated.

PR #582, opened 2017-03-06
Ensure kafka service gets cleaned up after test

cruhland:
When running the integration tests locally, I've noticed that they would leave a Kafka instance in the cluster after they finished. This would result in some tests failing on the next run. It's been happening for at least a few weeks, and it got annoying enough that I decided to just fix it.

### DIFF
--- a/cosmos-server/src/it/scala/com/mesosphere/cosmos/ServiceStartSpec.scala
+++ b/cosmos-server/src/it/scala/com/mesosphere/cosmos/ServiceStartSpec.scala
@@ -28,7 +28,7 @@ final class ServiceStartSpec extends FreeSpec with InstallQueueFixture with Befo
     val uninstallCassandra =
       rpc.v1.model.UninstallRequest("cassandra", appId = None, all = Some(true))
     val uninstallLinkerd =
-      rpc.v1.model.UninstallRequest("linkerd", appId = None, all = Some(true))
+      rpc.v1.model.UninstallRequest("kafka", appId = None, all = Some(true))
     val uninstallHelloworld =
       rpc.v1.model.UninstallRequest("helloworld", appId = None, all = Some(true))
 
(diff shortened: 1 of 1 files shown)


cosmos-server/src/it/scala/com/mesosphere/cosmos/ServiceStartSpec.scala
```diff
@@ -28,7 +28,7 @@ final class ServiceStartSpec extends FreeSpec with InstallQueueFixture with Befo
     val uninstallCassandra =
       rpc.v1.model.UninstallRequest("cassandra", appId = None, all = Some(true))
     val uninstallLinkerd =
-      rpc.v1.model.UninstallRequest("linkerd", appId = None, all = Some(true))
+      rpc.v1.model.UninstallRequest("kafka", appId = None, all = Some(true))
```
jsancio: 1. Help me understand this PR. What is this fixing?
1. The variable name `uinstallLinkerd` doesn't make the request the uninstalls `kafka`.

cruhland: Oh, good catch with the variable name. I'll update that.

I've updated the PR description with the reason for this PR.

cruhland: Ah, now I get why this might have been confusing. None of the tests in this file install linkerd, but one of them does install Kafka (without uninstalling it). I'm guessing that test used to install linkerd, but was changed at some point, and the cleanup code wasn't kept in sync.

larioj: Yep, that exactly what happened. I changed the test to use kafka, but I never changed the cleanup :(

PR #586, opened 2017-03-15
Add supported package definition type

larioj:
We are adding a new type to cosmos. This type will represent the supported packaging formats. The  idea is that we will use this type in contexts where we only want to support certain packages.

### DIFF
--- a/cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
+++ b/cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
@@ -1,6 +1,8 @@
 package com.mesosphere.universe.bijection
 
 import com.mesosphere.universe
+import com.mesosphere.universe.v3.model.V3Metadata
+import com.mesosphere.universe.v3.model.V3Package
 import com.netaporter.uri.Uri
 import com.twitter.bijection.Bijection
 import com.twitter.bijection.Conversion
@@ -97,7 +99,7 @@ object UniverseConversions {
       universe.v2.model.ReleaseVersion]
   }
 
-  implicit val v3MetadataToV3Package:
+  private implicit val v3MetadataToV3Package:
     Conversion[(universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion),
       universe.v3.model.V3Package] = Conversion.fromFunction { case (metadata, releaseVersion) =>
     universe.v3.model.V3Package(
@@ -122,6 +124,15 @@ object UniverseConversions {
       config=metadata.config,
       command=None // Command are not supported in Package.dcos. Setting to the default value.
     )
+  }
+
+  implicit val MetadataToSupportedPackageDefinition:
+    Conversion[(universe.v3.model.Metadata, universe.v3.model.ReleaseVersion),
+      universe.v3.model.SupportedPackageDefinition] = Conversion.fromFunction { case (metadata, releaseVersion) =>
+      metadata match {
+        case v3Metadata: V3Metadata =>
+          (v3Metadata, releaseVersion).as[V3Package]
+      }
   }
 
   implicit val v3PackageDefinitionToV2PackageDetails:
--- a/cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
+++ b/cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
@@ -1,8 +1,6 @@
 package com.mesosphere.universe.bijection
 
 import com.mesosphere.universe
-import com.mesosphere.universe.v3.model.V3Metadata
-import com.mesosphere.universe.v3.model.V3Package
 import com.netaporter.uri.Uri
 import com.twitter.bijection.Bijection
 import com.twitter.bijection.Conversion
@@ -126,12 +124,12 @@ object UniverseConversions {
     )
   }
 
-  implicit val MetadataToSupportedPackageDefinition:
+  implicit val metadataToSupportedPackageDefinition:
     Conversion[(universe.v3.model.Metadata, universe.v3.model.ReleaseVersion),
       universe.v3.model.SupportedPackageDefinition] = Conversion.fromFunction { case (metadata, releaseVersion) =>
       metadata match {
-        case v3Metadata: V3Metadata =>
-          (v3Metadata, releaseVersion).as[V3Package]
+        case v3Metadata: universe.v3.model.V3Metadata =>
+          (v3Metadata, releaseVersion).as[universe.v3.model.V3Package]
       }
   }
 
--- a/cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
+++ b/cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
@@ -97,7 +97,7 @@ object UniverseConversions {
       universe.v2.model.ReleaseVersion]
   }
 
-  private implicit val v3MetadataToV3Package:
+  implicit val v3MetadataToV3Package:
     Conversion[(universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion),
       universe.v3.model.V3Package] = Conversion.fromFunction { case (metadata, releaseVersion) =>
     universe.v3.model.V3Package(
--- a/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
+++ b/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
@@ -2,10 +2,12 @@ package com.mesosphere.universe.bijection
 
 import com.mesosphere.universe
 import com.twitter.bijection.Conversion
+import com.twitter.bijection.Conversion.asMethod
+
 
 object TestUniverseConversions {
 
-  implicit val v3PackageToV3Metadata:
+  private implicit val v3PackageToV3Metadata:
     Conversion[universe.v3.model.V3Package,
       (universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion)] = {
     Conversion.fromFunction { v3Package =>
@@ -38,27 +40,7 @@ object TestUniverseConversions {
       (universe.v3.model.Metadata, universe.v3.model.ReleaseVersion)] = {
     Conversion.fromFunction {
       case v3Package: universe.v3.model.V3Package =>
-        val metadata = universe.v3.model.V3Metadata(
-          packagingVersion = v3Package.packagingVersion,
-          name = v3Package.name,
-          version = v3Package.version,
-          maintainer = v3Package.maintainer,
-          description = v3Package.description,
-          tags = v3Package.tags,
-          scm = v3Package.scm,
-          website = v3Package.website,
-          framework = v3Package.framework,
-          preInstallNotes = v3Package.preInstallNotes,
-          postInstallNotes = v3Package.postInstallNotes,
-          postUninstallNotes = v3Package.postUninstallNotes,
-          licenses = v3Package.licenses,
-          minDcosReleaseVersion = v3Package.minDcosReleaseVersion,
-          marathon = v3Package.marathon,
-          resource = v3Package.resource,
-          config = v3Package.config
-        )
-
-        (metadata, v3Package.releaseVersion)
+        v3Package.as[(universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion)]
     }
   }
 
--- a/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
+++ b/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
@@ -6,7 +6,7 @@ import com.twitter.bijection.Conversion.asMethod
 
 object TestUniverseConversions {
 
-  private implicit val v3PackageToV3Metadata:
+  implicit val v3PackageToV3Metadata:
     Conversion[universe.v3.model.V3Package,
       (universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion)] = {
     Conversion.fromFunction { v3Package =>
--- a/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
+++ b/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
@@ -33,4 +33,33 @@ object TestUniverseConversions {
     }
   }
 
+  implicit val supportedPackageToV3Metadata:
+    Conversion[universe.v3.model.SupportedPackageDefinition,
+      (universe.v3.model.Metadata, universe.v3.model.ReleaseVersion)] = {
+    Conversion.fromFunction {
+      case v3Package: universe.v3.model.V3Package =>
+        val metadata = universe.v3.model.Metadata(
+          packagingVersion = v3Package.packagingVersion,
+          name = v3Package.name,
+          version = v3Package.version,
+          maintainer = v3Package.maintainer,
+          description = v3Package.description,
+          tags = v3Package.tags,
+          scm = v3Package.scm,
+          website = v3Package.website,
+          framework = v3Package.framework,
+          preInstallNotes = v3Package.preInstallNotes,
+          postInstallNotes = v3Package.postInstallNotes,
+          postUninstallNotes = v3Package.postUninstallNotes,
+          licenses = v3Package.licenses,
+          minDcosReleaseVersion = v3Package.minDcosReleaseVersion,
+          marathon = v3Package.marathon,
+          resource = v3Package.resource,
+          config = v3Package.config
+        )
+
+        (metadata, v3Package.releaseVersion)
+    }
+  }
+
 }
--- a/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
+++ b/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
@@ -4,7 +4,6 @@ import com.mesosphere.universe
 import com.twitter.bijection.Conversion
 import com.twitter.bijection.Conversion.asMethod
 
-
 object TestUniverseConversions {
 
   private implicit val v3PackageToV3Metadata:
--- a/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
+++ b/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
@@ -33,12 +33,12 @@ object TestUniverseConversions {
     }
   }
 
-  implicit val supportedPackageToV3Metadata:
+  implicit val supportedPackageToMetadata:
     Conversion[universe.v3.model.SupportedPackageDefinition,
       (universe.v3.model.Metadata, universe.v3.model.ReleaseVersion)] = {
     Conversion.fromFunction {
       case v3Package: universe.v3.model.V3Package =>
-        val metadata = universe.v3.model.Metadata(
+        val metadata = universe.v3.model.V3Metadata(
           packagingVersion = v3Package.packagingVersion,
           name = v3Package.name,
           version = v3Package.version,
--- a/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
+++ b/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
@@ -42,12 +42,12 @@ final class UniverseConversionsSpec extends FreeSpec with Matchers {
     }
 
     def testCase(
-      value: (universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion)
+      value: (universe.v3.model.Metadata, universe.v3.model.ReleaseVersion)
     ): Assertion = {
       assertResult(value) {
         value
-          .as[universe.v3.model.V3Package]
-          .as[(universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion)]
+          .as[universe.v3.model.SupportedPackageDefinition]
+          .as[(universe.v3.model.Metadata, universe.v3.model.ReleaseVersion)]
       }
     }
 
@@ -68,8 +68,10 @@ final class UniverseConversionsSpec extends FreeSpec with Matchers {
       val command = v3Package.command
 
       val roundTrip = v3Package
-        .as[(universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion)]
-        .as[universe.v3.model.V3Package]
+        .asInstanceOf[universe.v3.model.SupportedPackageDefinition]
+        .as[(universe.v3.model.Metadata, universe.v3.model.ReleaseVersion)]
+        .as[universe.v3.model.SupportedPackageDefinition]
+        .asInstanceOf[universe.v3.model.V3Package]
 
       assert(roundTrip.selected.isEmpty)
       assert(roundTrip.command.isEmpty)
--- a/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
+++ b/cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
@@ -68,10 +68,8 @@ final class UniverseConversionsSpec extends FreeSpec with Matchers {
       val command = v3Package.command
 
       val roundTrip = v3Package
-        .asInstanceOf[universe.v3.model.SupportedPackageDefinition]
-        .as[(universe.v3.model.Metadata, universe.v3.model.ReleaseVersion)]
-        .as[universe.v3.model.SupportedPackageDefinition]
-        .asInstanceOf[universe.v3.model.V3Package]
+        .as[(universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion)]
+        .as[universe.v3.model.V3Package]
 
       assert(roundTrip.selected.isEmpty)
       assert(roundTrip.command.isEmpty)
--- a/cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala
+++ b/cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala
@@ -129,4 +129,15 @@ object Decoders {
     }
   }
 
+  implicit val decodeSupportedPackageDefinition: Decoder[SupportedPackageDefinition] = {
+    Decoder.instance[SupportedPackageDefinition] { (hc: HCursor) =>
+      hc.downField("packagingVersion").as[PackagingVersion].flatMap {
+        case V3PackagingVersion => hc.as[V3Package]
+        case V2PackagingVersion => Left(DecodingFailure(
+          s"V2Package is not a supported package definition",
+          hc.history
+        ))
+      }
+    }
+  }
 }
--- a/cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/AddResponse.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/AddResponse.scala
@@ -4,4 +4,4 @@ import com.mesosphere.universe
 
 // This class is only a wrapper for the underlying package
 // We need a unique type so that the media type encoder can be resolved implicitly
-final class AddResponse(val supportedPackage: universe.v3.model.SupportedPackageDefinition) extends AnyVal
+final class AddResponse(val v3Package: universe.v3.model.V3Package) extends AnyVal
--- a/cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/AddResponse.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/AddResponse.scala
@@ -4,4 +4,4 @@ import com.mesosphere.universe
 
 // This class is only a wrapper for the underlying package
 // We need a unique type so that the media type encoder can be resolved implicitly
-final class AddResponse(val v3Package: universe.v3.model.V3Package) extends AnyVal
+final class AddResponse(val supportedPackage: universe.v3.model.SupportedPackageDefinition) extends AnyVal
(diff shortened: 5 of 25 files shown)


cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
```diff
@@ -128,11 +129,100 @@ case class V3Package(
...
+) extends PackageDefinition with Ordered[V3Package] with Spd {
   override def compare(that: V3Package): Int = {
     PackageDefinition.compare(
       (name, version, releaseVersion),
       (that.name, that.version, that.releaseVersion)
     )
   }
 }
+
+sealed trait Spd
```
cruhland: Please rename this to something understandable to someone unfamiliar with the intent of the code.

Also, you might want to use a "self type" to restrict what this can be mixed into:

```scala
sealed trait Spd { self: PackageDefinition => }
```

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
```diff
@@ -128,11 +129,100 @@ case class V3Package(
...
+
+object Spd {
+  implicit class SpdOps(spd: Spd) {
```
cruhland: Rename this too.

cruhland: For full efficiency, this should really be:

```scala
implicit final class SpdOps(val spd: Spd) extends AnyVal {
```

Extending `AnyVal` means that there won't be an instance of `SpdOps` at runtime -- all members are treated as having Java's `static` modifier, essentially.

cruhland: Is there a base trait that defines some of these ops already for `PackageDefinition`? I think you might be able to mix that trait into this class, to ensure we define all the expected methods.

cruhland:
@larioj left a few comments.

cosmos-http/src/main/scala/com/mesosphere/universe/MediaTypes.scala
```diff
@@ -31,4 +31,10 @@ object MediaTypes {
     Map("version" -> "v1")
   )
 
+  val supportedPackageDefinition = MediaType(
+    "application",
+    MediaTypeSubType("vnd.dcos.universe.supported-package", Some("json")), // TODO: @larioj check this
```
cruhland: This shouldn't exist; the package and/or repo objects we send and receive should always be of a concrete type. Unless I'm misunderstanding what this is for?

larioj: This is for storage. The ObjectStorage API requires a media type. But since this will be S3, then yes, I should replace this.

larioj: At the moment we don't use the media type for anything.

cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala
```diff
@@ -118,4 +118,15 @@ object Decoders {
...
+    Decoder.instance[SupportedPackageDefinition] { (hc: HCursor) =>
+      hc.downField("packagingVersion").as[PackagingVersion].flatMap {
+        case V3PackagingVersion => hc.as[V3Package]
+        case V2PackagingVersion => Left(DecodingFailure(
+          s"V2Package is not a supported package definition",
+          hc.history
+        ))
+      }
+    }
+  }
```
jsancio: Where do we use this decoder?

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
```diff
@@ -128,11 +128,28 @@ case class V3Package(
...
   override def compare(that: V3Package): Int = {
     PackageDefinition.compare(
       (name, version, releaseVersion),
       (that.name, that.version, that.releaseVersion)
     )
   }
 }
+
+sealed trait SupportedPackageDefinition extends PackageDefinition {
+  self: PackageDefinition =>
```
jsancio: Do you need `self: PackageDefinition =>` given that this trait extends `PackageDefinition`?

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
```diff
@@ -128,11 +128,28 @@ case class V3Package(
   resource: Option[V3Resource] = None,
   config: Option[JsonObject] = None,
   command: Option[Command] = None
-) extends PackageDefinition with Ordered[V3Package] {
+) extends PackageDefinition with Ordered[V3Package] with SupportedPackageDefinition {
```
jsancio: Change this to `extends SupportedPackageDefinition with Ordered[V3Package]` since `S..P..D` extends `PackageDefinition`.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
```diff
@@ -134,7 +134,9 @@ final class PackageDefinitionOps(val pkgDef: PackageDefinition) extends AnyVal {
 object PackageDefinitionOps {
   import scala.language.implicitConversions
 
-  implicit def packageDefinitionToPackageDefinitionOps(pkgDef: PackageDefinition): PackageDefinitionOps = {
+  implicit def packageDefinitionToPackageDefinitionOps[P <: PackageDefinition](
+    pkgDef: P
+  ): PackageDefinitionOps = {
```
jsancio: Why do we need a type parameter? What error do you get with the old signature?

larioj: The type parameter makes it so that this conversion works for SPD, since it is a subtype of PackageDefinition. Without the type parameter the implicit conversion does not get picked up.

cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Decoders.scala
larioj: We need this decoder for all the other decoders that depend on it. Since we are changing the Operation class to include SPD. The Operation class is used in a bunch of other classes that have decoders.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
larioj: I'll remove it.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
larioj: Done

cosmos-model/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
jsancio: Okay.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/AddResponse.scala
```diff
@@ -4,4 +4,4 @@ import com.mesosphere.universe
 
 // This class is only a wrapper for the underlying package
 // We need a unique type so that the media type encoder can be resolved implicitly
-final class AddResponse(val v3Package: universe.v3.model.V3Package) extends AnyVal
+final class AddResponse(val supportedPackage: universe.v3.model.SupportedPackageDefinition) extends AnyVal
```
cruhland: I'm wondering if this is applying abstraction in the wrong place. This file defines what an `rpc.v1.model.AddResponse` is, which is a `V3Package`. If we want to have an `AddResponse` as a `V4Package`, we should create an `rpc.v2.model.AddResponse`. The decision about which version to send would then be made by an instance of `DispatchingMediaTypedEncoder[rpc.v2.model.AddResponse]`. See [here](https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/rpc/v2/circe/MediaTypedEncoders.scala#L30) for an example.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/PackageDefinition.scala
```diff
@@ -128,11 +128,26 @@ case class V3Package(
...
       (that.name, that.version, that.releaseVersion)
     )
   }
 }
+
+sealed trait SupportedPackageDefinition extends PackageDefinition
+
+object SupportedPackageDefinition {
+
+  implicit val supportedPackageDefinitionOrdering = new Ordering[SupportedPackageDefinition] {
```
cruhland: Why is this needed now?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -62,9 +62,9 @@ final class PackageAddHandler(
 
     for {
       operation <- futureOperation
-      _ <- producerView.add(operation.v3Package.packageCoordinate, operation)
+      _ <- producerView.add(operation.packageDefinition.packageCoordinate, operation)
     } yield {
-      new rpc.v1.model.AddResponse(operation.v3Package)
+      new rpc.v1.model.AddResponse(operation.packageDefinition)
```
cruhland: Ah, this is why you changed the definition of `AddResponse`. We need to be careful about changing the definitions of anything under a `model` package: they are intended to be immutable once defined (except for changes which would not affect clients). Instead we use the version numbering (`v1.model`, `v2.model`, etc.) to account for changes. This ensures that we do not break old clients when releasing a new version of Cosmos.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/AddResponse.scala
jsancio: I think we need to add `universe.v5.model.SupportPackageDefinition` if we add support for v5 packages.

cruhland: Ah, if we're going to have a new version of `SupportedPackageDefinition` each time we add a new package version, then this change should be fine. But doesn't that kind of defeat the point of having the `SupportedPackageDefinition` trait in the first place? Since we'd have to update a bunch of places in the code to use the new version? Or is the idea that we'd keep all the existing code as is, and only new code would use the new version?

cruhland: For example, what about low-level, generic code that doesn't care which version of a package it's handling?

larioj: Does this mean that we would have to duplicate endpoints?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
larioj: Oh yeah, I thought we could break the experimental stuff. Or should I not assume that?

cruhland:
@larioj looks good, I'm just a bit uncertain about some use cases; where should or shouldn't we be using `SupportedPackageDefinition`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
jsancio: You can break the experimental stuff but let's make sure that we update the CLI if it breaks the CLI.

larioj:
# Notes on the latest changes

* Add response contains a v3Package.
* Package storage decodes types based on the media type. This is not compiler checked.
* The commits are actually well organized.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -37,8 +38,13 @@ final class PackageStorage private(objectStorage: ObjectStorage) {
   ): Future[Option[universe.v3.model.SupportedPackageDefinition]] = {
     val metadataName = getMetadataPath(packageCoordinate)
     objectStorage.readAsArray(metadataName).map { pendingRead =>
-      pendingRead.map { case (_, data) =>
-        decode[universe.v3.model.SupportedPackageDefinition](new String(data, StandardCharsets.UTF_8))
+      pendingRead.map {
+        case (universe.MediaTypes.universeV3Package, data) =>
```
larioj: This case statement is not checked by the compiler.

cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
```diff
@@ -124,6 +126,15 @@ object UniverseConversions {
     )
   }
 
+  implicit val MetadataToSupportedPackageDefinition:
```
cruhland: Nitpick: this name shouldn't be capitalized

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -37,8 +38,13 @@ final class PackageStorage private(objectStorage: ObjectStorage) {
...
+          decode[universe.v3.model.SupportedPackageDefinition](
```
larioj: maybe I should make this `decode[universe.v3.model.V3Package]`

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -65,7 +65,10 @@ final class PackageAddHandler(
       operation <- futureOperation
       _ <- producerView.add(operation.packageDefinition.packageCoordinate, operation)
     } yield {
-      new rpc.v1.model.AddResponse(operation.packageDefinition)
+      operation.packageDefinition match {
+        case v3Package: universe.v3.model.V3Package =>
```
larioj: When we add a v4 package, we will have to extend this case statement. I'm not sure, what I need to do when this is V4Package.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -22,21 +23,28 @@ final class PackageStorage private(objectStorage: ObjectStorage) {
...
   def readPackageDefinition(
     packageCoordinate: rpc.v1.model.PackageCoordinate
-  ): Future[Option[universe.v3.model.V3Package]] = {
+  ): Future[Option[universe.v3.model.SupportedPackageDefinition]] = {
     val metadataName = getMetadataPath(packageCoordinate)
     objectStorage.readAsArray(metadataName).map { pendingRead =>
-      pendingRead.map { case (_, data) =>
-        decode[universe.v3.model.V3Package](new String(data, StandardCharsets.UTF_8))
+      pendingRead.map {
+        case (universe.MediaTypes.universeV3Package, data) =>
```
larioj: This case statement is not checked by the compiler.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -55,23 +55,26 @@ final class PackageAddHandler(
...
     }
 
     for {
       operation <- futureOperation
-      _ <- producerView.add(operation.v3Package.packageCoordinate, operation)
+      _ <- producerView.add(operation.packageDefinition.packageCoordinate, operation)
     } yield {
-      new rpc.v1.model.AddResponse(operation.v3Package)
+      operation.packageDefinition match {
+        case v3Package: universe.v3.model.V3Package =>
```
larioj: This case statement will eventually include a v4 package. I'm not sure what the match will return.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -22,21 +23,28 @@ final class PackageStorage private(objectStorage: ObjectStorage) {
...
+          decode[universe.v3.model.SupportedPackageDefinition](
```
larioj: maybe I should make this `decode[universe.v3.model.V3Package]`

cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
```diff
@@ -1,6 +1,8 @@
 package com.mesosphere.universe.bijection
 
 import com.mesosphere.universe
+import com.mesosphere.universe.v3.model.V3Metadata
+import com.mesosphere.universe.v3.model.V3Package
```
larioj: Why are these two imported? Should we not be using the fully qualified name?

cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
```diff
@@ -2,10 +2,12 @@ package com.mesosphere.universe.bijection
 
 import com.mesosphere.universe
 import com.twitter.bijection.Conversion
+import com.twitter.bijection.Conversion.asMethod
+
```
larioj: useless space

cosmos-model/src/main/scala/com/mesosphere/universe/PackageDefinition.scala
```diff
@@ -101,4 +101,20 @@ package v3.model {
     }
   }
 
+  sealed trait SupportedPackageDefinition
+    extends PackageDefinition
+
+  object SupportedPackageDefinition {
+
+    implicit val supportedPackageDefinitionOrdering = new Ordering[SupportedPackageDefinition] {
```
cruhland: You should be able to simplify this to:

```scala
implicit val supportedPackageDefinitionOrdering: Ordering[SupportedPackageDefinition] =
  Ordering.by(identity)
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
```diff
@@ -507,4 +507,10 @@ case class OperationInProgress(coordinate: rpc.v1.model.PackageCoordinate) exten
 case class InvalidPackage(reason: PackageUtil.PackageError) extends CosmosError(Some(reason)) {
   override val getData: Option[JsonObject] = Some(reason.getData)
 }
+
+case class UnknownMediaType(msg: String) extends CosmosError {
```
cruhland: I'm not sure that we need a `CosmosError` for this. If we ever encounter a media type that isn't expected, it indicates a bug in Cosmos; there was nothing the client could have done to prevent it. Throwing an `AssertionError` or a `RuntimeException` with an informative message should be enough in case we need to debug.

cosmos-model/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/LocalPackage.scala
```diff
@@ -2,7 +2,7 @@ package com.mesosphere.cosmos.rpc.v1.model
 
 import com.mesosphere.cosmos.storage.v1.model.Operation
 import com.mesosphere.universe
-import com.mesosphere.universe.v3.syntax.PackageDefinitionOps._
+import com.mesosphere.universe.v3.syntax.PackageDefinitionOps.packageDefinitionToPackageDefinitionOps
```
larioj: Change this back

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
```diff
@@ -22,21 +23,28 @@ final class PackageStorage private(objectStorage: ObjectStorage) {
...
+            new String(data, StandardCharsets.UTF_8)
+          )
```
cruhland: I think we discussed in standup changing this to the following?

```scala
case (universe.MediaTypes.universeV3Package, data) =>
  decode[universe.v3.model.V3Package](new String(data, StandardCharsets.UTF_8))
```

We can still return `Future[Option[universe.v3.model.SupportedPackageDefinition]]` since `V3Package` is a subclass.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
cruhland: That's fine, since you have the default case, correct?

cosmos-model/src/main/scala/com/mesosphere/cosmos/storage/v1/model/PendingOperation.scala
```diff
@@ -1,11 +1,11 @@
 package com.mesosphere.cosmos.storage.v1.model
 
 import com.mesosphere.cosmos.rpc
-import com.mesosphere.universe.v3.syntax.PackageDefinitionOps._
+import com.mesosphere.universe.v3.syntax.PackageDefinitionOps.packageDefinitionToPackageDefinitionOps
```
larioj: Change this back.

cosmos-server/src/it/scala/com/mesosphere/cosmos/PackageAddSpec.scala
```diff
@@ -19,6 +19,7 @@ import com.mesosphere.cosmos.test.TestUtil
 import com.mesosphere.universe
 import com.mesosphere.universe.test.TestingPackages
 import com.mesosphere.universe.v3.circe.Decoders._
+import com.mesosphere.universe.v3.model.SupportedPackageDefinition
```
larioj: We should not need this import.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
cruhland: We discussed this previously. This might require changing `EndpointHandler` to accommodate the new use case of `PackageAddHandler`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
cruhland: :+1:

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/package.scala
```diff
@@ -2,33 +2,34 @@ package com.mesosphere.cosmos
 
 import com.mesosphere.cosmos.storage.PackageStorage
 import com.mesosphere.universe
+import com.mesosphere.universe.v3.syntax.PackageDefinitionOps.packageDefinitionToPackageDefinitionOps
```
larioj: Why do we need this import?

cosmos-server/src/main/scala/com/mesosphere/cosmos/search/package.scala
```diff
@@ -1,6 +1,7 @@
 package com.mesosphere.cosmos
 
 import cats.syntax.either._
+import com.mesosphere.universe.v3.syntax.PackageDefinitionOps.packageDefinitionToPackageDefinitionOps
```
larioj: We do not need this import

cosmos-model/src/main/scala/com/mesosphere/universe/PackageDefinition.scala
larioj: I don't understand. What does ordering by identity accomplish.

cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
```diff
@@ -68,8 +68,10 @@ final class UniverseConversionsSpec extends FreeSpec with Matchers {
       val command = v3Package.command
 
       val roundTrip = v3Package
-        .as[(universe.v3.model.V3Metadata, universe.v3.model.ReleaseVersion)]
-        .as[universe.v3.model.V3Package]
+        .asInstanceOf[universe.v3.model.SupportedPackageDefinition]
+        .as[(universe.v3.model.Metadata, universe.v3.model.ReleaseVersion)]
+        .as[universe.v3.model.SupportedPackageDefinition]
+        .asInstanceOf[universe.v3.model.V3Package]
```
jsancio: What are you testing here? Is the description still correct `V3Package => ... => V3Package`?

cosmos-model/src/main/scala/com/mesosphere/universe/PackageDefinition.scala
```diff
@@ -101,4 +101,20 @@ package v3.model {
...
+      override def compare(a: SupportedPackageDefinition, b: SupportedPackageDefinition): Int = {
+        PackageDefinition.compare(
+          (a.name, a.version, a.releaseVersion),
+          (b.name, b.version, b.releaseVersion)
+        )
+      }
+    }
+
+  }
```
jsancio: Let's move `SupportedPackageDefinition` after `PackageDefinition` and before `V2Package`.

cosmos-model/src/main/scala/com/mesosphere/universe/PackageDefinition.scala
cruhland: It's maybe a bit cryptic if you don't know the definition of `Ordering.by`. It's delegating to the implicit `Ordering[PackageDefinition]` defined earlier in the file.

cosmos-bijection/src/main/scala/com/mesosphere/universe/bijection/UniverseConversions.scala
```diff
@@ -97,7 +97,7 @@ object UniverseConversions {
       universe.v2.model.ReleaseVersion]
   }
 
-  implicit val v3MetadataToV3Package:
+  private implicit val v3MetadataToV3Package:
```
jsancio: `private[this]`.

cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/TestUniverseConversions.scala
```diff
@@ -2,10 +2,11 @@ package com.mesosphere.universe.bijection
 
 import com.mesosphere.universe
 import com.twitter.bijection.Conversion
+import com.twitter.bijection.Conversion.asMethod
 
 object TestUniverseConversions {
 
-  implicit val v3PackageToV3Metadata:
+  private implicit val v3PackageToV3Metadata:
```
jsancio: `private[this]`

cosmos-model/src/main/scala/com/mesosphere/universe/PackageDefinition.scala
larioj: Yup quite cryptic. But kinda makes sense.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
jsancio: Yea. Didn't we agree to make this `case class AddResponse(package: SupportedPackageDefinition)` and change the jsonschema for `application/...add-response+json...` the union of all of the supported packages?

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
larioj: Yup

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
```diff
@@ -55,23 +55,26 @@ final class PackageAddHandler(
...
+          new rpc.v1.model.AddResponse(v3Package)
```
jsancio: Should `AddResponse` be a `case class` if so, no need for `new`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/storage/PackageStorage.scala
jsancio: @larioj pattern matching doesn't match with the symbol is lowercase. I think you need to make it `UniverseV3Package` instead of `universeV3Package`.

larioj: Paths count as stable identifiers, and as such can be used as Literal pattern matches.

```scala
scala> import com.mesosphere.universe

scala> universe.MediaTypes.UniverseV3Repository match { case universe.MediaTypes.universeV3Package => true; }
scala.MatchError: MediaType(application,MediaTypeSubType(vnd.dcos.universe.repo,Some(json)),Map(charset -> utf-8, version -> v3)) (of class com.mesosphere.cosmos.http.MediaType)
  ... 43 elided
```

cosmos-bijection/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
larioj: Yes

jsancio: If that is the case then we should remove both `asInstanceOf`. This probably means that you will have to make the implicit converter public. If instead you are trying to test `SPD => ... SPD` then I think you just need to change the signature of this method to `(value: SPD) => ...`

cosmos-model/src/main/scala/com/mesosphere/universe/PackageDefinition.scala
larioj: For some reason this broke the tests.

cruhland: This should work:

```scala
    implicit val supportedPackageDefinitionOrdering: Ordering[SupportedPackageDefinition] =
      PackageDefinition.packageDefinitionOrdering.on(identity)
```

cosmos-model/src/main/scala/com/mesosphere/universe/PackageDefinition.scala
larioj: Done

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageAddHandler.scala
larioj: We'll make that change when we add V4Package

PR #588, opened 2017-03-23
Factor the main() method out of the Cosmos type

cruhland:
- Introduces a `CosmosApp` trait that declares an abstract `main()` method
- Introduces a `CosmosApi` class that defines the handlers and endpoints, given all their dependencies in the constructor
- The `Cosmos` object now extends `CosmosApp`, and implements `main()` to start the app with the endpoints from `CosmosApi`
- Projects which pull in the `cosmos-server` module as a dependency can define their own implementations of `CosmosApp`


cruhland:
@jsancio @larioj I rearranged all the code in `Cosmos.scala` -- it would probably be easier just to look at the new file rather than the diff.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -50,181 +41,70 @@ import com.twitter.server.Lifecycle
...
-
-  val packageSearch: Endpoint[Json] = {
-    route(post("package" :: "search"), packageSearchHandler)(RequestValidators.standard)
-  }
+  import CosmosApp._
 
-  val packageList: Endpoint[Json] = {
-    route(post("package" :: "list"), packageListHandler)(RequestValidators.standard)
-  }
+  def main(): Unit
```
jsancio: I would just remove this. JVM forces this to be static and defined as the entry point for the application.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -375,84 +244,102 @@ with Logging {
...
+    clazz.getName.replaceAllLiterally("$", ".")
   }
 
-  private[this] def enableIfSome[A, Req, Res](requirement: Option[A], operationName: String)(
-    f: A => EndpointHandler[Req, Res]
-  ): EndpointHandler[Req, Res] = {
-    requirement.fold[EndpointHandler[Req, Res]](new NotConfiguredHandler(operationName))(f)
+}
+
+object Cosmos extends CosmosApp {
```
jsancio: What did we decide to do here? Are we going to move this a project called `cosmos-main`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosApi.scala
```diff
@@ -0,0 +1,144 @@
...
+  val objectStorages: Option[(LocalPackageCollection, StagedPackageStorage)],
+  val repositories: MultiRepository,
+  val producerView: ProducerView,
+  val packageRunner: PackageRunner
+) {
+
+  import CosmosApi._
+
+  // Package Handlers
+  val packageInstall: Endpoint[Json] = {
```
jsancio: All of the `val`s except for `allEndpoints` should be `private[this]`, right?

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosApi.scala
```diff
@@ -0,0 +1,144 @@
...
+import com.mesosphere.cosmos.storage.installqueue.ProducerView
+
+final class CosmosApi(
+  val adminRouter: AdminRouter,
+  val sourcesStorage: PackageSourcesStorage,
+  val objectStorages: Option[(LocalPackageCollection, StagedPackageStorage)],
+  val repositories: MultiRepository,
+  val producerView: ProducerView,
+  val packageRunner: PackageRunner
+) {
```
jsancio: This looks like a function with the following type:
```scala
(
 AdminRouter,
 PackageSourcesStorage,
 Option[(LocalPackageCollection, StagedPackageStorage)],
 MultiRepository,
 ProducerView,
 PackageRunner
) => Endpoint[Json]
```
or
```scala
(
 AdminRouter,
 PackageSourcesStorage,
 Option[(LocalPackageCollection, StagedPackageStorage)],
 MultiRepository,
 ProducerView,
 PackageRunner
) => Service[Request, Response]
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -50,181 +41,70 @@ import com.twitter.server.Lifecycle
...
-  }
+  lazy val logger: Logger = org.slf4j.LoggerFactory.getLogger(getClass)
 
-  val packageRepositoryAdd: Endpoint[Json] = {
-    route(
-      post("package" :: "repository" :: "add"),
-      packageRepositoryAddHandler
-    )(RequestValidators.standard)
-  }
+  final def buildApi(): CosmosApi = {
```
jsancio: If you implement my other suggestions, you can change this to `final def buildApi(): Endpoint[Json]`

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: This is actually a subtlety of the `com.twitter.app.App` trait that we are extending. It defines the *actual* `main`, which must take a parameter `args: Array[String]`. It then uses reflection (!) to find a zero-argument `main` method defined in the implementing class, and executes it; see [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L169-L172). It's way too magical for my taste, but not worth messing with for now.

I only discovered that after I accidentally added an argument to `main` and was surprised when Cosmos quit immediately after starting!

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: I decided we didn't need to do that. Might as well keep things simpler. Unless I'm missing something?

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosApi.scala
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  // Capabilities
+  val capabilities: Endpoint[Json] = {
+    val handler = new CapabilitiesHandler
+    route(get("capabilities"), handler)(RequestValidators.noBody)
+  }
+
+  // Keep alphabetized
+  val allEndpoints = (
```
larioj: How come we are not giving this a type?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -50,181 +41,70 @@ import com.twitter.server.Lifecycle
 import com.twitter.server.Stats
```
larioj: It appears that the `import com.twitter.finagle.stats.NullStatsReceiver` above is unnecessary.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -375,84 +244,102 @@ with Logging {
...
-      case (Some(packageStorage), Some(stagedStorage)) =>
-        Some((packageStorage, stagedStorage, LocalPackageCollection(packageStorage, installQueue)))
-      case (None, None) =>
-        None
-      case (Some(_), None) =>
-        throw new IllegalArgumentException(
-          "Missing staged storage configuration. Staged storage configuration required if " +
-          "package storage provided."
+    val api = endpoints.handle {
+      case re: CosmosError =>
```
larioj: What does `re` stand for? Also what are our guidelines for naming? I was under the impression that we never used names like this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -375,84 +244,102 @@ with Logging {
...
-    val repositories = new MultiRepository(
-      sourcesStorage,
-      universeClient
-    )
+object CosmosApp {
 
-    val packageAddHandler = enableIfSome(objectStorages, "package add") {
-      case (_, stagedStorage) => new PackageAddHandler(repositories, stagedStorage, installQueue)
+  private def startServer(service: Service[Request, Response]): Option[ListeningServer] = {
+    getHttpInterface.map { iface =>
```
larioj: Is `iface` a typo?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -375,84 +244,102 @@ with Logging {
...
     }
+  }
 
-    val serviceStartHandler = enableIfSome(objectStorages, "service start") {
-      case (localPackageCollection, _) =>
-        new ServiceStartHandler(localPackageCollection, packageRunner)
+  private def startTlsServer(
+    service: Service[Request, Response]
+  ): Option[ListeningServer] = {
+    getHttpsInterface.map { iface =>
```
larioj: Is `iface` a typo?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -375,84 +244,102 @@ with Logging {
...
-      new PackageRepositoryDeleteHandler(sourcesStorage),
-      new PackageRepositoryListHandler(sourcesStorage),
-      new PackageSearchHandler(repositories),
-      new UninstallHandler(adminRouter, repositories),
-      serviceStartHandler
-    )(statsReceiver)
+  /**
+   * Removes characters from class names that are disallowed by some metrics systems.
+   *
+   * @param clazz the class whose name is to be santised
```
larioj: Typo in `santised` should be `sanitized` (cause 'merica)

cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosApi.scala
cruhland: The type would be `Endpoint[Json :+: Json :+: ... :+: CNil]`, with one `Json` per (actual) endpoint. I'm looking into whether Finch has an operation that will convert the type to `Endpoint[Json]`, but if not, then I'd rather let Scala infer the type than have to update it each time we add an endpoint.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: Removed.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: It's leftover from when we had a `RequestError` type. I can rename it.

As for naming guidelines, "never" is a strong word. It really depends on context. The goal is not to have the shortest code, or code using only fullyExplanatoryCamelCaseIdentifiers, but instead code that is easy to understand when reading it six months after it was written. Of course this is a fuzzy and subjective concept, but that's what code reviews are for!

I don't mind the short names for exceptions here because it's a common idiom, even in Java. There's only one exception variable in each case block, and its type must be explicitly written, so its name is mostly a placeholder.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: This was from existing code. I've updated it to `interface`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: This was from existing code. I've updated it to `interface`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: Fixed.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
larioj: Thanks

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -375,84 +338,133 @@ with Logging {
...
-    f: A => EndpointHandler[Req, Res]
-  ): EndpointHandler[Req, Res] = {
-    requirement.fold[EndpointHandler[Req, Res]](new NotConfiguredHandler(operationName))(f)
+  implicit def degenerateCoproduct[H, T <: Coproduct](implicit
+    toOne: T => H :+: CNil
+  ): (H :+: T => H :+: CNil) = {
+    case Inl(h) => Inl(h)
+    case Inr(t) => toOne(t)
   }
+
```
cruhland: @jsancio @larioj One last change; I figured out how to simplify the `Endpoint` type.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -42,189 +58,147 @@ import com.twitter.finagle.http.Response
...
-        :+: packageRender
-        :+: packageRepositoryAdd
-        :+: packageRepositoryDelete
-        :+: packageRepositoryList
-        :+: packageSearch
-        :+: packageUninstall
-        :+: serviceStart
-      )
+  // scalastyle:off method.length
+  protected final def buildEndpoints(components: Components): Endpoint[Json :+: CNil] = {
```
jsancio: Is there a way to make the return type `Json` instead of `Json :+: CNil`? Isn't this the same as `Json` vs `Some[Json]` (notice the `Some` type vs `Option` type). Meaning this type represent a "list" that always has one element.

cruhland: You're right, according to the docs for `CNil` it is impossible to have an instance of it. I'll update.

PR #589, opened 2017-03-23
Save config during Install

jsancio:


cosmos-render/src/main/scala/com/mesosphere/cosmos/render/MarathonLabels.scala
```diff
@@ -48,7 +49,8 @@ final class MarathonLabels(
     Seq(
       commandJson.map(
         command => MarathonApp.commandLabel -> MarathonLabels.encodeForLabel(command)
-      )
+      ),
+      Some(MarathonApp.optionsLabel -> MarathonLabels.encodeForLabel(options))
```
jsancio: This is the important change. Everything else falls out of this.

cosmos-server/src/test/scala/com/mesosphere/cosmos/circe/EncodersDecodersSpec.scala
```diff
@@ -305,4 +305,13 @@ class EncodersDecodersSpec extends FreeSpec with PropertyChecks with Matchers {
...
   }
 
+  "Base64 Decoder" - {
+    "basic example" in {
+      val ls = List("hello", "world")
+      val string = ls.asJson.noSpaces
+      val string64 = Base64.getEncoder.encodeToString(string.getBytes(StandardCharsets.UTF_8))
+      assertResult(ls)(decode64[List[String]](string64))
+    }
+  }
```
cruhland: Add a similar test for `parse64`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/internal/model/package.scala
```diff
@@ -28,7 +37,18 @@ package object model {
...
     } yield PackageOrigin(originUri)
 
-    def packageMetadata: Option[String] = app.labels.get(MarathonApp.metadataLabel)
+    def packageMetadata: label.v1.model.PackageMetadata = {
+      decode64[label.v1.model.PackageMetadata](
+        app.labels.get(MarathonApp.metadataLabel).getOrElse("")
+      )
+    }
+
+    def serviceOptions: Option[JsonObject] = {
```
cruhland: Is this used anywhere? Should there be a test for it?

cosmos-render/src/test/scala/com/mesosphere/cosmos/render/MarathonLabelsSpec.scala
```diff
@@ -13,53 +12,106 @@ import com.mesosphere.universe.v3.syntax.PackageDefinitionOps._
...
-import scala.util.Right
+import org.scalatest.Matchers
 import scala.util.Success
 import scala.util.Try
 
-final class MarathonLabelsSpec extends FreeSpec {
+final class MarathonLabelsSpec extends FreeSpec with Matchers {
 
   val RepoUri = Uri.parse("some/repo/uri")
+  val Options = JsonObject.fromMap(
```
larioj: Why are these guys not private?

larioj: I believe our convention is that values like this go in the companion object.

cosmos-render/src/test/scala/com/mesosphere/cosmos/render/MarathonLabelsSpec.scala
```diff
@@ -13,53 +12,106 @@ import com.mesosphere.universe.v3.syntax.PackageDefinitionOps._
...
-        .map(JsonUtil.decode64[universe.v3.model.Command](_).toOption.get)
-    )
+
+  "nonOverridableLabels" - {
+    "minimal" in {
+      MarathonLabels(
+        TestingPackages.MinimalV3ModelV2PackageDefinition,
+        RepoUri,
+        Options
+      ).nonOverridableLabels shouldBe (
```
larioj: The paren after `shouldBe` is unnecessary.

cosmos-render/src/test/scala/com/mesosphere/cosmos/render/MarathonLabelsSpec.scala
```diff
@@ -13,53 +12,106 @@ import com.mesosphere.universe.v3.syntax.PackageDefinitionOps._
...
+    }
+
+    "maximal" in {
+      val labels = MarathonLabels(
+        TestingPackages.MaximalV3ModelV2PackageDefinition,
+        RepoUri,
+        Options
+      ).nonOverridableLabels
+
+      decode64[universe.v3.model.Command](labels(MarathonApp.commandLabel)) shouldBe (
```
larioj: The paren after `shouldBe` is unnecessary.

cosmos-render/src/test/scala/com/mesosphere/cosmos/render/MarathonLabelsSpec.scala
```diff
@@ -13,53 +12,106 @@ import com.mesosphere.universe.v3.syntax.PackageDefinitionOps._
 import com.netaporter.uri.Uri
```
larioj: `import com.mesosphere.universe.common.JsonUtil` is not used.

cosmos-render/src/test/scala/com/mesosphere/cosmos/render/MarathonLabelsSpec.scala
larioj: Also let's add types or make them private.

larioj: Why is this name capitalized?

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Decoders.scala
```diff
@@ -4,18 +4,37 @@ import com.mesosphere.cosmos.CirceError
...
+      case Right(result) => result
+      case Left(error) => throw CirceError(error)
+    }
+  }
+
+  def decode64[T: Decoder](str: String): T = {
+    decode[T](new String(Base64.getDecoder.decode(str), StandardCharsets.UTF_8))
+  }
+
+  def parse64(str: String): Json = {
```
larioj: This method looks a lot like the one in MarathonLabelsSpec. Why do we need both of them?

jsancio: They are in different modules. The one in the `cosmos-render` cannot throw a CirceError exception because that exception is defined in `cosmos-server`. Instead it returns a `Either[io.circe.Error, Json]`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/internal/model/package.scala
jsancio: Not yet. This is going to get used when we implement the `/service/update` endpoint.

cosmos-render/src/test/scala/com/mesosphere/cosmos/render/MarathonLabelsSpec.scala
jsancio: Done.

cosmos-render/src/test/scala/com/mesosphere/cosmos/render/MarathonLabelsSpec.scala
jsancio: Done. I was just following the previous pattern. Here it doesn't matter but capitalization matters when patter matching. E.g. this two are different:

```scala
val two = 2
Some(3) match {
  case Some(two) => println(s"3 == $two")
  case _ =>
}
```
vs
```scala
val Two = 2
Some(2) match {
  case Some(Two) => println(s"2 == $Two")
  case _ =>
}
```

cosmos-render/src/test/scala/com/mesosphere/cosmos/render/MarathonLabelsSpec.scala
jsancio: Done.

cosmos-render/src/test/scala/com/mesosphere/cosmos/render/MarathonLabelsSpec.scala
jsancio: Done.

cosmos-server/src/test/scala/com/mesosphere/cosmos/circe/EncodersDecodersSpec.scala
jsancio: Done.

cosmos-server/src/main/scala/com/mesosphere/cosmos/circe/Decoders.scala
cruhland: You could potentially keep the methods where they were originally defined, in `cosmos-json`, and have them return `Either[io.circe.Error, Json]`. Then in `cosmos-server` you could define wrapper methods that just call `.valueOr(err => throw CirceError(err))` on the results.

Not sure if it's worth it, but it avoids duplicate implementations.

jsancio: I am going to leave it as is.

I would like us to think about getting rid of all of these modules. Not clear to me what we are gaining given the extra burden.

cosmos-model/src/main/scala/com/mesosphere/cosmos/thirdparty/marathon/model/MarathonApp.scala
```diff
@@ -21,4 +21,5 @@ object MarathonApp {
   val repositoryLabel = "DCOS_PACKAGE_SOURCE"
```
larioj: Just as a note. The frameworkNameLabel appears to be unused.

jsancio: Good catch. Fixed. We were not using constant variable but instead were copying the string.

cruhland:
@jsancio I'm going to wait to merge https://github.com/dcos/sbt-dcos/pull/3 until this is merged, otherwise the build will break.

PR #591, opened 2017-03-27
Move out PackageDefinition Inner classes and objects

larioj:
In order to add a v4 Package, we need to clean up the PackageDefinition hierarchy. One part of this clean up is to get all the inner objects out of PackageDefinition companion object. This is important so that these objects can be used in other packages, and most importantly versioned independently.

### DIFF
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/ReleaseVersion.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/ReleaseVersion.scala
@@ -1,0 +1,24 @@
+package com.mesosphere.universe.v3.model
+
+import com.twitter.util.Return
+import com.twitter.util.Throw
+import com.twitter.util.Try
+
+final class ReleaseVersion private(val value: Long) extends AnyVal
+
+object ReleaseVersion {
+
+  def apply(value: Long): Try[ReleaseVersion] = { //TODO: do not return Try. Most call sites do .get
+    if (value >= 0) {
+      Return(new ReleaseVersion(value))
+    } else {
+      val message = s"Expected integer value >= 0 for release version, but found [$value]"
+      Throw(new IllegalArgumentException(message))
+    }
+  }
+
+  implicit val packageDefinitionReleaseVersionOrdering: Ordering[ReleaseVersion] = {
+    Ordering.by(_.value)
+  }
+
+}
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/ReleaseVersion.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/ReleaseVersion.scala
@@ -8,7 +8,7 @@ final class ReleaseVersion private(val value: Long) extends AnyVal
 
 object ReleaseVersion {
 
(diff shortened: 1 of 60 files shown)


larioj:
A little bit of context:

This is in large part a rename. I moved PackageDefinition.Tag, PackageDefinition.Version, and PackageDefinition.ReleaseVersion, out into the universe.v3.model package.

The only noteworthy thing during the rename is that since Version is now part of universe.v3.model, any files that imported DcosReleaseVersion.Version had to be changed to refer to the fully qualified name. Other than that there rename was fairly straightforward.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/ReleaseVersion.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import com.twitter.util.Return
+import com.twitter.util.Throw
+import com.twitter.util.Try
+
+final class ReleaseVersion private(val value: Long) extends AnyVal
+
+object ReleaseVersion {
+
+  def apply(value: Long): Try[ReleaseVersion] = { //TODO: do not return Try. Most call sites do .get
```
cruhland: Don't forget about this TODO. You might want to do something similar to what I did with the `Path` stuff: have an `apply` method that throws an exception, which delegates to a `validate` method that returns a `Try` or an `Either`: https://github.com/dcos/cosmos/blob/6fab55dc505e6cef26d3f429268fb1cdf3b2847f/cosmos-common/src/main/scala/com/mesosphere/util/AbsolutePath.scala#L28-L39

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/Tag.scala
```diff
@@ -0,0 +1,29 @@
...
+  override def toString: String = value
+
+}
+
+object Tag {
+
+  val packageDetailsTagRegex: String = "^[^\\s]+$"
+  val packageDetailsTagPattern: Pattern = Pattern.compile(packageDetailsTagRegex)
+
+  def apply(s: String): Try[Tag] = { //TODO: Get rid of this Try. Most call sites do .get
```
cruhland: See [my comment on `ReleaseVersion`](https://github.com/dcos/cosmos/pull/591/files#r108242853).

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/ReleaseVersion.scala
larioj: I was thinking the todo, was for a later PR :)

larioj: This already has a lot of changes.

cruhland: Ah. It's fine to defer it, but I'd prefer not to have the comment in the code, just in case you don't get around to it. Much better to file a JIRA, unless you're going to literally do it next.

larioj: Okay, I'll remove the comment.

PR #592, opened 2017-03-28
Move PackageDefinition and PackageVersion out of module

larioj:
This is a simple change. All it does is move PackageDefinition and PackageVersion into Files under com.mesosphere.universe. All other things are maintained by making the objects live under a v3.model package.

The reason for this change is that it helps with the sealed hierarchy restriction. With this organization we can have a sealed hierarchy spanning multiple packages.

In particular we will be able to add V4Package to package v4.model {}, but have it still be part of the PackageDefinition hierarchy.


larioj:
@jsancio Yeah, it was one of the first things I checked. Exhaustiveness checks work too.

cosmos-model/src/main/scala/com/mesosphere/universe/PackageVersion.scala
```diff
@@ -0,0 +1,34 @@
+package com.mesosphere.universe
```
cruhland: Rename this file to `PackagingVersion.scala`, since that corresponds to the class name?

larioj: Thanks :)

PR #593, opened 2017-03-28
Move metadata to top level and create hierarchy

larioj:
Metadata is PackageDefinition minus a few fields. When adding a V4 Package, we will also need to add a V4Metadata. This change moves metadata out to universe, and creates a hierarchy akin to that of PackageDefinition.


cosmos-json/src/main/scala/com/mesosphere/universe/v3/circe/Encoders.scala
```diff
@@ -53,5 +53,8 @@ object Encoders {
   implicit val encodeV3Package: Encoder[V3Package] = deriveEncoder[V3Package]
   implicit val encodeV3Resource: Encoder[V3Resource] = deriveEncoder[V3Resource]
 
-  implicit val encodeMetadata: Encoder[Metadata] = deriveEncoder[Metadata]
+  implicit val encodeV3Metadata: Encoder[V3Metadata] = deriveEncoder[V3Metadata]
+  implicit val encodeMetadata: Encoder[Metadata] = Encoder.instance {
+    case v3Metadata: V3Metadata => v3Metadata.asJson
```
cruhland: Won't this fail the exhaustiveness check?

larioj: No, the exhaustiveness check works. FYI, this is how we encode package definition.

cruhland: Oh, I was getting confused by the other match, but that one is on `PackagingVersion`, which has a `V2` instance. Never mind.

PR #597, opened 2017-04-05
Move universe.v3.model decoders into companion objects

larioj:
Having decoders in a single file has lead to a bunch of hard to debug type and name clashes. @jsancio and I have discussed having the decoders and encoders in the companion objects moving forward with the addition of v4 packages. However, since cosmos-model does not depend on cosmos-json adding coders to the v4 package companion object is impossible. This PR moves the coders in universe.v3.model to their respective companion objects, making it possible to have the coders in the v4 package companion object.


jsancio:
So nice. It is hard to tell but did we completely remove the `cosmos-json` module?

larioj:
cosmos-json exists, it still holds rpc coders

build.sbt
```diff
@@ -25,6 +25,7 @@ lazy val model = project.in(file("cosmos-model"))
     libraryDependencies ++=
       Deps.scalaUri
         ++ Deps.circeCore
+        ++ Deps.circe
```
cruhland: If no other project is using only `Deps.circeCore`, let's combine it with `Deps.circe`.

cosmos-model/src/main/scala/com/mesosphere/universe/v3/model/Assets.scala
```diff
@@ -1,6 +1,17 @@
...
+
 case class Assets(
   uris: Option[Map[String, String]],
   container: Option[Container]
 )
+
+object Assets {
+  implicit val decodeAssets: Decoder[Assets] = deriveDecoder[Assets]
+  implicit val encodeAssets: Encoder[Assets] = deriveEncoder[Assets]
+}
```
cruhland: This pattern is happening quite often. Could we define a base class that these companion objects can implement? If we need a custom definition for one or both of the coders we can always override the method. Example:

```scala
abstract class DerivedCoders[A](implicit stuff: /* we will probably need some things here for the `deriveX` methods below */) {
  implicit def decode: Decoder[A] = deriveDecoder[A]
  implicit def encode: Encoder[A] = deriveEncoder[A]
}

object Assets extends DerivedCoders[Assets]
```

Does that work?

build.sbt
larioj: Deps.circe includes Deps.circeCore. I'll remove Deps.circeCode on the previous line, but I'll keep Deps.scala unchanged.

cruhland: I guess what I mean is, I think `Deps.circeCore` is not used anywhere else besides here. So once you get rid of it here, there's no reason for it to exist; it can be merged into `Deps.circe` and we can just use that.

PR #599, opened 2017-04-07
Make PackagingVersion and PackageDefinition sealed traits

larioj:
DescribeResponse depends on v3 PackageVersion, which means we may need to keep it around. Making it a sealed trait makes this much easier.

### DIFF
--- a/cosmos-model/src/main/scala/com/mesosphere/universe/PackagingVersion.scala
+++ b/cosmos-model/src/main/scala/com/mesosphere/universe/PackagingVersion.scala
@@ -58,6 +58,20 @@ package v3.model {
         }
       }
     }
+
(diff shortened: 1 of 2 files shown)


cosmos-model/src/main/scala/com/mesosphere/universe/PackagingVersion.scala
```diff
@@ -45,37 +50,63 @@ package v3.model {
...
-    def packagingVersionSubclassToString[V <: PackagingVersion](
-      expected: V
-    ): Decoder[V] = {
-      Decoder.instance { (c: HCursor) =>
-        c.as[String].flatMap {
-          case expected.show => Right(expected)
-          case x => Left(DecodingFailure(
-            s"Expected value [${expected.show}] for packaging version, but found [$x]",
-            c.history
-          ))
```
cruhland: We can't use this approach anymore?

PR #600, opened 2017-04-10
Give integration tests their own module

cruhland:
On Friday I investigated whether SBT could be made to depend on artifacts in the `IntegrationTest` configuration from a Maven repository. It turned out that SBT's POM parser [hardcodes the configurations that it supports](https://github.com/sbt/ivy/blob/2.3.x-sbt/src/java/org/apache/ivy/plugins/parser/m2/PomModuleDescriptorBuilder.java#L70), and `IntegrationTest` is not included.

This PR abandons that approach, and instead publishes the integration tests as a standard (`Compile` configuration) JAR. I also had to create a test util module so that some dependencies are available to the integration test JAR while still being available to "regular" modules in the project.

One nice result of this is the build settings get much simpler!


build.sbt
```diff
@@ -120,3 +138,57 @@ lazy val server = project.in(file("cosmos-server"))
     render % "compile;test->test",
     common % "compile;test->test"
   )
+
+/**
+ * Common test code. Sources are located in the "main" subdirectory so the JAR can be
+ * published to Maven repositories with a standard POM.
+ */
+lazy val testUtil = project.in(file("cosmos-test-util"))
```
jsancio: Not a fan of packages, classes, projects named "util". How about we use "common" instead?

jsancio: This will match in the future when we merge a lot of them projects. One possible future organization: common, test-common, cosmos-server, universe-service, etc.

cruhland: Ok, I'll rename to `cosmos-test-common`.

PR #601, opened 2017-04-12
Move all unit test code to cosmos-test-common

cruhland:
This is a follow-up to #600. It simplifies the module structure of the unit tests, making it easier to use them as a library dependency.


build.sbt
```diff
@@ -63,8 +62,7 @@ lazy val bijection = project.in(file("cosmos-bijection"))
       Deps.bijection
   )
   .dependsOn(
-    model % "compile;test->test",
-    testCommon % "test->compile"
+    model % "compile;test->test"
```
cruhland: Just noticed that we don't need the `test->test` part anymore. I'll update.

PR #602, opened 2017-04-14
Store only the user provided options

jsancio:
Change the options JSON object that we store in Marathon so that we only
store the user provided options. If the user doesn't provide any option
that is equivalent to the empty JSON objct. E.g. "{}".

The possible values for the DCOS_PACKAGE_OPTIONS are:
1. Missing - the application was deployed with an old
Cosmos so we don't have the user provided options.
2. Empty JSON object - the user didn't provide any options
3. JSON object - contains the user provided options

### DIFF
--- a/cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
+++ b/cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
@@ -56,7 +56,11 @@ object PackageDefinitionRenderer {
           renderTemplate(m.v2AppMustacheTemplate, mergedOptions).flatMap { mJson =>
             extractLabels(Json.fromJsonObject(mJson))
               .map { existingLabels =>
-                val labels = MarathonLabels(pkgDef, sourceUri, defaultOptionsAndUserOptions)
+                val labels = MarathonLabels(
+                  pkgDef,
+                  sourceUri,
+                  options.getOrElse(JsonObject.empty)
+                )
 
                 (marathonAppId.map(appIdDoc) ::
                   generateLabelsPartialObjects(labels, existingLabels).map(Some(_))
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageInstallHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageInstallHandler.scala
@@ -8,7 +8,14 @@ import com.mesosphere.cosmos.PackageRunner
 import com.mesosphere.cosmos.ServiceAlreadyStarted
 import com.mesosphere.cosmos.finch.EndpointHandler
 import com.mesosphere.cosmos.http.RequestSession
-import com.mesosphere.cosmos.render._
+import com.mesosphere.cosmos.render.InvalidLabelSchema
+import com.mesosphere.cosmos.render.MissingMarathonV2AppTemplate
+import com.mesosphere.cosmos.render.OptionsNotAllowed
+import com.mesosphere.cosmos.render.OptionsValidationFailure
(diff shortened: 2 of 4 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageInstallHandler.scala
```diff
@@ -34,8 +34,12 @@ private[cosmos] final class PackageInstallHandler(
...
       )
       .flatMap { case (pkg, sourceUri) =>
-        val packageConfig =
-          PackageDefinitionRenderer.renderMarathonV2App(sourceUri, pkg, request.options, request.appId)
+        val packageConfig = PackageDefinitionRenderer.renderMarathonV2App(
+          sourceUri,
+          pkg,
+          request.options,
+          request.appId
+        )
```
cruhland: Looks like this is just a formatting change?

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -56,7 +56,11 @@ object PackageDefinitionRenderer {
           renderTemplate(m.v2AppMustacheTemplate, mergedOptions).flatMap { mJson =>
             extractLabels(Json.fromJsonObject(mJson))
               .map { existingLabels =>
-                val labels = MarathonLabels(pkgDef, sourceUri, defaultOptionsAndUserOptions)
```
cruhland: Is there anything relying on the combined options being in this label? For example, the UI?

cruhland:
@jsancio for possible values of the options, what about the case where an older app is accessed by a newer cosmos, and the combined options were set in the Marathon label? How would we distinguish that from the newer label that only has the user options?

I know we're not reading the value of that label yet, but just thinking for when we do.

jsancio:
> @jsancio for possible values of the options, what about the case where an older app is accessed by a newer cosmos, and the combined options were set in the Marathon label? How would we distinguish that from the newer label that only has the user options?

@cruhland That cosmos doesn't exists. We never shipped a Cosmos that saves the "merged" options or any options at all.

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
jsancio: Not yet. We are doing that next. @larioj has a couple of issues that require reading the above options: `service/describe` and `service/update`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageInstallHandler.scala
jsancio: Yes. Sorry. Formatting change.

cruhland:
> @cruhland That cosmos doesn't exists. We never shipped a Cosmos that saves the "merged" options or any options at all.

Isn't that what this code was doing? https://github.com/dcos/cosmos/pull/602/files#diff-37e3af6098e6a689e70b948eb2c59565L59

cosmos-render/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
cruhland: To clarify my original comment: I'm asking about the `defaultOptionsAndUserOptions` value that **used** to be encoded in the label. Is there anything that relies on the fact that the default options are in there, and will break when they are no longer there?

cruhland: Resolved offline.

PR #605, opened 2017-04-17
Add v4 test examples

larioj:


cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
```diff
@@ -78,6 +78,22 @@ object Generators {
...
 
+  val genV4Package: Gen[universe.v4.model.V4Package] = {
+    genV3Package.map { v3 =>
+      universe.v4.model.V4Package(
+        name = v3.name,
+        version = v3.version,
+        releaseVersion = v3.releaseVersion,
+        maintainer = v3.maintainer,
+        description = v3.description
+      )
```
jsancio: How about the new properties: `upgradesFrom`, downgradeTo`? Should we generate those too?

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
```diff
@@ -78,6 +78,22 @@ object Generators {
...
+    }
+  }
+
+  val genSupportedPackageDefinition: Gen[universe.v4.model.SupportedPackageDefinition] = {
+    Gen.oneOf(genV4Package, genV3Package)
+  }
```
jsancio: How can we make sure that we update this when we add new `SupportedPackageDefinition`s?

larioj: No, anything we do would be hacky.

cruhland: How about a test that uses the generator, and passes the generated values to a function that does a pattern match on its input? When we add a new subclass, the compiler should detect that the match is not exhaustive. We'll have to include a comment explaining that the pattern match is important.

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
cruhland: If we're going to the trouble to update the generators, should we extend them to generate all fields? You could prevent some repetition by generating a common subset of fields, and `flatMap`ing with another generator that creates the actual `V3` and `V4Package`s.

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
```diff
@@ -78,6 +78,28 @@ object Generators {
...
+      v3 <- genV3Package
+    } yield {
+      universe.v4.model.V4Package(
+        name = v3.name,
+        version = v3.version,
+        releaseVersion = v3.releaseVersion,
+        maintainer = v3.maintainer,
+        description = v3.description,
+        upgradesFrom = upgradesFrom,
+        downgradesTo = downgradesTo
```
cruhland: Copying my comment in case it got lost:

> If we're going to the trouble to update the generators, should we extend them to generate all fields? You could prevent some repetition by generating a common subset of fields, and flatMaping with another generator that creates the actual V3 and V4Packages.

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
cruhland: I don't think a separate test case specifically for the purpose of catching updates to `SupportedPackageDefinition` would be hacky. It wouldn't change how we write the generator code, and it would be isolated to one test case.

cosmos-test-common/src/main/scala/com/mesosphere/universe/test/TestingPackages.scala
```diff
@@ -381,4 +382,13 @@ object TestingPackages {
   val MaximalV4ModelPackageDefinitionV4: universe.v4.model.PackageDefinition = MaximalV4ModelV4PackageDefinition
   val MinimalV4ModelPackageDefinitionV4: universe.v4.model.PackageDefinition = MinimalV4ModelV4PackageDefinition
 
+  val supportedPackageDefinitions: TableFor1[universe.v4.model.SupportedPackageDefinition] =
+    new TableFor1(
+      "supportedPackageDefinition",
+      MinimalV3ModelV3PackageDefinition,
+      MaximalV3ModelV3PackageDefinition,
+      MinimalV4ModelV4PackageDefinition,
+      MaximalV4ModelV4PackageDefinition
```
cruhland: Can we replace this table with the generator?

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
jsancio: @larioj If you implement @cruhland suggestion lets make sure we do the same thing to the `PackageDefinition` generator.

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
larioj: I'm don't have a lot of time allotted to this task. I'd rather add tests that check v4 packages, than spend a lot of time trying to fill in all the fields in these packages. There is a reason they were not filled in in the first place.

cosmos-test-common/src/main/scala/com/mesosphere/universe/test/TestingPackages.scala
larioj: Yeah, eventually. Not on this PR though, or on this issue for that matter. This changes are a good step towards that though.

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
```diff
@@ -78,6 +78,42 @@ object Generators {
...
+      )
+    }
+  }
+
+  // <block>
```
cruhland: What's this `<block>` tag for?

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
cruhland: Ok, makes sense.

cosmos-test-common/src/main/scala/com/mesosphere/universe/test/TestingPackages.scala
cruhland: Ok!

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
larioj: This is my way of saying that these two functions belong together. I don't know of a better way.

cosmos-test-common/src/test/scala/com/mesosphere/Generators.scala
larioj: @jsancio We are not generating packageDefinitions.

PR #607, opened 2017-04-19
Add more V4 Package Tests

larioj:


cosmos-test-common/src/test/scala/com/mesosphere/cosmos/render/PackageDefinitionRendererSpec.scala
```diff
@@ -9,7 +9,8 @@ import com.mesosphere.cosmos.label.v1.circe.Decoders._
 import com.mesosphere.cosmos.thirdparty.marathon.circe.Decoders.decodeAppId
```
jsancio: @larioj I am going to create an issue to refactor this code so can easily test rendering against all of the package version.

larioj: @jsancio Yeah, that's a good idea.

cruhland: Yeah, for that `"property add resource object to options"` test in particular, there is a lot of duplicated code which would be nice to get rid of.

PR #611, opened 2017-04-21
Finish adding v4 examples to unit tests

larioj:
This PR changes GarbageCollectorSpec to use the SupportedPackageDefinition generator. It does not alter anything else.

InstallQueueSpec: I wrapped the spec in a definition, parameterized on universeInstall. That way it can be passed the UniverseInstall with a particular package version. Other than the version of package all tests are the same.


cruhland:
@larioj the diff for `InstallQueueSpec.scala` is not very helpful. It looks like you just wrapped the existing test cases in a helper method, but could you add comments for anything else that changed?

larioj:
@cruhland I added the description. Let me know if you have more questions.

PR #613, opened 2017-04-25
Update repo schema

larioj:
The universe checker uses this schema to validate repositories. Since this is outdated, the universe checker is failing v4 repositories.

### DIFF
--- a/cosmos-common/src/main/resources/com/mesosphere/universe/v3/schema/repo
+++ b/cosmos-common/src/main/resources/com/mesosphere/universe/v3/schema/repo
@@ -386,7 +386,8 @@
           "type": "string"
         },
         "version": {
-          "type": "string"
+          "type": "string",
+          "pattern": "^[-a-zA-Z0-9.]+$"
         },
         "releaseVersion": {
           "type": "integer",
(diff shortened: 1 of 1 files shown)


cosmos-common/src/main/resources/com/mesosphere/universe/v3/schema/repo
```diff
@@ -386,8 +386,118 @@
           "type": "string"
         },
         "version": {
+          "type": "string",
```
jsancio: This is techncially `.../universe/v4/schema/repo`. Also, cosmos needs to accept both v3 repos and v4 repos.

larioj: v3 repos are a subset of v4 repos. So this will accept v3 repos. I have an outstanding issue to create the v4 schema. But that's for later. This is technically not blocking me, so I could make this change once I do the universe change.

cruhland: Should we have a parallel directory structure then? And have the universe checker use both schemas?

jsancio: Okay then please rename this file to `.../universe/v4/schema/repo` and put a note at the call site that uses this jsonschema saying that it is okay to use this schema to verify v3 repo because it is a subset of v4.

jsancio: Also, don't you need to verify that v3 repos are not v4 repos? How are we going to do that?

PR #614, opened 2017-04-25
Add upgradesTo() method to PackageCollection

cruhland:
I ended up going with the generator approach instead of the view approach that we talked about the other day. It does make the generators more complicated; I'll point out where. I decided not to use the view because it seemed too heavyweight a change for a simple feature.

### DIFF
--- a/cosmos-common/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
@@ -8,8 +8,6 @@ import io.circe.syntax.EncoderOps
 
 final class PackageDefinitionOps(val pkgDef: universe.v4.model.PackageDefinition) extends AnyVal {
 
-  import PackageDefinitionOps._
-
   def packageCoordinate: PackageCoordinate =  {
     PackageCoordinate(name, version)
   }
@@ -138,13 +136,13 @@ final class PackageDefinitionOps(val pkgDef: universe.v4.model.PackageDefinition
     case v4: universe.v4.model.V4Package => v4.resource
   }
 
-  def upgradesFrom: Option[List[universe.v3.model.Version]] = pkgDef match {
-    case v4: universe.v4.model.V4Package => v4.upgradesFrom
-    case _ => None
+  def upgradesFrom: List[universe.v3.model.VersionSpecification] = pkgDef match {
+    case v4: universe.v4.model.V4Package => v4.upgradesFrom.getOrElse(Nil)
+    case _ => Nil
   }
 
   def canUpgradeFrom(version: universe.v3.model.Version): Boolean = {
-    upgradesFrom.exists(_.exists(v => v == AnyVersion || v == version))
+    upgradesFrom.exists(_.matches(version))
   }
 
   // -------- Non top-level properties that we are safe to "jump" to --------------
@@ -172,8 +170,6 @@ final class PackageDefinitionOps(val pkgDef: universe.v4.model.PackageDefinition
 
 object PackageDefinitionOps {
   import scala.language.implicitConversions
-
-  val AnyVersion: universe.v3.model.Version = universe.v3.model.Version("*")
 
   implicit def packageDefinitionToPackageDefinitionOps[P <: universe.v4.model.PackageDefinition](
     pkgDef: P
--- a/cosmos-common/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
@@ -8,6 +8,8 @@ import io.circe.syntax.EncoderOps
 
 final class PackageDefinitionOps(val pkgDef: universe.v4.model.PackageDefinition) extends AnyVal {
 
+  import PackageDefinitionOps._
+
   def packageCoordinate: PackageCoordinate =  {
     PackageCoordinate(name, version)
   }
@@ -136,6 +138,15 @@ final class PackageDefinitionOps(val pkgDef: universe.v4.model.PackageDefinition
     case v4: universe.v4.model.V4Package => v4.resource
   }
 
+  def upgradesFrom: Option[List[universe.v3.model.Version]] = pkgDef match {
+    case v4: universe.v4.model.V4Package => v4.upgradesFrom
+    case _ => None
+  }
+
+  def canUpgradeFrom(version: universe.v3.model.Version): Boolean = {
+    upgradesFrom.exists(_.exists(v => v == AnyVersion || v == version))
+  }
+
   // -------- Non top-level properties that we are safe to "jump" to --------------
 
   def images: Option[universe.v3.model.Images] = pkgDef match {
@@ -161,6 +172,8 @@ final class PackageDefinitionOps(val pkgDef: universe.v4.model.PackageDefinition
 
 object PackageDefinitionOps {
   import scala.language.implicitConversions
+
+  val AnyVersion: universe.v3.model.Version = universe.v3.model.Version("*")
 
   implicit def packageDefinitionToPackageDefinitionOps[P <: universe.v4.model.PackageDefinition](
     pkgDef: P
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
@@ -92,6 +92,21 @@ final class MultiRepository(
     }
   }
 
+  override def upgradesTo(
+    name: String,
+    version: universe.v3.model.Version
+  )(implicit session: RequestSession): Future[List[universe.v3.model.Version]] = {
+    repositories().flatMap { repositories =>
+      Future.collect {
+        repositories.map { repository =>
+          repository.getPackagesByPackageName(name).map { packages =>
+            PackageCollection.upgradesTo(name, version, packages)
+          }
+        }
+      }
+    } map { versions => versions.flatten.toList }
+  }
+
   def getRepository(uri: Uri): Future[Option[CosmosRepository]] = {
     repositories().map { repositories =>
       repositories.find(_.repository.uri == uri)
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
@@ -99,9 +99,7 @@ final class MultiRepository(
     repositories().flatMap { repositories =>
       Future.collect {
         repositories.map { repository =>
-          repository.getPackagesByPackageName(name).map { packages =>
-            PackageCollection.upgradesTo(name, version, packages)
-          }
+          repository.upgradesTo(name, version)
         }
       }
     } map { versions => versions.flatten.toList }
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
@@ -3,6 +3,7 @@ package com.mesosphere.cosmos.repository
 import com.mesosphere.cosmos.http.RequestSession
 import com.mesosphere.cosmos.rpc
 import com.mesosphere.universe
+import com.mesosphere.universe.v3.syntax.PackageDefinitionOps._
 import com.netaporter.uri.Uri
 import com.twitter.util.Future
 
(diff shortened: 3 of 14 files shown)


cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
```diff
@@ -50,21 +56,47 @@ object Generators {
...
   private val genReleaseVersion: Gen[universe.v3.model.ReleaseVersion] = for {
     num <- Gen.posNum[Long]
   } yield universe.v3.model.ReleaseVersion(num).get
 
-  val genV3Package: Gen[universe.v3.model.V3Package] = {
-    val maxPackageNameLength = 64
-    val genPackageNameChar = Gen.oneOf(Gen.numChar, Gen.alphaLowerChar)
-    val genPackageName = maxSizedString(maxPackageNameLength, genPackageNameChar)
+  def genUpgradesFrom(
+    requiredVersion: Option[universe.v3.model.Version]
```
cruhland: Here's one complexity: the `requiredVersion` parameter. If it's `Some`, the specified version will appear in the return value. It might be simpler to split this into two definitions.

cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
```diff
@@ -50,21 +56,47 @@ object Generators {
...
+  ): Gen[Option[List[universe.v3.model.Version]]] = {
+    genVersionSet(requiredVersion).flatMap(genUpgradesFrom)
+  }
 
+  private def genUpgradesFrom(versionSet: VersionSet): Gen[Option[List[universe.v3.model.Version]]] = {
```
cruhland: I created a `VersionSet` type to simplify writing the generator for the `upgradesFrom` field. It ensures that all important variations of the field value are covered.

cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
```diff
@@ -78,11 +110,30 @@ object Generators {
...
+        maintainer = v3.maintainer,
+        description = v3.description,
+        marathon = universe.v3.model.Marathon(marathonTemplate)
+      )
+    }
+  }
+
+  def genV4Package(
+    genName: Gen[String] = genPackageName,
+    genUpgrades: Gen[Option[List[universe.v3.model.Version]]] = genUpgradesFrom(requiredVersion = None)
```
cruhland: Here's another place where the generator has become more complicated -- these parameters can be used to adjust the behavior of the `name` and `upgradesFrom` fields.

project/Deps.scala
```diff
@@ -95,7 +95,7 @@ object Deps {
 
   // APLv2.0
   val scalaCheckShapeless = Seq(
-    "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % V.scalaCheckShapeless % "test"
+    "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % V.scalaCheckShapeless
```
cruhland: I moved `Generators.scala` to `main` so that it can be used downstream of cosmos; it needs to use this library.

cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
cruhland: As something to consider for the future, `VersionSet` more closely models the semantics of the `upgradesFrom` field than its current type (which needs to have the same structure as the JSON schema). If we ever have an internal representation of packages, we could use `VersionSet` directly in it.

cosmos-common/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
```diff
@@ -136,6 +138,15 @@ final class PackageDefinitionOps(val pkgDef: universe.v4.model.PackageDefinition
     case v4: universe.v4.model.V4Package => v4.resource
   }
 
+  def upgradesFrom: Option[List[universe.v3.model.Version]] = pkgDef match {
+    case v4: universe.v4.model.V4Package => v4.upgradesFrom
+    case _ => None
+  }
```
jsancio: Let's make this type `List[Version]`. @larioj defined this property so that `Nil` and `None` are semantically the same.

jsancio: Also, this type is slightly different from `Version`. It is technically a version specification. We currently support exact matches and wildcard matches. Eventually, I would like us to support something like this: http://doc.crates.io/specifying-dependencies.html#specifying-dependencies-from-cratesio

cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
```diff
@@ -50,21 +56,47 @@ object Generators {
...
+  private def genVersionSet(requiredVersion: Option[universe.v3.model.Version]): Gen[VersionSet] = {
+    val genExplicit = Gen.containerOf[Set, universe.v3.model.Version](genVersion)
+      .map(otherVersions => ExplicitVersions(otherVersions union requiredVersion.toSet))
+
+    Gen.frequency((1, AllVersions), (10, genExplicit))
+  }
+
+  private sealed trait VersionSet
+  private case class ExplicitVersions(versions: Set[universe.v3.model.Version]) extends VersionSet
+  private case object AllVersions extends VersionSet
```
jsancio: I see. I think in the future we want `upgradesFrom: List[VersionSpecification]` where `canUpgradeFrom(x) = upgradesFrom.exist(_.matches(x))`. `matches` is a method that returns `True` if the give version matches the specification. cc @larioj .

E.g. If `upgradesFrom: List("1.2.*", "2.*", "3.1.1")` then `"1.2.3"`, `"2.5"` and `"3.1.1"` will returns `True` while `"1.3.2"`, `"3.1.2"` will return `False`.

cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
```diff
@@ -50,21 +56,47 @@ object Generators {
...
+  def genUpgradesFrom(
+    requiredVersion: Option[universe.v3.model.Version]
+  ): Gen[Option[List[universe.v3.model.Version]]] = {
+    genVersionSet(requiredVersion).flatMap(genUpgradesFrom)
+  }
 
+  private def genUpgradesFrom(versionSet: VersionSet): Gen[Option[List[universe.v3.model.Version]]] = {
+    versionSet match {
+      case ExplicitVersions(versions) if versions.isEmpty => Gen.oneOf(None, Some(Nil))
+      case ExplicitVersions(versions) => Some(versions.toList)
```
jsancio: Isn't the type here `Some[List[Version]]` yet you want `Gen[Option[List[Version]]]`. Does scalacheck have an implicit `t: T => Gen[T]` that always returns `t`?

cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
jsancio: Can you document what this method is suppose to do?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -24,3 +25,21 @@ trait PackageCollection {
...
+   */
+  def upgradesTo(
+    name: String,
+    version: universe.v3.model.Version,
+    possibleUpgrades: List[universe.v4.model.PackageDefinition]
+  ): List[universe.v3.model.Version] = {
+    possibleUpgrades.collect {
+      case pkg if pkg.name == name && pkg.canUpgradeFrom(version) => pkg.version
+    }
+  }
```
jsancio: Cool. I would like it if the rest of the Cosmos code doesn't call this method and instead calls something like`val pkgs: List[PackageDefinition] = (pkgCollection: PackageCollection).upgradesTo(name, version)`. What do you think?

cosmos-common/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
cruhland: Does circe use `Nil` as the default value if the field is omitted? I think that's why `Option` was used here. I'll make sure that's the case before changing it.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
cruhland: Sure, I can add that method.

cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
cruhland: Yeah, there is an implicit conversion. But I didn't intend to use it here; I'll update the code. The implicit conversion is helpful when passing arguments but I find it confusing anywhere else.

cosmos-common/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
cruhland: Turns out circe gives a parse error if a non-`Option` field is missing from the JSON. So I'll have to keep this the way it is. In the future if we have an internal model of a package that's separate from the RPC format, we could use the better type there.

cruhland: Looks like there is a way to use default field values, but you have to enable it: https://github.com/circe/circe/blob/master/modules/generic-extras/src/main/scala/io/circe/generic/extras/Configuration.scala#L8

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
cruhland: @jsancio currently I'm returning `List[Version]` from this method. Should we be returning `List[PackageDefinition]` instead? I used `Version` because that's all the describe endpoint needs, but maybe I overlooked a use case.

cosmos-common/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
jsancio: We can model it as `Option[List[String]]` (or `Option[List[VersionSpecification]]`) in the case class (V4Package). In the `Ops` method above we can do the "right" mapping.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
jsancio: @cruhland Actually, I think it needs to be `Future[List[PackageDefinition]]`.

cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
larioj: Yeah, this seems like a good idea.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
jsancio: Yeah. I think you are right. @larioj should know better if he needs `PackageDefinition` or `Version`.

cruhland:
@jsancio @larioj updated with your suggestions!

cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
```diff
@@ -92,6 +92,21 @@ final class MultiRepository(
...
 
+  override def upgradesTo(
+    name: String,
+    version: universe.v3.model.Version
+  )(implicit session: RequestSession): Future[List[universe.v3.model.Version]] = {
+    repositories().flatMap { repositories =>
+      Future.collect {
+        repositories.map { repository =>
+          repository.getPackagesByPackageName(name).map { packages =>
+            PackageCollection.upgradesTo(name, version, packages)
```
cruhland: Here's the slight inefficiency I mentioned the other day -- we end up filtering on the package name twice.

cruhland: Oops, just noticed I could avoid that by using the instance method, will update.

jsancio: I think we may have to formalize our MultiRepository semantic. For example `getPackagesByPackageName`, will only return packages from the first repo that contains the given package name. In other words if the repository list is repo A & B, and repo A contains packages for the name X, Cosmos will only return the packages for X in repo A.

I think we need to do the same thing here. What do you think? cc @larioj .

jsancio: We talked offline. @cruhland implementation is correct!

PR #615, opened 2017-04-27
Change cosmos to accept v4 repositories

larioj:
Updates cosmos to accept v4 repositories and updates the integration tests to use v4 packages


cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -48,18 +44,25 @@ extends FreeSpec with TableDrivenPropertyChecks with Matchers {
     }
 
     "can successfully describe helloworld" in {
-      describeHelloworld()
```
jsancio: Let's make sure that this still works. We should be able to describe the "latest" package without having to specify a version.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageInstallIntegrationSpec.scala
```diff
@@ -315,7 +315,7 @@ private object PackageInstallIntegrationSpec extends Matchers with TableDrivenPr
     ("expected response", "force version", "Labels"),
     (
       InstallResponse("helloworld", PackageDetailsVersion("0.1.0"), AppId("helloworld")),
-      false,
+      true,
```
jsancio: Why is this changing?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/PackageRenderHandlerSpec.scala
```diff
@@ -112,6 +112,13 @@ class PackageRenderHandlerSpec extends FreeSpec {
       assertResult(expectedBody)(actualBody)
     }
 
+    "works for v4 packages" in {
+      val renderRequest = RenderRequest("helloworld", Some(PackageDetailsVersion("0.4.1")))
+      val response = packageRender(renderRequest)
+      assertResult(Status.Ok)(response.status)
+      assertResult(MediaTypes.RenderResponse.show)(response.contentType.get)
+    }
```
jsancio: This doesn't test that you got the expected response body.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
```diff
@@ -24,26 +25,25 @@ import scala.util.Left
...
-    originalRepositories.foreach { repo =>
-      addRepo(PackageRepositoryAddRequest(repo.name, repo.uri))
-    }
+  override def afterAll(): Unit = {
+    super.afterAll()
+    ItUtil.replaceRepositoriesWith(originalRepositories)
+  }
+
+  before {
+    ItUtil.replaceRepositoriesWith(defaultRepos)
```
jsancio: Why is this "before" while replacing the original "afterAll"?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositoryIntegrationSpec.scala
```diff
@@ -1,36 +1,37 @@
...
-    originalRepositories.foreach { repo =>
-      addRepository(repo)
-    }
+  override def afterAll(): Unit = {
+    super.afterAll()
+    replaceRepositoriesWith(originalRepositories)
+  }
+
+  before {
+    replaceRepositoriesWith(defaultRepos)
```
jsancio: Why is this "before" while replacing the original "afterAll"?

cosmos-test-common/src/main/scala/com/mesosphere/universe/test/TestingPackages.scala
```diff
@@ -357,7 +358,7 @@ object TestingPackages {
 
   val MaximalV4ModelV4PackageDefinition: universe.v4.model.V4Package = universe.v4.model.V4Package(
     universe.v4.model.V4PackagingVersion,
-    Name,
+    Name + "v4",
```
jsancio: Why the name change in all of the packages above?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
```diff
@@ -0,0 +1,42 @@
...
+
+object ItUtil {
+
+  def listRepositories(): Seq[rpc.v1.model.PackageRepository] = {
+    val request = CosmosRequests.packageRepositoryList
+    val Right(response) = CosmosClient.callEndpoint[rpc.v1.model.PackageRepositoryListResponse](request)
+    response.repositories
+  }
+
+  def deleteRepository(
```
cruhland: There's a method that does something similar in `CosmosClient`. Could you remove that method and have all its callers use this one instead?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
```diff
@@ -0,0 +1,42 @@
...
+    source: rpc.v1.model.PackageRepository
+  ): rpc.v1.model.PackageRepositoryDeleteResponse = {
+    val repoDeleteRequest = rpc.v1.model.PackageRepositoryDeleteRequest(name = Some(source.name))
+    val request = CosmosRequests.packageRepositoryDelete(repoDeleteRequest)
+    val Right(response) = CosmosClient.callEndpoint[rpc.v1.model.PackageRepositoryDeleteResponse](request)
+    response
+  }
+
+  def addRepository(
```
cruhland: There's a method that does something similar in `CosmosClient`. Could you remove that method and have all its callers use this one instead?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ListVersionsSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import _root_.io.circe.syntax._
+import _root_.io.circe.jawn.parse
+import com.mesosphere.cosmos.http.CosmosRequests
+import com.mesosphere.cosmos.test.CosmosIntegrationTestClient.CosmosClient
+import com.twitter.finagle.http.Status
+import org.scalatest.FreeSpec
+import org.scalatest.Matchers
+
+class ListVersionsSpec extends FreeSpec with Matchers {
```
cruhland: `final class`

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ListVersionsSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+  "ListVersionHandler should" - {
+    "list only helloworld versions found in the first repo with helloworld packages" in {
+      val request = CosmosRequests.packageListVersions(
+        rpc.v1.model.ListVersionsRequest("helloworld", true))
```
cruhland: Could you use a named argument for the boolean parameter? I can't tell what it means.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ListVersionsSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+      val response = CosmosClient.submit(request)
```
cruhland: Why not use `CosmosClient.callEndpoint()` and get a strongly-typed response object back? You also wouldn't need to check the status yourself since `callEndpoint()` already does it.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -48,18 +44,25 @@ extends FreeSpec with TableDrivenPropertyChecks with Matchers {
...
     }
 
-    "can successfully describe all versions from Universe" in {
-      forAll (PackageVersionsTable) { (packageName, versions) =>
-        packageListVersionsAndAssert(
-          packageName=packageName,
-          status=Status.Ok,
-          content=versions.asJson
-        )
-      }
```
cruhland: I'm guessing you got rid of this because of the package resolution logic preventing the new v4 packages from being accessed? Let's not delete it; instead it might be a better idea to name the v4 packages something other than `helloworld`.

The `PackageVersionsTable` only has one element though, so you can go ahead and delete that. And this test is the only caller of `packageListVersionsAndAssert()` too, so you can inline that method code here. Should make it easier to understand the test. Looks like you should be able to make use of `CosmosClient.callEndpoint()`.

cruhland: Oh, I think I see...you moved this test to `ListVersionsSpec`, since it's not relevant to `/package/describe` at all. Never mind about keeping it here then, but my other suggestions should still apply.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -48,18 +44,25 @@ extends FreeSpec with TableDrivenPropertyChecks with Matchers {
...
+    "can successfully describe helloworld v4 w/o updates" in {
+      val response = describeRequest(
+        rpc.v1.model.DescribeRequest(
+          "helloworld",
+          Some(universe.v2.model.PackageDetailsVersion("0.4.0")))
+      )
+      assertResult(Status.Ok)(response.status)
```
cruhland: Since you're just testing response status here, this could be a good opportunity to implement a `desribeRequest()` method on `ItUtil` that uses `CosmosClient.callEndpoint()`.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -48,18 +44,25 @@ extends FreeSpec with TableDrivenPropertyChecks with Matchers {
...
+    }
+
+    "fails to describe helloworld v4 w/ updates when requesting v2 describe response" in {
+      val response = describeRequestV2(
+        rpc.v1.model.DescribeRequest(
+          "helloworld",
+          Some(universe.v2.model.PackageDetailsVersion("0.4.1")))
+      )
+      assertResult(Status.BadRequest)(response.status)
```
cruhland: Same here, should be able to use `CosmosRequests.callEndpoint()`.

cruhland: Could you also check that the error type is what we expect?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -137,6 +126,13 @@ extends FreeSpec with TableDrivenPropertyChecks with Matchers {
     CosmosClient.submit(CosmosRequests.packageDescribeV1(describeRequest))
   }
 
+  private def describeRequestV2
+  (
+    describeRequest: rpc.v1.model.DescribeRequest
+  ): Response = {
+    CosmosClient.submit(CosmosRequests.packageDescribeV2(describeRequest))
```
cruhland: Not sure about this, but it might make the tests easier to read if you deleted `describeRequestV2()` and `describeRequest()` and just called the underlying methods directly. Especially if you're switching to using `CosmosClient.callEndpoint()` anyway.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
```diff
@@ -62,6 +61,26 @@ final class PackageListIntegrationSpec
     }
   }
 
+  "Package list endpoint responds with" - {
```
cruhland: Any reason why this moved? Would prefer things stay in the same place if possible so that the diff only shows meaningful changes.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
```diff
@@ -93,28 +112,22 @@ final class PackageListIntegrationSpec
...
+      assertResult(None) {
+        actualDelete.repositories.find(_.name == repository.name)
       }
 
       action
     } finally {
-      val repoAddRequest = PackageRepositoryAddRequest(repository.name, repository.uri)
-      val request = CosmosRequests.packageRepositoryAdd(repoAddRequest)
-      val actualAdd = CosmosClient.callEndpoint[PackageRepositoryAddResponse](request)
+      ItUtil.replaceRepositoriesWith(originalRepositories)
```
cruhland: Much easier to understand! :100:

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositoryIntegrationSpec.scala
cruhland: Yeah, I'm also confused by this. The previous code restored the `originalRepositories` after each test. Now `defaultRepos` are restored before each test. Why the change?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceStartSpec.scala
```diff
@@ -141,8 +163,11 @@ object ServiceStartSpec {
     CosmosClient.submit(request)
   }
 
-  def addUniversePackage(packageName: String): Response = {
-    val addRequest = rpc.v1.model.UniverseAddRequest(packageName, packageVersion = None)
+  def addUniversePackage(
+    packageName: String,
+    packageVersion: Option[universe.v3.model.Version] = None
+  ): Response = {
```
cruhland: I'm not sure how this function is used, but would it make sense to have it call `CosmosClient.callEndpoint()` instead of `CosmosClient.submit()`?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceStartSpec.scala
```diff
@@ -129,6 +129,28 @@ final class ServiceStartSpec extends FreeSpec with InstallQueueFixture with Befo
...
+      assertResult(Status.Accepted)(addResponse.status)
+
+      awaitEmptyInstallQueue()
+
+      val startResponse = startService(packageName)
+      assertResult(Status.Ok)(startResponse.status)
+      assertResult(Some(rpc.MediaTypes.ServiceStartResponse.show))(startResponse.contentType)
+
+      val typedResponse = decode[rpc.v1.model.ServiceStartResponse](startResponse.contentString)
+      assertResult(packageName)(typedResponse.packageName)
```
cruhland: We should probably check the package version here, like the other test does.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandlerSpec.scala
```diff
@@ -54,7 +56,11 @@ final class UninstallHandlerSpec extends FreeSpec {
       assertResult(Status.Ok, s"install failed: $installRequest1")(installResponse1.status)
 
       val appId2 = AppId(UUID.randomUUID().toString)
-      val installRequest2 = InstallRequest("helloworld", appId = Some(appId2))
+      val installRequest2 = InstallRequest(
+        "helloworld",
+        packageVersion = Some(PackageDetailsVersion("0.4.1")),
```
cruhland: If you change the name of the v4 packages then you could avoid using an explicit version here.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
cruhland: Yeah, I'm also confused by this. The previous code restored the `originalRepositories` after each test. Now `defaultRepos` are restored before each test. Why the change?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
```diff
@@ -67,7 +67,7 @@ final class PackageRepositorySpec
         assertAdd(SourceMesosphere +: defaultRepos, SourceMesosphere, Some(0)) // prepend
       }
       "insert" in {
-        assertAdd(defaultRepos.head :: SourceExample :: defaultRepos(1) :: Nil, SourceExample, Some(1)) // insert
```
cruhland: Could also use `defaultRepos.take(1) ++ (SourceExample :: defaultRepos.drop(1))` which will be safe even if `defaultRepos` is empty.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/UniverseClientSpec.scala
```diff
@@ -63,18 +64,25 @@ final class UniverseClientSpec extends FreeSpec {
...
     "should be able to fetch" - {
 
+      "1.10 json" in {
+        val version = DcosReleaseVersionParser.parseUnsafe("1.10")
+        val baseRepoUri = "https://downloads.mesosphere.com/universe/ebdcd8b7522e37f33184d343ae2a02ad0b63903b/repo"
+        val repoFilename = "repo-up-to-1.10.json"
+        val repository = PackageRepository("repo", baseRepoUri / repoFilename)
+        val repo = Await.result(universeClient(repository, version))
+        getVersions(repo, "helloworld") shouldBe
+          List(Version("0.4.0"), Version("0.4.1"))
```
cruhland: Could you update the helper methods and values so that this test has a similar structure to the tests for other versions? It will make it easier to understand that the tests are doing similar things.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -104,7 +104,9 @@ final class DefaultUniverseClient(
           throw RepositoryUriSyntax(repository, t)
       } flatMap { case conn: HttpURLConnection =>
         // Set headers on request
-        conn.setRequestProperty("Accept", MediaTypes.UniverseV3Repository.show)
+        conn.setRequestProperty(
+          "Accept", s"${MediaTypes.UniverseV4Repository.show}, ${MediaTypes.UniverseV3Repository.show}"
```
cruhland: We should use the `CompoundMediaType` here, it was designed for this purpose.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -164,11 +166,17 @@ final class DefaultUniverseClient(
...
+    val v3Compatible = contentType.isCompatibleWith(MediaTypes.UniverseV3Repository)
+
     // Decode the packages
-    val repo = if (contentType.isCompatibleWith(MediaTypes.UniverseV3Repository)) {
-      val v3Scope = decodeScope.scope("v3")
-      v3Scope.counter("count").incr()
-      Stat.time(v3Scope.stat("histogram")) {
+    val repo = if (
+      v3Compatible ||
+        contentType.isCompatibleWith(MediaTypes.UniverseV4Repository)
```
cruhland: Let's split this into two separate `if` clauses for readability. I know that both v3 and v4 repos still get decoded as `universe.v3.model.Repository` (why did we decide to do it that way again?), but in this case I think a little repetition is helpful to show that we're making a decision based on the version.

cosmos-test-common/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
```diff
@@ -22,8 +22,12 @@ final class UniverseConversionsSpec extends FreeSpec with Matchers with TableDri
...
       MaximalV3ModelPackageDefinitionV3 -> expectV3(MaximalV2ModelPackageDetails),
       MinimalV3ModelPackageDefinitionV3 -> expectV3(MinimalV2ModelPackageDetails),
-      MaximalV4ModelPackageDefinitionV4 -> expectV4(MaximalV2ModelPackageDetails),
-      MinimalV4ModelPackageDefinitionV4 -> expectV4(MinimalV2ModelPackageDetails)
+      MaximalV4ModelPackageDefinitionV4 -> expectV4(
+        MaximalV2ModelPackageDetails.copy(name = MaximalV4ModelPackageDefinitionV4.name)
+      ),
+      MinimalV4ModelPackageDefinitionV4 -> expectV4(
+        MinimalV2ModelPackageDetails.copy(name = MinimalV4ModelPackageDefinitionV4.name)
+      )
```
cruhland: Is this related to the name change?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
larioj: Maybe I could delete this one instead and use the one in CosmosClient. What's better? Should we have methods in this file or in CosmosClient?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
larioj: See reply above.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ListVersionsSpec.scala
larioj: I'll pass the parameter by name.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ListVersionsSpec.scala
larioj: Yeah, I guess I could. But I like testing the responses in Json format.

larioj: When it's not too taxing.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
larioj: Yeah, this test was out of place.

larioj: Personally I like the structure of the test. It makes it easy to add new examples.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
larioj: So you do think that the ItUtil file is a good place for this sorts of methods. :) cool

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageInstallIntegrationSpec.scala
larioj: Setting this to true tells the test to specify the version.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
larioj: I got tired of having to look for the test in case there was another further down. It's much better for the tests to all be at the top.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
larioj: Yeah, this was actually buggy. It caused the helloworld repo to be placed at the end of the list regardless of its original position. It took a while to figure that out.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositoryIntegrationSpec.scala
larioj: This is actually a little bit more conceptually sound. The tests are saying before each test replace the repositories with my view of what the repositories should be rather than what the ensemble thinks it should be. But after all the tests, replace it to what it was before the tests run. Granted this does not matter because the cosmos we are running the tests against has it's own zk, but if we ever change that this would be better.

cosmos-test-common/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
larioj: Yes

cosmos-test-common/src/main/scala/com/mesosphere/universe/test/TestingPackages.scala
larioj: I added a TableDriven tests that uses all these packages for package add. Since Maximal and MaximalV4 had the same name, the new iteration of the test would start before the uninstall had finished which caused the test to fail. Giving the two packages different coordinates prevents the race.

cosmos-test-common/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
larioj: I don't know if there is a better way of doing this. We have a lot of structures which are almost the same. It seems a waste to have to write out the whole thing again.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
larioj: Okay!

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
larioj: Oh I did not know that existed.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
larioj: deleted the other method

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
larioj: deleted other method

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceStartSpec.scala
larioj: Done

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandlerSpec.scala
larioj: It's elucidating to have two repositories that share packages.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
larioj: Done

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -0,0 +1,326 @@
...
+import org.scalatest.prop.TableFor2
+
+object ItObjects {
+
+  val helloWorldPackageSource1: String = {
+    DefaultRepositories().getOrThrow(1).uri.toString
+  }
+
+  val helloWorldPackageSource2: String = {
+    DefaultRepositories().getOrThrow(2).uri.toString
```
cruhland: Using an `Int` to index into the repository list is brittle -- when we update the list in the future, it could take a while for us to discover that this code is not returning the intended repo. Could you add a method to `DefaultRepositories` (or an appropriate test utility) that retrieves a repo by its name, or throws an exception if it's not present?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -0,0 +1,326 @@
...
+      |    "docker": {
+      |      "image": "python:3",
+      |      "network": "HOST"
+      |    }
+      |  }
+      |}
+    """.stripMargin
+  }
+
+  // scalastyle:off
```
cruhland: Could you include the ID(s) of the check(s) that you're disabling here? That way we can still catch other warnings.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -0,0 +1,326 @@
...
+      |    "upgradesFrom": [
+      |        "0.4.0"
+      |    ],
+      |    "version": "0.4.1",
+      |    "website": "https://github.com/mesosphere/dcos-helloworld"
+      |}
+      |
+    """.stripMargin
+  }
+  // scalastyle:on
```
cruhland: You'll have to update this too.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
cruhland: Might want to put the repo name into the variables as well, instead of using the index.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -0,0 +1,326 @@
...
+        "postInstallNotes",
+        "description",
+        "packagingVersion",
+        "tags",
+        "selected",
+        "framework",
+        "maintainer",
+        "version",
+        "preInstallNotes"
+      ).contains(k)
```
cruhland: It turns out that `Set[A]` extends `A => Boolean`, so it can be used directly as a predicate function! I'd recommend factoring out the `Set` into a `val` outside of this `def` though, and then referencing it:

```scala
packageDefinition.asObject.get.filterKeys(MetadataFields)
```

cruhland: I should also ask though, why are you filtering a JSON to produce another JSON? Wouldn't it be easier and less error prone to start with a `PackageDefinition`, convert it to a `Metadata`, and then, if needed, convert it to JSON?

I have the same question for most of the fields in this object -- are you keeping them as JSON so you can directly compare them to the response body from Cosmos? It would likely make tests simpler if you parsed the JSON back into case classes and compared those. Ideally we have unit tests which ensure that every (case class) => JSON => (case class) is the identity function.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
cruhland: Yeah, were you going to move this there?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -174,11 +161,10 @@ private object PackageDescribeSpec extends TableDrivenPropertyChecks {
     preInstallNotes = Some("A sample pre-installation message"),
     postInstallNotes = Some("A sample post-installation message"),
     tags = List("mesosphere", "example", "subcommand"),
-    selected = Some(false),
```
cruhland: Why did this get removed?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageListIntegrationSpec.scala
cruhland: Ok.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
larioj: It's about testing the behavior that the client sees. I don't think we test Json => case class => Json. (I'm fairly sure that broken since nulls are not removed). Also none of our tests are independent of our data structures. When our types change, so do our tests which prevents us from actually noticing breaking changes. There is something fishy about using the infrastructure that we are testing to check the tests.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositoryIntegrationSpec.scala
cruhland: Well, we definitely need to restore the `originalRepositories`, since other tests probably depend on them.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
larioj: The decoding code for zip universe, arbitrarily set selected to false when None. I removed that code, so this test had to change.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandlerSpec.scala
cruhland: Ok

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -308,7 +319,7 @@ final class DefaultUniverseClient(
           // source is universe-server.
           universe.v3.model.Tag(tag).get
       },
-      details.selected.orElse(Some(false)),
```
cruhland: Why did this get removed?

cosmos-test-common/src/main/scala/com/mesosphere/universe/test/TestingPackages.scala
cruhland: Would it be "more correct" to have each iteration wait until the uninstall actually completed before starting the next one? This might even be something we want the uninstall endpoint to do itself. What do you think @jsancio ?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
larioj: This is an arbitrary and unnecessary transformation. This makes it so that we store something different than the input.

cosmos-test-common/src/test/scala/com/mesosphere/universe/bijection/UniverseConversionsSpec.scala
cruhland: All of these pieces of test data could benefit from some refactoring. We could define most of the fields on a single general type, and then use bijections or other conversions to obtain the specific types needed by specific tests. Possibly Shapeless would be useful in cases like this, where we would ideally start with something like a coproduct of almost all fields, and then add the name as the last step before "projecting into" the specific type.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
cruhland: You make a good point about the JSON => case class => JSON test. Could you file a JIRA for creating a unit test suite that catches breaking JSON changes?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
cruhland: But what if the field is expected to have the specific value of `Some(false)` for a reason? Then this test would be how we enforce that behavior. Granted, enforcing it in a unit test would be better, and we should probably have a dedicated test for it with a specific name.

Could you see if you can find the commit/PR that introduced this change? Watch out for commits where the file was moved or renamed.

jsancio: I think the UI/CLI expect this. See the conversion: https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala#L105-L108

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: Let's make sure that this doesn't break our integration tests and eventually our CLI tests.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
jsancio: Yes. I want us to sit down and design how we want our integration tests to work. I would like our next hire to execute this design.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
cruhland: If we do find out that it's required though, let's make sure the reason why is included in the test code, either in the form of a comment or a dedicated test case.

cruhland:
@larioj getting closer! Thanks for pointing out the need for testing the actual JSON we generate.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
larioj: @jsancio Those conversions make sense. But that's for v2 describe. This test is doing a v1 describe https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala#L60

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
larioj: @jsancio @cruhland I looked into this. This line makes it so that when a package comes from a zip file selected is set. However when a package comes from JSON selected does not have to be set. Describing "helloworld" from JSON in v1 describeResponse yields selected not set. v1 Describe on "helloworld" from zip yields selected set. So the describe v1 endpoint is dependent on whether the source is zip or JSON which is clearly wrong. We have two choices, we either drop this transformation, or we make sure v1 describe always returns selected set. I'm fine with either. Either way this change stays, since the decision to return selected set should be made in the same place as in v2 DescribeResponse.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
larioj: Created Jira Issue https://jira.mesosphere.com/browse/DCOS-15541

cosmos-test-common/src/main/scala/com/mesosphere/universe/test/TestingPackages.scala
larioj: This is outside of the scope of this PR.

PR #616, opened 2017-05-04
Service describe

larioj:


cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
```diff
@@ -0,0 +1,116 @@
...
+  private val path: String = "service/describe"
+  private val contentType: String = "application/vnd.dcos.service.describe-request+json;charset=utf-8;version=v1"
+  private val accept: String = "application/vnd.dcos.service.describe-response+json;charset=utf-8;version=v1"
+
+  feature("The service/describe endpoint") {
+    scenario("The user would like to know the upgrades available to a service") {
+      serviceDescribeTest { (content, _, expectedUpgrades, _, _) =>
+        Then("the user should be able to observe the upgrades available to that service")
+        val Right(actualUpgrades) = content.hcursor.get[Json]("upgradesTo")
+        expectedUpgrades.asJson shouldBe actualUpgrades
```
jsancio: Let's standardize on `actual shouldBe expected`. This applies to a few other tests.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
```diff
@@ -0,0 +1,116 @@
...
+  private def serviceDescribeTest(
+    testCode: (Json, universe.v4.model.PackageDefinition, List[String], List[String], Json) => Assertion
+  ): Unit = {
+    forAll(helloWorldPackageDefinitions) {
+      (packageDefinition, expectedUpgrades, expectedDowngrades, resolvedOptions) =>
+        Given("a running service and its appId")
+        val appId = AppId(UUID.randomUUID().toString)
+        val name = packageDefinition.name
+        val version = packageDefinition.version.toString
+        val Right(install) = ItUtil.packageInstall(name, Some(version), appId = Some(appId))
```
jsancio: Okay but it going to be difficult to use this pattern when you test "service/update" because we are not going to allow the user to override appId. I really want us to deprecate `appId` from "package/install". For example, I don't think "service/start" allows you to override the appId.

cosmos-server/src/main/scala/com/mesosphere/cosmos/MultiRepository.scala
```diff
@@ -105,6 +106,18 @@ final class MultiRepository(
     } map { versions => versions.flatten.toList }
   }
 
+  override def downgradesTo(
+    packageDefinition: PackageDefinition)(implicit
```
jsancio: Copy the formatting patterned used above in `upgradesTo`

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceDescribeHandler.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  private def getPackageDefinition(
+    app: MarathonApp)(implicit
+    session: RequestSession
+  ): Future[universe.v4.model.PackageDefinition] = {
+    app.packageDefinition.map(Future.value).getOrElse {
+      val (name, version) =
+        app.packageName.flatMap(name => app.packageVersion.map(name -> _))
+          .getOrElse(throw new RuntimeException(
+            "The name and version of the service were not found in the labels"))
```
jsancio: Let's use a different exception. Maybe `IllegalStateException`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceDescribeHandler.scala
```diff
@@ -0,0 +1,62 @@
...
+      packageCollection
+        .getPackageByPackageVersion(name, Some(version))
+        .map(_._1)
```
jsancio: I am worried that this is going to cause problems similar to the one we saw with "package/list". People are abusing Universe and deleting packages. For Local Universe, unfortunately our current implementation allows for package to be removed.

Maybe for now, we leave this as is but be aware that we may see reports from the field around this RPC.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceDescribeHandler.scala
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  private def getResolvedOptions(
+    packageDefinition: PackageDefinition,
+    serviceOptions: Option[JsonObject]
+  ): Option[JsonObject] = {
+    serviceOptions.map { userSuppliedOptions =>
+      PackageDefinitionRenderer.mergeDefaultAndUserOptions(packageDefinition, Some(userSuppliedOptions))
+    }
+  }
```
jsancio: I know that your design document talks about returning the "union" of these to objects. How about just returning `userSuppliedOptions`? Did we talk about that? I am thinking of returning that only because that is what the user specified during the "package/install" RPC.

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/ServiceDescribeResponse.scala
```diff
@@ -0,0 +1,27 @@
...
+import io.circe.Encoder
+import io.circe.JsonObject
+import io.circe.generic.semiauto.deriveDecoder
+import io.circe.generic.semiauto.deriveEncoder
+
+case class ServiceDescribeResponse(
+  `package`: universe.v4.model.PackageDefinition,
+  upgradesTo: List[universe.v3.model.Version],
+  downgradesTo: List[universe.v3.model.Version],
+  resolvedOptions: Option[JsonObject]
```
larioj: Personally, I lean towards this version. Here we are answering the question "What options were used to run this service?" rather than "What options did I give this service?". I think the first question is more in line with the name of the endpoint. A description of a service should include all of the configuration parameters used, not just the ones that the user provided.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -221,17 +238,78 @@ object ItObjects {
       .asJson
   }
 
-  def helloWorldPackageDefinitionEnvelope(
-    packageDefinition: Json
+  def decodeEncodedPartsOfRenderResponse(renderResponse: Json): Json = {
```
larioj: I agree. I'll try to address this with the Integration Tests Design document.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -221,17 +238,78 @@ object ItObjects {
...
+    decodePackageDefinition _ andThen
+      decodePackageMetadata andThen
+      decodeOptions apply
+      renderResponse
+  }
```
larioj: I'll write myself a note to address this later. This is actually not relevant to this PR.

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/ServiceDescribeRequest.scala
```diff
@@ -0,0 +1,24 @@
...
+case class ServiceDescribeRequest(appId: AppId)
+
+object ServiceDescribeRequest {
+  implicit val encode: Encoder[ServiceDescribeRequest] = deriveEncoder
+  implicit val decode: Decoder[ServiceDescribeRequest] = deriveDecoder
+  implicit val mediaTypedEncoder: MediaTypedEncoder[ServiceDescribeRequest] =
+    MediaTypedEncoder(MediaTypes.ServiceDescribeRequest)
+  implicit val mediaTypedRequestDecoder: MediaTypedRequestDecoder[ServiceDescribeRequest] =
+    MediaTypedRequestDecoder(MediaTypedDecoder(MediaTypes.ServiceDescribeRequest))
+}
```
larioj: Done

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceDescribeHandler.scala
larioj: I agree. We are resorting to this only after we try to get the package definition from the label. We could very well not even do this, and fail if the package definition is not in the label. But I thought this would be more appropriate.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceDescribeHandler.scala
larioj: I'm okay with either, but we need to think about how this component of the result is going to be used. People will be querying this endpoint when update fails to merge options. Will users want the mergedOptions, or will they want just the options they specified.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -38,6 +38,14 @@ trait PackageCollection {
     version: universe.v3.model.Version
   )(implicit session: RequestSession): Future[List[universe.v3.model.Version]]
 
+  /**
+    * Return the versions of packages in this collection that the package with the given `name` and
+    * `version` can be downgrade to.
+    */
+  def downgradesTo(
+    packageDefinition: universe.v4.model.PackageDefinition
+  )(implicit session: RequestSession): Future[List[universe.v3.model.Version]]
```
jsancio: Let's fix the comment to match the signature and semantic.

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/ServiceDescribeRequest.scala
```diff
@@ -0,0 +1,24 @@
...
+
+case class ServiceDescribeRequest(appId: AppId)
+
+object ServiceDescribeRequest {
+  implicit val encode: Encoder[ServiceDescribeRequest] = deriveEncoder
+  implicit val decode: Decoder[ServiceDescribeRequest] = deriveDecoder
+  implicit val mediaTypedEncoder: MediaTypedEncoder[ServiceDescribeRequest] =
+    MediaTypedEncoder(MediaTypes.ServiceDescribeRequest)
+  implicit val mediaTypedRequestDecoder: MediaTypedRequestDecoder[ServiceDescribeRequest] =
+    MediaTypedRequestDecoder(MediaTypedDecoder(MediaTypes.ServiceDescribeRequest))
```
cruhland: I might be misremembering, but don't we have a way to generate an implicit `MediaTypedRequestDecoder[A]` for any `A` that has a `MediaTypedDecoder[A]`? If so you'd only need to define a `MediaTypedDecoder[ServiceDescribeRequest]` here.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+          content,
+          packageDefinition,
+          expectedUpgrades,
+          expectedDowngrades,
+          expectedResolvedOptions,
+          expectedUserProvidedOptions
+        )
+
+        // clean up
+        ItUtil.packageUninstall(name, appId, all = true)
```
cruhland: If an assertion fails in `testCode()` this cleanup code won't be run because `AssertionError` would have been thrown.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+      (ItObjects.helloWorldPackage0, List(), List(), ItObjects.helloWorldResolvedOptions(), Json.obj()),
+      (ItObjects.helloWorldPackage3, List("0.4.1"), List(), ItObjects.helloWorldResolvedOptions(), Json.obj()),
+      (ItObjects.helloWorldPackage4, List(), List("0.4.0"), ItObjects.helloWorldResolvedOptions(), Json.obj())
+    )
+  }
+
+  private def serviceDescribeTest(
+    testCode: (Json, universe.v4.model.PackageDefinition, List[String], List[String], Json, Json) => Assertion
+  ): Unit = {
+    forAll(helloWorldPackageDefinitions) {
```
cruhland: I know that we used tables of packages in older integration tests, but I've slowly changed my mind on their usefulness over time. I'd rather see each test case exercise just *one* package if possible; in general, we should use the minimal amount of information required for the test to be effective. The intent is that integration tests should not be exhaustive, but should instead ensure that if a user scenario is not satisfied, we can detect it from a test failure.

If you'd like to test various combinations of inputs, to ensure that edge cases are covered, that can be done in a unit test. There are some unit tests which cover the logic of entire endpoints (e.g. `PackageAddHandlerSpec`, `PackageSearchHandlerSpec`), but there might be a more specific component (such as a helper method) that could serve as the unit under test.

What do you think?

jsancio: :+1: This sounds like a good policy. If we agree, let's include it in your design doc @larioj .

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/ServiceDescribeRequest.scala
larioj: This fails to compile w/o it. We also have a whole file of this things so I'm pretty sure it's necessary.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
larioj: I think the design document will be a good place to discuss these things and come to an agreement. For now I'll be leaving this as is.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+        install.appId shouldBe appId
+
+        When("the user makes a request to the service/describe endpoint")
+        val response = serviceDescribe(appId.toString)
+        response.status shouldBe Status.Ok withClue response.contentString
+        response.contentType shouldBe Some(accept)
+        val Right(content) = parse(response.contentString)
+
+        // the actual test
+        testCode(
```
cruhland: I've also become less of a fan of this inversion of control (IoC) pattern in tests over time. I've definitely used it in the past, although in more recent tests, I've only used it in limited contexts to remove very repetitive code (see `buildHandler()` in `PackageAddHandlerSpec`), and not to abstract away the setup and teardown of the whole test.

Instead, a direct style is more clear to a newcomer reading the tests. Helper methods can still be used, but they don't result in IoC: they return data instead of passing data to a function taken as an argument. For example, here's what your first test would look like in direct style:

```scala
scenario("The user would like to know the upgrades available to a service") {
  Given("a running service and its appId")
  val packageDefinition = ??? // some appropriate data...
  val appId = testStartService(packageDefinition)

  try {
    When("the user makes a request to the service/describe endpoint")
    val content = testServiceDescribe(appId)

    Then("the user should be able to observe the upgrades available to that service")
    val expectedUpgrades = ??? // some appropriate data...
    val actualUpgrades = content.hcursor.get[Json]("upgradesTo")
    actualUpgrades shouldBe Right(expectedUpgrades.asJson)
  } finally {
    ItUtil.packageUninstall(packageDefinition.name, appId, all = true)
  }
}
```

This format is more repetitive, but someone new to the code can make sense of it more quickly. Also, only the data needed by the test is created; there are no values that need to be ignored using underscores in argument lists. The hope with this format is that it's easier to change if requirements change, since each test is more independent of the others. Currently, since every test case relies on `serviceDescribeTest()`, that method must be complex enough to work for all of them...over time, it will get more and more difficult to make changes to it as the number of tests grows.

What do you think? Let's discuss more in person!

larioj: Yeah, I considered it doing it that way, but it is more repetitive. There is this tradeoff between compactness and readability. And we need to set some guidelines as to what is expected for tests.

jsancio: Yes! I tend to agree with your observations. Repetition or code duplication in tests, specially integration tests, is good thing. Like you mentioned it ideally makes it a direct translation of the specification.

Having said that, there are some very common patterns that we may want to standardize like `withService(pkg) { appId => ... }`

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceDescribeHandler.scala
```diff
@@ -0,0 +1,64 @@
...
+        userProvidedOptions = userProvidedOptions
+      )
+    }
+  }
+
+  private def getPackageDefinition(
+    app: MarathonApp)(implicit
+    session: RequestSession
+  ): Future[universe.v4.model.PackageDefinition] = {
+    app.packageDefinition.map(Future.value).getOrElse {
```
cruhland: Is the `None` case something that could actually happen? I guess it would be for a service that had been started with an old Cosmos? We should probably test that case -- unit test?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -49,7 +57,20 @@ object PackageCollection {
...
+  }
+
+  def downgradesTo(
+    packageDefinition: universe.v4.model.PackageDefinition,
+    possibleDowngrades: List[universe.v4.model.PackageDefinition]
+  ): List[universe.v3.model.Version] = {
+    possibleDowngrades.collect {
+      case possibleDowngrade
+        if packageDefinition.name == possibleDowngrade.name &&
+          packageDefinition.canDowngradeTo(possibleDowngrade.version) => possibleDowngrade.version
```
cruhland: There's a bug on this line! Can you find it? Let's add tests for `downgradesTo()` to `PackageCollectionSpec`.

cruhland: Oh, never mind, the direction of the `canDowngradeTo()` check works differently than `canUpgradeFrom()`. So, no bug, but a test would help prevent a similar mistake in the future.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
cruhland: I talked with @larioj offline, and we decided to leave the code as is for now, and include this discussion in the test design doc.

PR #619, opened 2017-05-10
Add infrastructure for update

larioj:
I need the service update endpoint in marathon in order to implement service update.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/MarathonClient.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/MarathonClient.scala
@@ -23,7 +23,7 @@ class MarathonClient(
     client(post("v2" / "apps" , Json.fromJsonObject(appJson)))
   }
 
-  def updateApp(appId: AppId, appJson: JsonObject)(implicit session: RequestSession): Future[Response] = {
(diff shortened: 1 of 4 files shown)


benclarkwood:
I think I'll rename my method to `modify` since it effectively modifies the app in place, versus this a true "update".

cosmos-server/src/main/scala/com/mesosphere/cosmos/MarathonClient.scala
```diff
@@ -21,6 +23,10 @@ class MarathonClient(
     client(post("v2" / "apps" , Json.fromJsonObject(appJson)))
   }
 
+  def update(appId: AppId, appJson: JsonObject)(implicit session: RequestSession): Future[Response] = {
+    client(put("v2" / "apps" / appId.toUri, Json.fromJsonObject(appJson)))
+  }
```
jsancio: @larioj I take it that `POST` doesn't work if the `id` already exists? cc @jasongilanfarr

larioj: https://mesosphere.github.io/marathon/docs/rest-api.html#example-create-an-app-with-an-already-existing-id

PR #620, opened 2017-05-10
Add Service Update Errors

larioj:
Will be using this errors in cosmos-enterprise for service/update

### DIFF
--- a/cosmos-common/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
@@ -516,4 +516,47 @@ case class ConversionFromPackageToV2DescribeResponse(
     Some(JsonObject.singleton("msg", msg.asJson))
   }
 }
+
+case class OptionsNotStored(
+  msg: String = "The service being updated does not have the user provided " +
+    "options stored. Please retry with options set"
+) extends CosmosError {
+  override val getData: Option[JsonObject] = {
+    Some(JsonObject.singleton("msg", msg.asJson))
+  }
+}
+
+case class AppIdChanged(
+  msg: String = "During a service update the app ID must not change"
+) extends CosmosError {
+  override val getData: Option[JsonObject] = {
+    Some(JsonObject.singleton("msg", msg.asJson))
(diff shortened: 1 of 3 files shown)


cosmos-common/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
```diff
@@ -516,4 +516,47 @@ case class ConversionFromPackageToV2DescribeResponse(
...
+  }
+}
+
+case class ServiceUpdateError(
+  msg: String
+) extends CosmosError {
+  override val getData: Option[JsonObject] = {
+    Some(JsonObject.singleton("msg", msg.asJson))
+  }
+}
```
jsancio: For all of these errors, we should include all the contextual information. E.g. what service id did you try to update? What is the old app id vs the new id? etc.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -408,7 +408,7 @@ object CosmosApp {
     requirement.fold[EndpointHandler[Req, Res]](new NotConfiguredHandler(operationName))(f)
   }
 
-  private def standardEndpoint[Req, Res](
+  def standardEndpoint[Req, Res](
```
jsancio: @cruhland Can you take a look at this? Is this how you intended for Cosmos to get extended?

cruhland: @jsancio yeah, this is fine.

cosmos-common/src/main/scala/com/mesosphere/cosmos/CosmosError.scala
cruhland: @larioj I noticed you changed all the arguments to `String`. While that will be convenient for the producer of the error, any consumers will not be able to extract contextual information, as @jsancio mentioned. We should have explicit fields for each piece of information, that get surfaced in `getData`.

That said, if you want to merge this as is in the interest of time, that's fine with me. But you'll have to update it later...

PR #621, opened 2017-05-10
Don't propagate status codes from Universe

jsancio:


cosmos-server/src/main/scala/com/mesosphere/cosmos/MarathonClient.scala
```diff
@@ -33,7 +33,7 @@ class MarathonClient(
       response.status match {
         case Status.Ok => Some(decodeJsonTo[MarathonAppResponse](response))
         case Status.NotFound => None
-        case s: Status => throw GenericHttpError(HttpMethod.GET, uri, s)
+        case s: Status => throw GenericHttpError(HttpMethod.GET, uri, s, s)
```
cruhland: This will make us respond with the same status as Marathon. Is that what we want?

cosmos-server/src/main/scala/com/mesosphere/cosmos/ServiceClient.scala
```diff
@@ -60,7 +60,7 @@ abstract class ServiceClient(baseUri: Uri) {
       case Status.Ok =>
         Future.value(response)
       case s: Status =>
-        throw new GenericHttpError(method, uri, s)
+        throw GenericHttpError(method, uri, s, s)
```
cruhland: This will make us respond with the same status as the upstream service. Is that what we want?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -133,7 +135,15 @@ final class DefaultUniverseClient(
               throw UnsupportedRedirect(List(repository.uri.scheme.get), loc)
             case x =>
               fetchScope.scope("status").counter(x.toString).incr()
-              throw GenericHttpError("GET", repository.uri, x)
+              /* If we are unable to get the latest Universe we should not forward the status code returned.
+               * We should instead return 500 to the client and include the actuall error in the message.
```
cruhland: `actuall` -> `actual`

cosmos-server/src/main/scala/com/mesosphere/cosmos/MarathonClient.scala
jsancio: Yes. In this case I think we need to forward Marathon error to the client. Specially 403 errors.

cosmos-server/src/main/scala/com/mesosphere/cosmos/ServiceClient.scala
jsancio: Yes. I think this is what we want for all of our Finagle clients for now (Marathon and Mesos).

cruhland:
@jsancio thanks for answering my questions! :ship: it

cosmos-common/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -234,8 +234,8 @@ object Encoders {
           case None =>
             s"Unspecified Content-Encoding Accept-Encoding: $acceptMsg"
         }
-      case GenericHttpError(method, uri, status) =>
-        s"Unexpected down stream http error: ${method.getName} ${uri.toString} ${status.code}"
+      case GenericHttpError(method, uri, clientStatus, status) =>
```
larioj: Are we using status? Or should it be _ ?

PR #622, opened 2017-05-11
Add list deployments method to admin router

larioj:
This is necessary to check that deployments have finished.

### DIFF
--- a/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
+++ b/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
@@ -1,12 +1,17 @@
 package com.mesosphere.cosmos
 
 import _root_.io.circe.JsonObject
+import _root_.io.circe.jawn.parse
+import cats.syntax.either._
 import com.mesosphere.cosmos.http.CosmosRequests
 import com.mesosphere.cosmos.repository.DefaultRepositories
 import com.mesosphere.cosmos.rpc.v1.circe.Decoders._
 import com.mesosphere.cosmos.test.CosmosIntegrationTestClient.CosmosClient
 import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
 import com.mesosphere.universe
+import com.twitter.finagle.http.Status
+import com.twitter.util.Await
+import scala.concurrent.duration._
 
 object ItUtil {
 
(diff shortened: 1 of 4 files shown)


cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
```diff
@@ -81,4 +86,16 @@ object ItUtil {
     )
   }
 
+  def waitForDeployment(adminRouter: AdminRouter)(attempts: Int): Boolean = {
+    Stream.tabulate(attempts) { _ =>
+      Thread.sleep(1.second.toMillis)
+      val response = Await.result {
+        adminRouter.listDeployments()
+      }
+      assert(response.status == Status.Ok || response.status == Status.Conflict)
```
cruhland: I don't think the list deployments endpoint returns a 409 status?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
```diff
@@ -81,4 +86,16 @@ object ItUtil {
     )
   }
 
+  def waitForDeployment(adminRouter: AdminRouter)(attempts: Int): Boolean = {
+    Stream.tabulate(attempts) { _ =>
+      Thread.sleep(1.second.toMillis)
```
jsancio: We should be able to do this async. I think you want something similar to what we do here: https://github.com/dcos/cosmos/blob/master/cosmos-test-common/src/main/scala/com/mesosphere/cosmos/test/TestUtil.scala#L50

Twitter Future don't use more memory when the Future recursion is tail recursive.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
larioj: You are correct. This should not be here.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
larioj: This does not timeout?

larioj: This is just for integration tests. We are going to await it anyways.

PR #623, opened 2017-05-11
Fix some incorrect tests

larioj:
This tests are misbehaving here and in enterprise. The change to ServiceDescribeSpec makes it so that the package is cleaned up if service/describe fails. The change to PackageInstallIntegrationSpec, makes it so that the failure gives a better error message. The test also waits for any deployments to finish before installing. This failing in Enterprise consistently because of a Conflict.


jsancio:
@larioj Can you explain why we are doing this? LGTM otherwise.

larioj:
@jsancio Updated the comment

jsancio:
Thanks! :ship: it!

PR #625, opened 2017-05-12
Translate host suffixes in no_proxy to JVM syntax

cruhland:
This ended up being a bigger change than intended. I decided to write property tests for the `no_proxy` parsing to make sure all the edge cases got covered. But to make it easier to write the tests, I first separated the initialization code in `HttpProxySupport` into two parts:

1. parse the values of the system properties (without side effects)
2. set the values in the global system properties

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpProxySupport.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpProxySupport.scala
@@ -124,10 +124,9 @@ private[cosmos] object HttpProxySupport {
       noProxy.map(np => (HttpProxyNoHosts, translateNoProxy(np)))
   }
 
-  private[cosmos] def parseHostPortProperties(
-    hostProperty: String,
-    portProperty: String
-  )(proxyUri: Uri): List[(String, String)] = {
+  def parseHostPortProperties(hostProperty: String, portProperty: String)(
+    proxyUri: Uri
+  ): List[(String, String)] = {
     Uris.extractHostAndPort(proxyUri)
       .toOption
       .toList
@@ -140,7 +139,7 @@ private[cosmos] object HttpProxySupport {
    * Normalizes the input into the format required by the `http.nonProxyHosts` system property.
    * @see https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html#Proxies
    */
-  private[cosmos] def translateNoProxy(noProxy: String): String = {
+  def translateNoProxy(noProxy: String): String = {
     noProxy
       .split(Array(',', '|'))
       .map(suffix => if (suffix.startsWith(".")) "*" + suffix else suffix)
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpProxySupport.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpProxySupport.scala
@@ -137,7 +137,10 @@ private[cosmos] object HttpProxySupport {
   }
 
   private[cosmos] def translateNoProxy(noProxy: String): String = {
-    noProxy.replaceAllLiterally(",", "|")
+    noProxy
+      .split(',')
+      .map(suffix => if (suffix.startsWith(".")) "*" + suffix else suffix)
+      .mkString("|")
   }
 
   private[this] def setPropertyIfUnset(key: String, value: String): Unit = {
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpProxySupport.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpProxySupport.scala
@@ -136,9 +136,13 @@ private[cosmos] object HttpProxySupport {
       }
   }
 
+  /**
+   * Normalizes the input into the format required by the `http.nonProxyHosts` system property.
+   * @see https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html#Proxies
+   */
   private[cosmos] def translateNoProxy(noProxy: String): String = {
     noProxy
-      .split(',')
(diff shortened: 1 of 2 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpProxySupport.scala
```diff
@@ -102,43 +102,50 @@ private[cosmos] object HttpProxySupport {
...
-        }
+  private[cosmos] def parseHostPortProperties(
+    hostProperty: String,
+    portProperty: String
+  )(proxyUri: Uri): List[(String, String)] = {
+    Uris.extractHostAndPort(proxyUri)
+      .toOption
+      .toList
+      .flatMap { case ConnectionDetails(host, port, _) =>
+        List((hostProperty, host), (portProperty, port.toString))
```
jsancio: I am okay making all of the "functional" method public if you need them for test while keeping all of the side-effecting helper methods `private[this]`.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/HttpProxySupportSpec.scala
```diff
@@ -25,18 +25,29 @@ final class HttpProxySupportSpec extends FreeSpec with BeforeAndAfter {
...
+
+        "respects http.proxyHost and http.proxyPort" in {
           System.setProperty(HttpHost, "host")
           System.setProperty(HttpPort, "port")
-          HttpProxySupport.initHttpProxyProperties(Some(proxyUri), HttpHost, HttpPort)
+
+          val httpProxyUri = "http://localhost:3128"
+          val proxyEnvVariables = ProxyEnvVariables(Some(httpProxyUri), None, None)
+          HttpProxySupport.initProxyConfig(proxyEnvVariables, _ => ())
+
```
jsancio: I see. Maybe you should ignore my `private[this]` vs `private[cosmos]` comment since we have this side-effecting change. In this case `private[cosmos]` is almost public. There are a bunch of unrelated classes in `com.mesosphere.cosmos`.

Actually, if I understand this test correctly, this test looks dangerous. Aren't we overriding the system properties?

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/HttpProxySupportSpec.scala
```diff
@@ -141,6 +141,12 @@ final class HttpProxySupportSpec extends FreeSpec with BeforeAndAfter {
             }
           }
 
+          "should add wildcard in front of '.'-delimited domain suffixes" in {
+            assertResult("*.mesos|*.example.com") {
+              HttpProxySupport.translateNoProxy(".mesos,.example.com")
+            }
+          }
+
```
jsancio: Should we add a test for IPs? E.g `...translateNoProxy(".0.0.100") shouldBe "*.0.0.100"`. Is that even allowed? We can leave it as a known "bug".

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/HttpProxySupportSpec.scala
```diff
@@ -297,4 +291,47 @@ object HttpProxySupportSpec {
...
+      .map { host =>
+        val prefix = host.wildcard match {
+          case NoWildcard => ""
+          case _ => "*."
+        }
+
+        prefix + host.elements.mkString(".")
+      }
+      .mkString("|")
+  }
```
jsancio: Very cool. Having said that explicit tests are easier to read. Maybe we want both. Static tests like we had before and generated tests.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/HttpProxySupportSpec.scala
cruhland: Yeah, these tests were written this way before my change, and it's a unit test so I'm not as concerned about modifying system properties. What do you think should be done differently?

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/HttpProxySupportSpec.scala
cruhland: Good point, I'll bring back the static tests.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/HttpProxySupportSpec.scala
cruhland: I mean, I could try to isolate the side effects to a single method, and update most of the unit tests to use pure helper methods instead. Then we'd still need a side-effecting test, but it would be very restricted.

jsancio: Yeah. We don't need to change the test. I was just pointed it out. I didn't notice it when the test was first introduced.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/HttpProxySupportSpec.scala
cruhland: Oops, missed this comment. I agreed that we don't need to worry about this case until it happens in the wild. Even then, I'm not sure we should support it -- IPs should always be explicit.

larioj:
@cruhland It seems to me that the other PRs are passing with some frequency. How many time have you rerun this?

cruhland:
@larioj probably about ten times by now...

cruhland:
It's not consistently the same test that fails, but it's definitely the same error. Seems really unlikely that it could be related to my PR.

cruhland:
Ninja-merging this PR, as the tests did pass once (see status for commit `ab58211`).

PR #628, opened 2017-05-16
Add RoundTrip abstraction

larioj:
I worked on this over the weekend. RoundTrip is an abstraction that takes care of clean up. You can construct it with a pair of functions: one that goes forward (makes a state change) and one that goes backwards (reverts the state change). The value that the RoundTrip stores is the result of the forward function. You can use map and flatMap to create new RoundTrips. You can execute a RoundTrip by either .get()ing the value inside the RoundTrip or .apply()ing a transformation to the value inside it. This is a generalization of the `withSomeStateChange()` methods that are scattered over the code.

### DIFF
--- a/cosmos-common/src/main/scala/com/mesosphere/util/RoundTrip.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/util/RoundTrip.scala
@@ -1,0 +1,62 @@
+package com.mesosphere.util
+
+class RoundTrip[A](
+  private val withResource: (A => Unit) => Unit
+) {
+
+  def map[B](transform: A => B): RoundTrip[B] = {
+    def withResourceB(bInner: B => Unit): Unit = {
+      withResource { a: A =>
+        bInner(transform(a))
+      }
+    }
+    new RoundTrip[B](withResourceB)
+  }
+
+  def flatMap[B](transform: A => RoundTrip[B]): RoundTrip[B] = {
+    def withResourceB(bInner: B => Unit): Unit = {
+      withResource{ a: A =>
+        transform(a).withResource { b: B =>
+          bInner(b)
+        }
+      }
+    }
+    new RoundTrip[B](withResourceB)
+  }
+
+  def get(): A = {
+    var innerA: Option[A] = None
+    withResource { a: A =>
+      innerA = Some(a)
+    }
+    innerA.get
+  }
+
+  def apply[B](transform: A => B): B = {
+    map(transform).get()
+  }
+
+}
+
+object RoundTrip {
+
+  def apply[A](
+    forward: => A)(
+    backwards: A => Unit
+  ): RoundTrip[A] = {
+    def withResource(aInner: A => Unit): Unit = {
+      val a = forward
+      try {
+        aInner(a)
+      } finally {
+        backwards(a)
+      }
+    }
+    new RoundTrip[A](withResource)
+  }
+
+  def value[A](a: A): RoundTrip[A] = {
+    RoundTrip(a)((_: A) => ())
+  }
+
+}
--- a/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/HttpErrorResponse.scala
+++ b/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/HttpErrorResponse.scala
@@ -8,10 +8,8 @@ final case class HttpErrorResponse(
   status: Status,
   errorResponse: ErrorResponse
 ) extends TestFailedException(
-  _ => Some(
-    s"Status: $status, ErrorResponse: $errorResponse"
-  ),
-  None,
-  _ => 0,
-  None
+  messageFun = _ => Some(s"Status: $status, ErrorResponse: $errorResponse"),
+  cause = None,
+  failedCodeStackDepthFun = _ => 0,
+  payload = None
 )
--- a/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/HttpErrorResponse.scala
+++ b/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/HttpErrorResponse.scala
@@ -4,12 +4,12 @@ import com.mesosphere.cosmos.rpc.v1.model.ErrorResponse
 import com.twitter.finagle.http.Status
 import org.scalatest.exceptions.TestFailedException
 
-case class HttpErrorResponse(
+final case class HttpErrorResponse(
   status: Status,
   errorResponse: ErrorResponse
 ) extends TestFailedException(
   _ => Some(
-    s"${errorResponse.`type`} $status: ${errorResponse.message}"
+    s"Status: $status, ErrorResponse: $errorResponse"
   ),
   None,
   _ => 0,
--- a/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/HttpErrorResponse.scala
+++ b/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/HttpErrorResponse.scala
@@ -1,0 +1,17 @@
+package com.mesosphere.cosmos
+
+import com.mesosphere.cosmos.rpc.v1.model.ErrorResponse
+import com.twitter.finagle.http.Status
+import org.scalatest.exceptions.TestFailedException
+
+case class HttpErrorResponse(
+  status: Status,
+  errorResponse: ErrorResponse
+) extends TestFailedException(
+  _ => Some(
+    s"${errorResponse.`type`} $status: ${errorResponse.message}"
+  ),
+  None,
+  _ => 0,
+  None
+)
--- a/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
+++ b/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
@@ -5,6 +5,8 @@ import com.mesosphere.cosmos.rpc.v1.model.ErrorResponse
 import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
 import com.mesosphere.universe
 import com.twitter.bijection.Conversion
+import io.circe.Json
+import io.circe.jawn.parse
 import java.util.UUID
 
 object ItOps {
@@ -22,8 +24,14 @@ object ItOps {
   implicit final class ItStringOps(val string: String) extends AnyVal {
     def version: universe.v3.model.Version =
       universe.v3.model.Version(string)
+
     def detailsVersion: universe.v2.model.PackageDetailsVersion =
       universe.v2.model.PackageDetailsVersion(string)
+
+    def json: Json = {
+      val Right(result) = parse(string)
+      result
+    }
   }
 
 }
--- a/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
+++ b/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
@@ -8,20 +8,11 @@ import com.twitter.bijection.Conversion
 import io.circe.Json
 import io.circe.jawn.parse
 import java.util.UUID
-import scala.language.implicitConversions
 
 object ItOps {
 
   implicit def cosmosErrorToErrorResponse[E <: CosmosError]: Conversion[E, ErrorResponse] = {
     Conversion.fromFunction(_.exception.errorResponse)
-  }
-
-  implicit def stringToVersion(s: String): universe.v3.model.Version = {
-    universe.v3.model.Version(s)
-  }
-
-  implicit def stringToDetailsVersion(s: String): universe.v2.model.PackageDetailsVersion = {
-    universe.v2.model.PackageDetailsVersion(s)
   }
 
   implicit val uuidToAppId: Conversion[UUID, AppId] = {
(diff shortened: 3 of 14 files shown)


cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/DefaultInstallerSpec.scala
```diff
@@ -1,20 +1,32 @@
...
 final class DefaultInstallerSpec extends FreeSpec with Matchers with PropertyChecks {
-  "Test that installing new package succeeds" in TestUtil.withObjectStorage { tempObjectStorage =>
-    TestUtil.withObjectStorage { objectStorage =>
+
+  private def storagePair: RoundTrip[(ObjectStorage, ObjectStorage)] = {
+    for {
+      temp <- TestUtil.objectStorage
+      actual <- TestUtil.objectStorage
+    } yield {
+      temp -> actual
```
cruhland: Style nitpick: let's not use `->` to create pairs if they don't represent key-value mappings.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/DefaultInstallerSpec.scala
```diff
@@ -1,20 +1,32 @@
...
+    }
+  }
+
+  "Test that installing new package succeeds" in
+    storagePair { case (tempObjectStorage, objectStorage) =>
```
cruhland: This creates a `RoundTrip` instance but never executes it, correct? It looks like none of the test updates below do that. To confirm, try breaking some of the code being tested; the test should still succeed.

Let's think about how we can prevent `RoundTrip` from accidentally disabling tests.

cosmos-common/src/main/scala/com/mesosphere/util/RoundTrip.scala
```diff
@@ -0,0 +1,62 @@
+package com.mesosphere.util
+
+class RoundTrip[A](
+  private val withResource: (A => Unit) => Unit
```
cruhland: This is great! We should be able to simplify the representation somewhat. Want to discuss offline?

Also, could you write some unit tests to ensure this is behaving the way we expect?

cosmos-common/src/main/scala/com/mesosphere/util/RoundTrip.scala
```diff
@@ -0,0 +1,62 @@
...
+    new RoundTrip[B](withResourceB)
+  }
+
+  def get(): A = {
+    var innerA: Option[A] = None
+    withResource { a: A =>
+      innerA = Some(a)
+    }
+    innerA.get
+  }
```
jsancio: I recommend figuring out a way to merge `get` and `withResource`. It would be nice if the only side-effect operations are `forward` and `backwards`.

jsancio: This API would be nice:
```scala
val resource = RoundTrip(open("/tmp/file"), close(_)).map(decode[Config](_)).run
```

with:

```scala
def apply(forward: => A, backward: A => Unit): RoundTrip[A]
def apply(a: A): RoundTrip[A] // useful for lifting and implementing map in terms of flatMap

def flatMap(f: A => RoundTrip[B]): RoundTrip[B]

def run: A
```

Even this API feels strange. What happens if the user does:
```scala
val file = RoundTrip(open(...), close(_)).run
// Or with your API
val file = RoundTrip(open(...), close(_)).get
```
In both cases `file` is closed.

cruhland: Yes, if you don't perform any `map` or `flatMap` operations the return value is unusable. I think this would be okay to use in test code, as long as we're aware of that pitfall. The [robust alternative](https://github.com/fpinscala/fpinscala/wiki/Chapter-14:-Local-effects-and-mutable-state#notes) is more complicated.

jsancio: If we want this only for tests then let's move it to the integration or test scope.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/DefaultInstallerSpec.scala
cruhland: Discussed offline, this is fine -- I misunderstood what was happening here. We decided to adopt a naming convention to make this use case more obvious.

cruhland:
@larioj waiting for an update with comments and the other changes we discussed offline.

cosmos-common/src/main/scala/com/mesosphere/util/RoundTrip.scala
larioj: @jsancio The reason why the API is

```scala
def apply(forward: => A)(backwards: A => Unit): RoundTrip[A]
```

is so that users are able to do something like:

```scala
val install: RoundTrip[Response] = RoundTrip {
  val request = CosmosRequests.installRequest(
    InstallRequest("helloworld")
  )
  CosmosClient.submit(request)
} { response =>
  ItUtil.unistall(respose.name)
}
```
The equivalent for the comma separated argument list is a little bit more awkward.

cosmos-common/src/main/scala/com/mesosphere/util/RoundTrip.scala
```diff
@@ -0,0 +1,62 @@
+package com.mesosphere.util
+
+class RoundTrip[A](
```
larioj: I meant to make this constructor private.

jsancio:
@larioj What if we do this:
```scala
def withResource[T <: AutoCloseable, R](closeable: T)(handler: T => R): R = {
  try {
    handler(closeable)
  } finally closeable.close()
}

// And for As that are not AutoClosesable...
def makeCloseable(init: => A, close: A => Unit): Closing[A] = ???

trait Closing[A] extends AutoCloseable {
  val resource: A
}
```

You can use them together in:
```scala
val managed = withResource(makeCloseable(init, close)) { closing =>
  // use closing.resource
}
```

larioj:
I worked on this over the weekend: There are a few ideas here that I wanted to try out.

1. RoundTrip which is discussed in the description.
2. Reduction of noise.
    * I added an ItOps class with implicit conversions from string to some other types like Version, and Option. This is quite nice. Since it let's us inline a lot of data, which reduces the noise.
3. A new method that instead of returning either throws an HttpErrorResponse.
    * The callEndpoint methods return eithers that then get pattern matched away, which defeats their purpose and makes a lot of boilerplate and noise. It also makes test failures hard to read.
    * Throwing an HttpResponse allows us to use the built in scalastyle functionality like intercept which makes testing much more readable and concise.

With these changes I found myself writing more robust tests, that surprisingly ended up being independent of the state of the cluster! The size of the tests themselves were significantly reduced. The reason the diff has a lot more code added is because 1) Tests in RoundTripSpec.scala 166 lines. 2) Duplicated functionality in Requests.scala 141 lines 3) Duplicated tests in PackageRepositorySpec.scala ~80 lines.

I'd like to discuss the things that I've learned from this with you guys @jsancio @cruhland @takirala

cosmos-test-common/src/main/scala/com/mesosphere/cosmos/util/RoundTrip.scala
```diff
@@ -0,0 +1,98 @@
...
+    roundTrips.foldRight(RoundTrip.value(List(): List[A])) { (aRt, listRt) =>
+      aRt.flatMap { a =>
+        listRt.map { list =>
+          a :: list
+        }
+      }
+    }
+  }
+
+  def value[A](a: => A): RoundTrip[A] = {
```
larioj: I'm inclined to rename this method to `pure` since it takes a code block rather than a value. There is a good reason for these semantics. Let me know if you'd like to talk.

cosmos-test-common/src/main/scala/com/mesosphere/cosmos/util/RoundTrip.scala
```diff
@@ -0,0 +1,98 @@
...
+      } finally {
+        backwards(a)
+        ()
+      }
+    }
+    new RoundTrip[A](withResource)
+  }
+
+  // scalastyle:off method.name
+  implicit class RoundTripHListOps[B <: HList](self: => RoundTrip[B]) {
```
larioj: This methods help to avoid unnecessary nesting. It was mostly just me having fun with scala. They aren't actually used in the tests.

larioj: It allows the user to write

```scala
(withA &: withB &: withC) { case a :: b :: c :: HNil =>
    ???
}
```

rather than the nested version.

larioj: I wanted to avoid the use of operators, but this functionality requires the right associativity which is only available to operators ending in `:`

jsancio:
> I'd like to discuss the things that I've learned from this with you guys @jsancio @cruhland @takirala 

Please schedule a meeting so we can discuss this.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  implicit def stringToDetailsVersion(s: String): universe.v2.model.PackageDetailsVersion = {
+    universe.v2.model.PackageDetailsVersion(s)
+  }
+
+  implicit def uuidToAppId(
+    uuid: UUID
+  ): AppId = {
+    AppId(uuid.toString)
+  }
```
jsancio: Regarding this converters. The recommended way of doing this in Scala is to use typeclasses like:
```scala
trait ConversionOps[A] {
  def to[B](implicit converter: Converter[A, B]): B
}

trait Converter[A, B] {
  def converter(value: A): B
}
```
This is what twitter bijection library does. Can we use that instead of these implicit converters?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
```diff
@@ -0,0 +1,38 @@
...
+import java.util.UUID
+import scala.language.implicitConversions
+
+object ItOps {
+
+  implicit def cosmosErrorToErrorResponse[E <: CosmosError]: Conversion[E, ErrorResponse] = {
+    Conversion.fromFunction(_.exception.errorResponse)
+  }
+
+  implicit def stringToVersion(s: String): universe.v3.model.Version = {
```
larioj: I insist on this conversion. It reduces boilerplate quite significantly.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
```diff
@@ -0,0 +1,38 @@
...
+    universe.v3.model.Version(s)
+  }
+
+  implicit def stringToDetailsVersion(s: String): universe.v2.model.PackageDetailsVersion = {
```
larioj: I also insist on this conversion.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/HttpErrorResponse.scala
```diff
@@ -0,0 +1,17 @@
...
+final case class HttpErrorResponse(
+  status: Status,
+  errorResponse: ErrorResponse
+) extends TestFailedException(
+  _ => Some(
+    s"Status: $status, ErrorResponse: $errorResponse"
+  ),
+  None,
+  _ => 0,
+  None
```
cruhland: Could you use named arguments here for clarity?

larioj: Sure

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
jsancio: Why? Let's explore more options. How about `"1.1.1".version` and `List(1,2,3).map(_.version)`? You can do this with:

```scala
package object ... {
  implicit final class StringOps(val string: String) extends AnyVal {
    def version: universe.v3.model.Version = ???
  }
}
```

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
jsancio: https://github.com/dcos/cosmos/pull/628/files#r130928000

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItOps.scala
```diff
@@ -0,0 +1,38 @@
...
+  implicit val uuidToAppId: Conversion[UUID, AppId] = {
+    Conversion.fromFunction { id =>
+      AppId(id.toString)
+    }
+  }
+
+  implicit def stringToJson(s: String): Json = {
+    val Right(res) = parse(s)
+    res
+  }
```
jsancio: If you implement my suggestions above, you can change this to `"{}".json`.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
```diff
@@ -1,137 +1,40 @@
...
-
-class ServiceDescribeSpec
-  extends FeatureSpec
-    with GivenWhenThen
-    with Matchers
-    with TableDrivenPropertyChecks {
-
-  import ServiceDescribeSpec._
 
+class ServiceDescribeSpec extends FeatureSpec with Matchers {
```
cruhland: `final`

PR #629, opened 2017-05-18
Fix encoding for strings

jsancio:


cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -23,13 +23,21 @@ import io.circe.jawn.parse
...
 
 object PackageDefinitionRenderer {
-  private[this] final val MustacheFactory = new DefaultMustacheFactory()
+  private[this] final val MustacheFactory = new DefaultMustacheFactory {
+    override def encode(value: String, writer: Writer): Unit = {
+      val string = value.asJson.noSpaces
+      // Remove the start and end quote. This is okay because JSON string's length must be >= 2.
+      writer.write(string.slice(1, string.length - 1))
+    }
+  }
```
jsancio: Override the default encoder to use an JSON encoder instead of an HTML encoder.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/render/PackageDefinitionRendererSpec.scala
```diff
@@ -426,6 +426,50 @@ class PackageDefinitionRendererSpec extends FreeSpec with Matchers with TableDri
...
+          Map(
+            ("string", "\n\'\"\\\r\t\b\f".asJson),
+            ("int", 42.asJson),
+            ("double", 42.1.asJson),
+            ("boolean", Json.False)
+          )
+        )
+      )
+    }
+  }
```
jsancio: Check it out @gabrielhartmann and @nlsun . This test verifies that we don't use the HTML encoder and instead use a JSON encoder for strings.

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -133,35 +167,19 @@ object PackageDefinitionRenderer {
...
       jsonBoolean = identity,
       jsonNumber = n => n.toInt.getOrElse(n.toDouble),
-      jsonString = identity,
+      jsonString = value => {
+        /* Encode the string using a JSON string encoding and remove the beginning and ending ".
+         * The slicing operation always succeeds because the small JSON string is "".
+         */
+        val string = value.asJson.noSpaces
+        string.slice(1, string.length - 1)
+      },
```
jsancio: This is the actual fix.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/render/PackageDefinitionRendererSpec.scala
```diff
@@ -426,6 +426,49 @@ class PackageDefinitionRendererSpec extends FreeSpec with Matchers with TableDri
...
+      PackageDefinitionRenderer.renderTemplate(
+        template,
+        context
+      ) shouldBe Right(
+        JsonObject.fromMap(
+          Map(
+            ("string", "\n\'\"\\\r\t\b\f".asJson),
+            ("int", 42.asJson),
+            ("double", 42.1.asJson),
+            ("boolean", Json.False)
```
cruhland: Is this a case where we should use a string literal for the expected result, for maximum accuracy, and clarity? That way we know that the result is byte-for-byte identical with what we expect.

jsancio: Maybe I don't understand the question but the "render" code generates a `JsonObject` not a `String`. So in essences we are testing that the JSON object generated by the render code matches what we expect.

cruhland: Oops, I missed that! Ok, this is good to go then.

jsancio:

PR #631, opened 2017-05-20
Omit user config options from /service/describe response

cruhland:
Addresses [DCOS_OSS-1112](https://jira.mesosphere.com/browse/DCOS_OSS-1112).

The encoder for the response body is modified to drop the user config fields. In order to support changing this behavior in other codebases, a `Handlers` class containing a field for each endpoint handler is introduced.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -83,12 +83,14 @@ lazy val integrationTests = project.in(file("cosmos-integration-tests"))
       (oneJar in server).value,
       (streams in runMain).value
     ),
-    // Uses (compile in Compile) instead of (compile in IntegrationTest), the default
+    // Uses (compile in Compile) in addition to (compile in IntegrationTest), the default
     definedTests in IntegrationTest := {
       val frameworkMap = (loadedTestFrameworks in IntegrationTest).value
-      val analysis = (compile in Compile).value
+      val compileAnalysis = (compile in Compile).value
+      val itAnalysis = (compile in IntegrationTest).value
       val s = (streams in IntegrationTest).value
-      Tests.discover(frameworkMap.values.toList, analysis, s.log)._1
+      Tests.discover(frameworkMap.values.toList, compileAnalysis, s.log)._1 ++
+        Tests.discover(frameworkMap.values.toList, itAnalysis, s.log)._1
     }
   )
   .dependsOn(
--- a/cosmos-integration-tests/src/it/scala/com/mesosphere/cosmos/NonSharedServiceDescribeSpec.scala
+++ b/cosmos-integration-tests/src/it/scala/com/mesosphere/cosmos/NonSharedServiceDescribeSpec.scala
@@ -1,0 +1,30 @@
+package com.mesosphere.cosmos
+
+import _root_.io.circe.Json
+import org.scalatest.FeatureSpec
+import org.scalatest.GivenWhenThen
+import org.scalatest.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks
+
+final class NonSharedServiceDescribeSpec
+  extends FeatureSpec
+    with GivenWhenThen
+    with Matchers
+    with TableDrivenPropertyChecks {
+
+  scenario("The user cannot retrieve the options used to run a service") {
+    ServiceDescribeSpec.serviceDescribeTest { (content, _, _, _) =>
+      Then("the user should not be able to observe the options used to run that service")
+      val actualResolvedOptions = content.hcursor.get[Json]("resolvedOptions")
+      assert(actualResolvedOptions.isLeft)
+    }
+  }
+  scenario("The user cannot retrieve the options they provided to run a service") {
+    ServiceDescribeSpec.serviceDescribeTest { (content, _, _, _) =>
+      Then("the user should not be able to observe the options they provided to run a service")
+      val actualUserProvidedOptions = content.hcursor.get[Json]("userProvidedOptions")
+      assert(actualUserProvidedOptions.isLeft)
+    }
+  }
+
+}
--- a/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
+++ b/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
@@ -43,26 +43,12 @@ class ServiceDescribeSpec
         actualDowngrades shouldBe Right(expectedDowngrades.asJson)
       }
     }
-    scenario("The user cannot know the options used to run a service") {
-      serviceDescribeTest { (content, _, _, _) =>
-        Then("the user should not be able to observe the options used to run that service")
-        val actualResolvedOptions = content.hcursor.get[Json]("resolvedOptions")
-        assert(actualResolvedOptions.isLeft)
-      }
-    }
     scenario("The user would like to know the package definition used to run a service") {
       serviceDescribeTest { (content, packageDefinition, _, _) =>
         Then("the user should be able to observe the package definition used to run that service")
         val expectedDefinition = ItObjects.dropNullKeys(packageDefinition.asJson)
         val actualDefinition = content.hcursor.get[Json]("package").map(ItObjects.dropNullKeys)
         actualDefinition shouldBe Right(expectedDefinition)
-      }
-    }
-    scenario("The user cannot know the options they provided to run a service") {
-      serviceDescribeTest { (content, _, _, _) =>
-        Then("the user should not be able to observe the options they provided to run a service")
-        val actualUserProvidedOptions = content.hcursor.get[Json]("userProvidedOptions")
-        assert(actualUserProvidedOptions.isLeft)
       }
     }
   }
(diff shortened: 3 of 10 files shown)


cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
```diff
@@ -18,77 +18,84 @@ import org.scalatest.FeatureSpec
...
         actualDowngrades shouldBe Right(expectedDowngrades.asJson)
       }
     }
-    scenario("The user would like to know the options used to run a service") {
-      serviceDescribeTest { (content, _, _, _, expectedResolvedOptions, _) =>
-        Then("the user should be able to observe the options used to run that service")
-        val actualResolvedOptions = content.hcursor.get[Json]("resolvedOptions")
-        actualResolvedOptions shouldBe Right(expectedResolvedOptions)
-      }
-    }
```
cruhland: I had to remove this test (and another below) because it fails for cosmos-enterprise. Not sure what we want to do here -- should we move them to a separate class instead?

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -129,23 +130,9 @@ with Logging {
     )
   }
 
-  // scalastyle:off method.length
-  protected final def buildEndpoints(components: Components): Endpoint[Json] = {
+  protected final def buildHandlers(components: Components): Handlers = {
```
cruhland: This allows a handler to be reused for an endpoint without having to construct it again.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -402,6 +389,62 @@ object CosmosApp {
...
+    val packageRepositoryAdd: EndpointHandler[rpc.v1.model.PackageRepositoryAddRequest, rpc.v1.model.PackageRepositoryAddResponse],
+    val packageRepositoryDelete: EndpointHandler[rpc.v1.model.PackageRepositoryDeleteRequest, rpc.v1.model.PackageRepositoryDeleteResponse],
+    val packageRepositoryList: EndpointHandler[rpc.v1.model.PackageRepositoryListRequest, rpc.v1.model.PackageRepositoryListResponse],
+    val packageSearch: EndpointHandler[rpc.v1.model.SearchRequest, rpc.v1.model.SearchResponse],
+    val packageUninstall: EndpointHandler[rpc.v1.model.UninstallRequest, rpc.v1.model.UninstallResponse],
+    val serviceDescribe: EndpointHandler[rpc.v1.model.ServiceDescribeRequest, rpc.v1.model.ServiceDescribeResponse],
+    val serviceStart: EndpointHandler[rpc.v1.model.ServiceStartRequest, rpc.v1.model.ServiceStartResponse]
+  )
+
+  final case class Endpoints(
```
cruhland: We could consider using Shapeless to get rid of the boilerplate repeated fields in `Handlers` and `Endpoints`.

larioj:
@cruhland Do the changes to cosmos.scala break cosmos-enterprise?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
larioj: Why does it fail in cosmos-enterprise?

larioj: I should succeed for enterprise.

cruhland: Oops, I meant that they fail for open cosmos. While developing this feature I had changed these tests to check that the options were *not* included in responses, which worked great for open cosmos, but failed when the same test was run for cosmos-enterprise (because we run all open cosmos tests against cosmos-enterprise). So I was forced to remove these tests, instead of changing them to match the behavior of open cosmos.

The intent of my question above is, should we have separate tests for the behavior of open cosmos, that are *not* run against cosmos-enterprise?

larioj: That's probably a good idea. We're likely gonna run into this scenario again so might as well have an established way to handle this types of tests.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
jsancio: `protected` + `final` feels like a strange modifier combination, specially since this method is pure. Okay for now. but maybe this should just be public.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -155,49 +142,49 @@ with Logging {
...
+      serviceDescribe = new ServiceDescribeHandler(adminRouter, repositories),
+      serviceStart = serviceStartHandler
+    )
+  }
 
-    // Service endpoints
-    val serviceStart = standardEndpoint("service" :: "start", serviceStartHandler)
-    val serviceDescribe = standardEndpoint("service" :: "describe", serviceDescribeHandler)
+  protected final def buildEndpoints(handlers: Handlers): Endpoints = {
+    import handlers._
```
jsancio: Same here. `protected` + `final` for pure function doesn't fell right. `protected` is an OO thing.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ServiceDescribeSpec.scala
cruhland: Separate test added

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
cruhland: Done

cosmos-integration-tests/src/it/scala/com/mesosphere/cosmos/NonSharedServiceDescribeSpec.scala
```diff
@@ -0,0 +1,30 @@
+package com.mesosphere.cosmos
+
+import _root_.io.circe.Json
+import org.scalatest.FeatureSpec
+import org.scalatest.GivenWhenThen
+import org.scalatest.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks
+
+final class NonSharedServiceDescribeSpec
```
cruhland: Looking for alternative name proposals!

build.sbt
```diff
@@ -83,12 +83,14 @@ lazy val integrationTests = project.in(file("cosmos-integration-tests"))
...
+    // Uses (compile in Compile) in addition to (compile in IntegrationTest), the default
     definedTests in IntegrationTest := {
       val frameworkMap = (loadedTestFrameworks in IntegrationTest).value
-      val analysis = (compile in Compile).value
+      val compileAnalysis = (compile in Compile).value
+      val itAnalysis = (compile in IntegrationTest).value
       val s = (streams in IntegrationTest).value
-      Tests.discover(frameworkMap.values.toList, analysis, s.log)._1
+      Tests.discover(frameworkMap.values.toList, compileAnalysis, s.log)._1 ++
+        Tests.discover(frameworkMap.values.toList, itAnalysis, s.log)._1
```
jsancio: Can you explain this to me?

cosmos-integration-tests/src/it/scala/com/mesosphere/cosmos/NonSharedServiceDescribeSpec.scala
jsancio: How do you plan to discover that these tests are not shared?

jsancio: I see. This is in `it` not `main`. :+1:

build.sbt
cruhland: I think you noticed already, but I put the non-shared test under `it/`, so we have to discover tests there as well.

cosmos-integration-tests/src/it/scala/com/mesosphere/cosmos/NonSharedServiceDescribeSpec.scala
larioj: How about `OpenOnlyServiceDescribeSpec`?

larioj: Or we could put it under a module and name it the same. So like `open.ServiceDescribeSpec`.

cruhland: I like that idea. Let's do it the next time we need one of these tests.

PR #635, opened 2017-06-05
Package describe should always succeed.

jsancio:


cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -64,15 +64,20 @@ final class PackageDescribeSpec
     }
 
     "when requesting a v2 response" - {
-      "fails to describe helloworld v4 w/ updates when requesting v2 describe response" in {
+      "should succeed to describe helloworld v4 when requesting v2 describe response" in {
```
cruhland: Nitpick: "should succeed to describe" -> "successfully describes"

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -149,7 +154,7 @@ final class PackageDescribeSpec
       )
     )
 
-    val Right(actualContent) = parse(response.contentString)
```
cruhland: Why did this change?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageDescribeSpec.scala
```diff
@@ -162,7 +167,7 @@ final class PackageDescribeSpec
       rpc.v1.model.DescribeRequest("helloworld", version)
     )
     assertResult(Status.Ok)(response.status)
-    val Right(packageInfo) = parse(response.contentString)
```
cruhland: Why did this change?

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
```diff
@@ -79,45 +79,39 @@ object Response {
...
-          pkg.config,
-          pkg.command
-        )
+      val packagingVersion = pkg match {
+        case _: universe.v3.model.V2Package =>
+          universe.v3.model.V2PackagingVersion
+        case _: universe.v3.model.V3Package =>
+          universe.v3.model.V3PackagingVersion
+        case _: universe.v4.model.V4Package =>
+          universe.v3.model.V3PackagingVersion
```
cruhland: Should this be `V4PackagingVersion`?

cruhland: Based on this change, it looks like we are now okay with dropping information in responses. Should the `README` or `CONTRIBUTING` files be updated?

Do we have a ScalaCheck test that verifies this conversion works on any randomly-generated `PackageDefinition`?

cruhland:
@jsancio there seems to be a problem with the `dcos-launch` script for the integration tests, any ideas?

cosmos-server/src/main/scala/com/mesosphere/cosmos/converter/Response.scala
jsancio: If we don't do this we need to ask CLI and UI to update to the new version of the response in DCOS-15084.

jsancio:
Going with a different solution.

PR #637, opened 2017-06-06
Uninstall bug fixes and improvements.

benclarkwood:
[INFINITY-1686](https://jira.mesosphere.com/browse/INFINITY-1686) Rapid concurrent uninstalls fail
  - Previously, Marathon deployments were presumed to be sufficiently fast as to
    always complete before the first uninstall check. This is not accurate.
  - Switch to checking Marathon deployments to verify that the deployment
    from the Marathon update is complete.
  - Update tests.

[INFINITY-1691](https://jira.mesosphere.com/browse/INFINITY-1691) Force Marathon update for uninstall
  - Currently, if the service is in the middle of a deployment, the uninstall
    cannot be initiated.
  - Switch to issuing the update in modifyApp with force == true


cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouter.scala
```diff
@@ -21,12 +21,13 @@ class AdminRouter(
   def getApp(appId: AppId)(implicit session: RequestSession): Future[MarathonAppResponse] = marathon.getApp(appId)
 
   def modifyApp(
-    appId: AppId
+    appId: AppId,
+    force: Boolean
```
jsancio: I don't think we can/should do this. `force` behaviour is undefined by Marathon. Please confirm with the Marathon team.

benclarkwood: Force is explicitly defined and supported by Marathon:

https://mesosphere.github.io/marathon/docs/rest-api.html#put-v2appsappid

jsancio: Why not just fail the uninstall with the reason?

Just because it is documented it doesn't mean that it is suggested. Last time I talked to the Marathon team they were interested in removing that feature because the behaviour was undefined.

cruhland: @jsancio @benclarkwood I'm going to hold off on reviewing this PR until the expected behavior is agreed upon.

benclarkwood: The update must be a force.

In the case that the user installs the service, and it is unable to deploy the scheduler (e.g. bad config, or a bug in the scheduler is preventing launch), Marathon will never complete the deployment.

This will make Uninstall unable to be run in an extremely frustrating way for the user.

Additionally, it's a regression in the existing Uninstall behavior. The user wants the installed service gone. Let us not stand in their way.

benclarkwood: @jsancio @cruhland How can we move this forward?

jsancio: Yes. Sounds good!

jsancio:
LGTM! Thanks @benclarkwood

cosmos-server/src/main/scala/com/mesosphere/cosmos/MarathonClient.scala
```diff
@@ -37,7 +37,9 @@ class MarathonClient(
           .remove("uris")
           .remove("version")
 
-      client(put("v2" / "apps" / appId.toUri, Json.fromJsonObject(f(appJson))))
+      var uri = "v2" / "apps" / appId.toUri
+      uri = if (force) uri ? ("force" -> "true") else uri
```
cruhland: Oh, I see, you're using a `var` here -- I must have missed that when we discussed earlier. It's preferable in Scala to use a `val` when possible. Why not store the first line in `val uriPrefix` and the second line in `val uri`?

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouter.scala
```diff
@@ -21,12 +21,13 @@ class AdminRouter(
...
   )(
     f: JsonObject => JsonObject
   )(
     implicit session: RequestSession
-  ): Future[Response] = marathon.modifyApp(appId)(f: JsonObject => JsonObject)
+  ): Future[Response] = marathon.modifyApp(appId, force)(f: JsonObject => JsonObject)
```
cruhland: Nitpick: the type ascription on `f` isn't needed: `marathon.modifyApp(appId, force)(f)`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -131,11 +132,20 @@ private[cosmos] final class UninstallHandler(
...
-          UninstallDetails.from(op)
+          Decoders.parse(response.contentString).asObject match {
+            case Some(responseObject) =>
+              sdkJanitor.delete(
+                op.appId,
+                responseObject.apply("deploymentId").get.asString.get,
+                session)
+              UninstallDetails.from(op)
+            case _ =>
+              throw FailedToStartUninstall(op.appId, "Marathon update response is not a JSON Object: %s".format(response.contentString))
```
cruhland: Since extracting the deployment ID from the JSON could also fail, could this error case be used for that too? The message could be changed to `Could not obtain deployment ID from Marathon update response: %s`, which would be appropriate for both error conditions.

It might be more readable to factor out the extraction logic into a `parseDeploymentId()` method:

```scala
response.status match {
  case Status.Ok =>
    val deploymentId = parseDeploymentId(response.contentString)
    sdkJanitor.delete(op.appId, deploymentId, session)
    UninstallDetails.from(op)
  // ...
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/janitor/JanitorWorker.scala
```diff
@@ -81,6 +82,33 @@ final class JanitorWorker(
...
+  def checkDeployment(request: JanitorRequest): Boolean = {
+    logger.info("Checking the status of the deployment {} for app: {}", request.deploymentId, request.appId)
+
+    try {
+      val response = Await.result(adminRouter.listDeployments()(session = request.session))
+      Decoders.parse(response.contentString).asArray match {
+        case Some(deployments) =>
+          deployments.foreach { deployment =>
+            // If the deployment is present in the list, it is not yet complete.
+            if (deployment.asObject.get.apply("id").get.asString.get == request.deploymentId) {
```
cruhland: Using a cursor here can reduce the number of method calls:

```scala
deployment.cursor.get[String]("id").right.get == request.deploymentId
```

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/handler/UninstallHandlerSpec.scala
```diff
@@ -87,10 +87,13 @@ class UninstallHandlerSpec extends FreeSpec with BeforeAndAfterEach with Mockito
     "If marathon is updated, then the janitor is told to delete the app" in {
       when(mockSdkJanitor.claimUninstall(appId)).thenReturn(SdkJanitor.UninstallClaimGranted)
 
+      val updateResponse = Response.apply(Status.Ok)
+      updateResponse.setContentString("{\"deploymentId\": \"test-deployment\"}")
```
cruhland: Scala supports triple-quoted strings:

```scala
"""{"deploymentId": "test-deployment"}"""
```

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/janitor/JanitorWorkerSpec.scala
```diff
@@ -89,21 +95,49 @@ final class JanitorWorkerSpec extends FreeSpec with MockitoSugar
...
+          val mockResponse = Response(status = Status.Accepted)
+          val mockFuture = Future.value(mockResponse)
+          when(mockAdminRouter.getSdkServicePlanStatus(
+                                                        service = appId.toString,
+                                                        apiVersion = "v1",
+                                                        plan = "deploy"
+                                                      )(mockSession)).thenReturn(mockFuture)
+
+          val deploymentResponse = Response(status = Status.Ok)
+          deploymentResponse.setContentString("[{\"id\": \"deploymentid\"}]")
```
cruhland: Triple-quotes

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/janitor/JanitorWorkerSpec.scala
```diff
@@ -89,21 +95,49 @@ final class JanitorWorkerSpec extends FreeSpec with MockitoSugar
...
+            .thenReturn(Future.value(MarathonAppResponse(new MarathonApp(appId, Map()))))
 
-        val mockResponse = Response(status = Status.Accepted)
-        val mockFuture = Future.value(mockResponse)
-        when(mockAdminRouter.getSdkServicePlanStatus(
-          service = appId.toString,
-          apiVersion = "v1",
-          plan = "deploy"
-        )(mockSession)).thenReturn(mockFuture)
+          val mockResponse = Response(status = Status.Accepted)
```
cruhland: If you change this to `Status.Ok`, so that the uninstall is done, then I think this test will fail with your current code, because of the `foreach` in `checkDeployment()`.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/janitor/JanitorWorkerSpec.scala
```diff
@@ -89,21 +95,49 @@ final class JanitorWorkerSpec extends FreeSpec with MockitoSugar
...
+          val mockFuture = Future.value(mockResponse)
+          when(mockAdminRouter.getSdkServicePlanStatus(
+                                                        service = appId.toString,
+                                                        apiVersion = "v1",
+                                                        plan = "deploy"
```
cruhland: Fix indentation

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/janitor/JanitorWorkerSpec.scala
```diff
@@ -89,21 +95,49 @@ final class JanitorWorkerSpec extends FreeSpec with MockitoSugar
...
 
-        worker.doWork(request)
-        verifyZeroInteractions(mockTracker)
-        assertResult(1)(queue.size())
+          val mockResponse = Response(status = Status.Accepted)
+          val mockFuture = Future.value(mockResponse)
+          when(mockAdminRouter.getSdkServicePlanStatus(
+                                                        service = appId.toString,
+                                                        apiVersion = "v1",
+                                                        plan = "deploy"
```
cruhland: Fix indentation

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/janitor/JanitorWorkerSpec.scala
```diff
@@ -89,21 +95,49 @@ final class JanitorWorkerSpec extends FreeSpec with MockitoSugar
...
+                                                      )(mockSession)).thenReturn(mockFuture)
+
+          val deploymentResponse = Response(status = Status.Ok)
+          deploymentResponse.setContentString("[{\"id\": \"some-other-id\"}]")
```
cruhland: Triple quotes

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -145,8 +149,16 @@ private[cosmos] final class UninstallHandler(
...
+  }
+
+  private[this] def parseDeploymentId(content: String, op: UninstallOperation): String = {
+    try {
+      Decoders.parse(content).cursor.get[String]("deploymentId") match {
+        case Right(deploymentId) => deploymentId
+        case Left(_) => throw FailedToStartUninstall(op.appId, DeploymentIdErrorMessage.format(content))
+      }
+    } catch {
+      case _: CirceError => throw FailedToStartUninstall(op.appId, DeploymentIdErrorMessage.format(content))
```
cruhland: This is totally optional, but since you're dealing with `Either` here anyway, you might as well unify the error paths:

```scala
_root_.io.circe.jawn.parse(content)  // (this is what Decoders.parse() delegates to)
  .right.flatMap(_.cursor.get[String]("deploymentId"))
  .right.getOrElse {
    throw FailedToStartUninstall(op.appId, DeploymentIdErrorMessage.format(content))
  }
```

PR #638, opened 2017-06-08
Update README

jsancio:


README.md
```diff
@@ -148,67 +150,31 @@ The below table is a compatibility matrix between Cosmos and Universe packaging
...
+   1. [RAML](https://downloads.dcos.io/cosmos/0.4.0-SNAPSHOT-232-master-2fdc5cf8ad/service.raml)
+   1. [Swagger](https://downloads.dcos.io/cosmos/0.4.0-SNAPSHOT-232-master-2fdc5cf8ad/service.swagger)
+1. **Package Manager**
+   1. [HTML](https://downloads.dcos.io/cosmos/0.4.0-SNAPSHOT-232-master-2fdc5cf8ad/package.html)
+   1. [RAML](https://downloads.dcos.io/cosmos/0.4.0-SNAPSHOT-232-master-2fdc5cf8ad/package.raml)
+   1. [Swagger](https://downloads.dcos.io/cosmos/0.4.0-SNAPSHOT-232-master-2fdc5cf8ad/package.swagger)
+1. **Capability Manager**
+   1. [HTML](https://downloads.dcos.io/cosmos/0.4.0-SNAPSHOT-232-master-2fdc5cf8ad/capabilities.html)
+   1. [RAML](https://downloads.dcos.io/cosmos/0.4.0-SNAPSHOT-232-master-2fdc5cf8ad/capabilities.raml)
+   1. [Swagger](https://downloads.dcos.io/cosmos/0.4.0-SNAPSHOT-232-master-2fdc5cf8ad/capabilities.swagger)
```
cruhland: Hmm, I wonder if we should do something similar to the CLI and have the latest version of these files available at some unique URL. Then we wouldn't have to keep updating these links.

jsancio: I like for us to have immutable URLs. Even those these URLs say 0.4.0-snapshot, they are for 0.3.0 version of Cosmos. Unfortunately, we didn't have documentation publish by the time we released 0.3.0 so we used the snapshot build instead.

PR #639, opened 2017-06-08
Move the Repository class to v4

jsancio:


cosmos-common/src/main/scala/com/mesosphere/universe/v4/model/Repository.scala
```diff
@@ -1,4 +1,4 @@
-package com.mesosphere.universe.v3.model
```
cruhland: Shouldn't we keep a v3 `Repository` around that contains `v3` `PackageDefinition`s?

jsancio: We can't. We don't have `universe.v3.model.PackageDefinition`. We are going to need `universe.v3.model.Repository` when we implement the Universe Server. For that, I think we want to move to `type universe.v3.model.PackageDefinition = V2Package :+: V3Package` or something like that.

jsancio:

PR #645, opened 2017-06-20
GZIP Storage data

jsancio:


cosmos-common/src/main/scala/com/mesosphere/cosmos/model/StorageEnvelope.scala
```diff
@@ -27,21 +32,50 @@ final case class StorageEnvelope private (metadata: Map[String, String], data: B
...
+      if (encoding == StorageEnvelope.gzipEncoding) {
+        val byteStream = new ByteArrayOutputStream()
+        val gzipStream = new GZIPInputStream(new ByteArrayInputStream(ByteBuffers.getBytes(data)))
+
+        StreamIO.copy(gzipStream, byteStream)
+
+        byteStream.toByteArray()
+      } else {
+        throw new IllegalArgumentException(
+          "Error while trying to deserialize envelope data. Unknown Content-Encoding: gzip."
```
cruhland: Should this include the unknown encoding, and say that `gzip` is expected?

cosmos-common/src/main/scala/com/mesosphere/cosmos/model/StorageEnvelope.scala
```diff
@@ -27,21 +32,50 @@ final case class StorageEnvelope private (metadata: Map[String, String], data: B
       )
     }
 
+    val storageData = metadata.get(Fields.ContentEncoding).map { encoding =>
```
cruhland: It might be cleaner to select the decoding function independently from reading the data:

```scala
// The type of decodeData is Array[Byte] => Array[Byte]
val decodeData = metadata.get(Fields.ContentEncoding) match {
  case Some(StorageEnvelope.GzipEncoding) =>
    decodeGzip
  case Some(_) =>
    throw new IllegalArgumentException(...)
  case _ =>
    identity
}

val storageData = decodeData(ByteBuffers.getBytes(data))
mediaTypedDecode(
  new String(storageData, StandardCharsets.UTF_8),
  contentType
)
```

Then the decoding function can be implemented separately as:

```scala
def decodeGzip(data: Array[Byte]): Array[Byte] = {
  val byteStream = new ByteArrayOutputStream()
  // ...
  byteStream.toByteArray()
}
```

cosmos-common/src/main/scala/com/mesosphere/cosmos/model/StorageEnvelope.scala
```diff
@@ -27,21 +32,50 @@ final case class StorageEnvelope private (metadata: Map[String, String], data: B
...
     mediaTypedDecode(
-      new String(ByteBuffers.getBytes(data), StandardCharsets.UTF_8),
+      new String(storageData, StandardCharsets.UTF_8),
       contentType
     )
   }
 }
 
 object StorageEnvelope {
+  private val gzipEncoding = "gzip"
```
cruhland: It's usually a good idea to capitalize the names of constants so that they can be used in pattern matches. This should also have an explicit type, for whenever we get around to adding Scalastyle checks.

cosmos-common/src/main/scala/com/mesosphere/cosmos/model/StorageEnvelope.scala
```diff
@@ -27,21 +32,50 @@ final case class StorageEnvelope private (metadata: Map[String, String], data: B
...
 
   def apply[T](data: T)(implicit mediaTypedEncoder: MediaTypedEncoder[T]): StorageEnvelope = {
     val (json, mediaType) = mediaTypedEncoder(data)
 
+    val bytes = {
+      val byteStream = new ByteArrayOutputStream()
+      val gzipStream = new GZIPOutputStream(byteStream)
+      gzipStream.write(json.noSpaces.getBytes(StandardCharsets.UTF_8))
+      gzipStream.close()
+      ByteBuffer.wrap(byteStream.toByteArray())
```
cruhland: If you end up implementing my decoding comment above, would it make sense to factor this out into an `encodeGzip` method with the same `Array[Byte] => Array[Byte]` type?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -379,15 +380,15 @@ object ItObjects {
...
       .downField("marathonJson")
       .downField("labels")
       .downField("DCOS_PACKAGE_DEFINITION")
       .withFocus(base64Decode)
-      .downField("data")
-      .withFocus(base64Decode)
-      .top.get
+      .as[StorageEnvelope].right.get
+
+    envelope.decodeData[universe.v4.model.PackageDefinition].asJson
```
cruhland: Why decode and then convert to JSON? For validation?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -379,15 +380,15 @@ object ItObjects {
...
+    val envelope = renderResponse
       .hcursor
       .downField("marathonJson")
       .downField("labels")
       .downField("DCOS_PACKAGE_DEFINITION")
       .withFocus(base64Decode)
-      .downField("data")
-      .withFocus(base64Decode)
-      .top.get
+      .as[StorageEnvelope].right.get
```
cruhland: Doing `.as[StorageEnvelope]` feels a bit weird when `StorageEnvelope`'s constructor is `private`. We're bypassing the `StorageEnvelope.apply()` method, which should be the only way to create instances. Not sure if there's a good alternative though.

cosmos-common/src/main/scala/com/mesosphere/cosmos/model/StorageEnvelope.scala
jsancio: Yes. Fixed.

cosmos-common/src/main/scala/com/mesosphere/cosmos/model/StorageEnvelope.scala
jsancio: Done.

cosmos-common/src/main/scala/com/mesosphere/cosmos/model/StorageEnvelope.scala
jsancio: Done.

cosmos-common/src/main/scala/com/mesosphere/cosmos/model/StorageEnvelope.scala
jsancio: Done.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
jsancio: This was incorrect. Fixed it.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
jsancio: Technically we are calling `StorageEnvelope.decoder.apply` which call the private constructor.

jsancio:
@cruhland I addressed your comments.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -379,15 +380,28 @@ object ItObjects {
...
     renderResponse
       .hcursor
       .downField("marathonJson")
       .downField("labels")
       .downField("DCOS_PACKAGE_DEFINITION")
       .withFocus(base64Decode)
       .downField("data")
-      .withFocus(base64Decode)
-      .top.get
+      .withFocus(_ => pkg)
```
cruhland: I'm confused by this. Why can't you just return `pkg`? Could you give an example of what this is doing?

cruhland:
@jsancio one more question. Also, the render tests are still failing.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
jsancio: Yeah. I was confused too at first. This code is actually modifying the JSON object. E.g. it is turning:
```
{
  "id": "helloworld",
  "label": {
    "DCOS_PACKAGE_DEFINITION": "ABCD...XYZ"
  },
  ...
}
```
to
```
{
  "id": "helloworld",
  "label": {
    "DCOS_PACKAGE_DEFINITION": {
      "metadata": { ... },
      "data": {
        "name": "helloworld",
        "version": "0.1.0",
        ...
      }
    }
  },
  ...
}
```

cruhland: Ah, okay. I think that means we can remove the duplication in navigating to the right spot. Something like this (note that this is just a guess; I haven't tested it):

```scala
renderResponse
  .hcursor
  .downField("marathonJson")
  .downField("labels")
  .downField("DCOS_PACKAGE_DEFINITION")
  .withFocus(base64Decode)
  .withFocus { pkgDefJson =>
    decode[StorageEnvelope](pkgDefJson)
      .decodeData[universe.v4.model.PackageDefinition]
      .asJson
  }
  .top
  .get
```

jsancio: Done.

jsancio:
> @jsancio one more question. Also, the render tests are still failing.

@cruhland Done.

jsancio:

PR #646, opened 2017-06-20
Add source repository to Universe Client http error

larioj:
The meat of the changes lives in PackageRepositoryAddHandler and UniverseClient. All other changes are either codec changes (I added codecs to companion object of ErrorResponse) or tests.


cosmos-common/src/main/scala/com/mesosphere/cosmos/error/UniverseClientHttpError.scala
```diff
@@ -0,0 +1,33 @@
...
+
+final case class UniverseClientHttpError(
+  packageRepository: rpc.v1.model.PackageRepository,
+  method: HttpMethod,
+  clientStatus: Status
+) extends CosmosError {
+  override def data: Option[JsonObject] = CosmosError.deriveData(this)
+
+  override def message: String = {
+    "Unexpected down stream http error " +
```
jsancio: It is technically an upstream: https://en.wikipedia.org/wiki/Upstream_server

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -149,9 +150,9 @@ final class DefaultUniverseClient(
               /* If we are unable to get the latest Universe we should not forward the status code returned.
                * We should instead return 500 to the client and include the actual error in the message.
                */
-              throw GenericHttpError(
+              throw UniverseClientHttpError(
```
cruhland: We should be careful about changing error class names. The CLI and UI could potentially be using them to handle errors in a specific way.

larioj: I agree. I considered changing GenericHttpError to contain a PackageRepository, however It's too generic to do so, and it's used elsewhere. Do you have a suggestion as to how we should handle this?

larioj:
@jsancio The tests that was changed confirms this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
jsancio: I think it is okay to add new errors. We do that all of the time, right? What we have to be careful is with changing existing errors.

Most if not all of our clients should only be printing the `msg` field.

larioj: That's what the CLI is doing.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositoryIntegrationSpec.scala
```diff
@@ -87,24 +92,60 @@ final class PackageRepositoryIntegrationSpec extends FreeSpec with BeforeAndAfte
...
-        val sources = listRepositories()
-        assertResult(PackageRepository(addRequest.name, addRequest.uri))(sources(defaultRepos.size))
+      val sources = listRepositories()
+      val repoIndex = defaultRepos.size - 1
+      assertResult(repo)(sources(repoIndex)).withClue(sources)
     }
 
-    "allows insertion at specific index" in {
-        val addRequest = PackageRepositoryAddRequest("name", "http://fake.fake", Some(0))
+    "allows insertion at a specific index" in {
```
jsancio: It looks like either this test or the test "append to the list if no index is defined" changes the order of the default repos, right? This is going to affect other tests since we are changing the context.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositoryIntegrationSpec.scala
```diff
@@ -87,24 +92,60 @@ final class PackageRepositoryIntegrationSpec extends FreeSpec with BeforeAndAfte
...
+    }
 
-        val sources = listRepositories()
-        assertResult(PackageRepository(addRequest.name, addRequest.uri))(sources.head)
+    "returns an UniverseClientHttpError when uri return an error" in {
+      val name = "name"
+      val uri = "https://www.mesosphere.com/uontehusantoehusanth"
+      val addRequest = PackageRepositoryAddRequest(name, uri)
+      val response = sendAddRequest(addRequest)
+      assertResult(Status.InternalServerError)(response.status).withClue(response.contentString)
```
jsancio: Hmm. This should be a 400. `RepositoryUriConnection` returns a `400`. They are very similar error and should return the same HTTP status code, right?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
```diff
@@ -310,15 +283,7 @@ object PackageRepositorySpec extends FreeSpec with TableDrivenPropertyChecks {
...
-        DeleteSourceAssertion(deleteRequestByName(UniverseRepository), Nil))),
-    (List(SourceMesosphere, SourceCliTest4, SourceExample, UniverseRepository),
-      List(
-        DeleteSourceAssertion(deleteRequestByUri(SourceExample),
-          List(SourceMesosphere, SourceCliTest4, UniverseRepository)),
-        DeleteSourceAssertion(deleteRequestByUri(SourceCliTest4),
-          List(SourceMesosphere, UniverseRepository)),
-        DeleteSourceAssertion(deleteRequestByUri(SourceMesosphere), List(UniverseRepository)),
-        DeleteSourceAssertion(deleteRequestByUri(UniverseRepository), Nil)))
+        DeleteSourceAssertion(deleteRequestByName(UniverseRepository), Nil)))
```
jsancio: Can you explain this change to me?

jsancio: This spec is very similar to the other one you changed "PackageRepositoryIntegrationSpec" (or something like that). Do we remember why we have two?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
```diff
@@ -87,8 +86,6 @@ object CosmosIntegrationTestClient extends Matchers {
       decoder: Decoder[Res]
     ): Either[ErrorResponse, Res] = {
       val response = submit(request)
-      assertResult(expectedStatus)(response.status)
-
```
jsancio: Why did we remove this? If we are going to remove this please remove the `expectedStatus` argument.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageRepositoryAddHandler.scala
```diff
@@ -17,13 +19,16 @@ private[cosmos] final class PackageRepositoryAddHandler(
...
-          rpc.v1.model.PackageRepository(request.name, request.uri)
-        ) map { sources =>
-          rpc.v1.model.PackageRepositoryAddResponse(sources)
+        universeClient(repository).flatMap { _ =>
+          sourcesStorage.add(
+            request.index,
+            repository
+          ) map { sources =>
+            rpc.v1.model.PackageRepositoryAddResponse(sources)
+          }
```
jsancio: Let's document with a comment why we are doing this and why we are dropping the response: `flatMap { _ =>`

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
```diff
@@ -87,8 +86,6 @@ object CosmosIntegrationTestClient extends Matchers {
       decoder: Decoder[Res]
     ): Either[ErrorResponse, Res] = {
       val response = submit(request)
-      assertResult(expectedStatus)(response.status)
```
cruhland: Why did this get deleted?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
cruhland: Ok yeah, sounds like the CLI and UI aren't looking for `GenericHttpError` specifically, so this is fine.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
larioj: This is making the  method fail in cases where it should not. For example, when we are expecting the response to be Left(ErrorResponse). With this in there we will never get any lefts.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
larioj: This is making the method fail in cases where it should not. For example, when we are expecting the response to be Left(ErrorResponse). With this in there we will never get any lefts.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
larioj: I don't see a reason to have two (other than tech debt).

larioj: This change deletes one of the table examples (the one starting at line 314). For one it's redundant with the rest of the examples. Second one of the sources it uses is not a valid repo so it fails at add time when the tests expects it to succeed.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositoryIntegrationSpec.scala
larioj: We have a `before` step that sets the repo to what is expected. And we have an `afterAll` that sets it to what it was before the tests run.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
cruhland: This method is called in a bunch of places; having the assertion is likely important for at least some of them. For the cases where you need the `Left` I'd rather you use a separate method. Unless you can patch up all the call sites for this method to assert on the status in some other way.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
jsancio: Okay then let's check the status is 200 when it is `Right` and not 200 when it is `Left`.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
jsancio: > I don't see a reason to have two (other than tech debt).

Then let's merge them.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositoryIntegrationSpec.scala
larioj: That would make sense. The only question is why we have been returning 500 for this error in the past?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
larioj: Should I do it as part of this PR, or should I create an issue and tackle it separately?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
larioj: I'm taking Charles suggestion of creating a new method that does not do an assert.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageRepositoryAddHandler.scala
larioj: Done

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositoryIntegrationSpec.scala
jsancio: When we get this error from `/package/install` it makes sense to get a 500 because there is nothing the user can do to the HTTP request to make it not return an error. When it comes from `package/repo/add` then it needs to be a 400 because the user can change the request to make it succeed.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/repository/PackageRepositorySpec.scala
jsancio: This PR.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositorySpec.scala
```diff
@@ -0,0 +1,229 @@
...
+    scenario("the user should receive an error when trying to add a non-repository") {
+      val uri = "https://www.mesosphere.com/uontehusantoehusanth"
+      val name = "not-repository"
+      val repository = rpc.v1.model.PackageRepository(name, uri)
+      val actual = ItUtil.addRepositoryEither(repository)
+      val expected = UniverseClientHttpError(
+        PackageRepository(name, uri),
+        HttpMethod.GET,
+        Status.NotFound
+      ).exception(Status.BadRequest).errorResponse
```
jsancio: Where do we check that the HTTP status returned matches?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
```diff
@@ -121,6 +116,22 @@ object CosmosIntegrationTestClient extends Matchers {
...
+          case Left(_) => fail("Could not decode as successful response: " + response.contentString)
+          case Right(successfulResponse) => Right(successfulResponse)
+        }
+      } else {
+        decode[ErrorResponse](response.contentString) match {
+          case Left(_) => fail("Could not decode as error response: " + response.contentString)
+          case Right(errorResponse) => Left(errorResponse)
+        }
+      }
+    }
```
jsancio: :+1: 

We are still not checking the status for the error. We either need to check the status here or return it. E.g.
`toEither[Res](...)(....): Either[(Status, ErrorResponse, Res]`

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageRepositoryAddHandler.scala
```diff
@@ -17,16 +21,32 @@ private[cosmos] final class PackageRepositoryAddHandler(
...
     }
   }
 
+  private[this] def checkThatRepoWorks(
+    repository: rpc.v1.model.PackageRepository
+  )(
+    implicit session: RequestSession
+  ): Future[Unit] = {
+    // We get the repo to see if the operation succeeds
+    // We don't need the actual response
```
jsancio: Please use multiline comments if the sentences are related. E.g.
```
/* Get the repo to see if the operation succeeds. Ignore the returned value as it will be re-fetched
 * by the CosmosRepository class.
 */
```
vs unrelated comments:
```
// TODO: Address some issue
// TODO: Address another independent issue
```

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItUtil.scala
```diff
@@ -103,4 +121,10 @@ object ItUtil {
     }.dropWhile(done => !done).nonEmpty
   }
 
+
+  def insert[T](seq: Seq[T], index: Int, value: T): Seq[T] = {
+    val (front, back) = seq.splitAt(index)
+    front ++ Seq(value) ++ back
+  }
```
jsancio: This is the issue with "util" classes. This functionality has nothing to do with the functions above.

The probability of us reusing this code in a future PR is very low and this function is only used once. Just move it to the file that uses it.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositorySpec.scala
larioj: Added assert for status

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/test/CosmosIntegrationTestClient.scala
larioj: 👍

PR #648, opened 2017-06-23
Remove package storage

cruhland:
Partially addresses [DCOS-15975](https://jira.mesosphere.com/browse/DCOS-15975).

High-level description:
- Removes the `/package/add` and `/service/start` endpoints, and all code that was only needed for those two endpoints
- Removes test code that exercised any of the code that was removed
- Keeps the `packageStorageUri` and `stagedPackageStorageUri` flags and supporting code, because the integration tests won't run properly without them

I'll have a follow-up PR that updates the README and removes the package storage command-line flags. That change will require an update to the sbt-dcos build plugin so that we don't pass the storage flags to the cosmos instance under test.


cosmos-common/src/main/scala/com/mesosphere/cosmos/finch/RequestValidators.scala
```diff
@@ -37,31 +37,7 @@ object RequestValidators {
...
-    allValidators.map {
-      case authorization :: responseEncoder :: contentType :: bodyParser :: bodyBytes :: HNil =>
-        val requestBody = bodyParser(bodyBytes).get  // Exceptions will be caught by Endpoint.map()
-        val session = RequestSession(authorization, Some(contentType))
-        EndpointContext(requestBody, session, responseEncoder)
-    }
-  }
-
-  private[this] def baseValidator[Res](
+  def baseValidator[Res](
```
larioj: Why did this stay in open?

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PackageRepository.scala
```diff
@@ -1,5 +1,28 @@
...
+import com.mesosphere.universe.common.circe.Decoders._
+import com.mesosphere.universe.common.circe.Encoders._
 import com.netaporter.uri.Uri
+import io.circe.Decoder
+import io.circe.Encoder
+import io.circe.generic.semiauto._
 
 case class PackageRepository(name: String, uri: Uri)
+
+object PackageRepository {
```
larioj: 👍

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/test/TestUtil.scala
```diff
@@ -0,0 +1,9 @@
+package com.mesosphere.cosmos.test
+
+import com.mesosphere.cosmos.http.RequestSession
+
+object TestUtil {
```
larioj: Why do we need this file?

cosmos-common/src/main/scala/com/mesosphere/cosmos/finch/RequestValidators.scala
cruhland: It's referenced by the `noBody` and `standard` validators earlier in the file.

larioj: 👍

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/test/TestUtil.scala
cruhland: It originally lived [here](https://github.com/dcos/cosmos/pull/648/files/768068ed47cf903387e95f176b817c46272f4f01#diff-7e070c108afb5c54e5e737828cae05fdL59), but it's not used outside of this project so I moved it to the `test` build configuration.

jsancio:
@cruhland Can I get a high-level description of what this PR is doing?

cruhland:
@jsancio updated. Let me know if you want more detail, e.g. a summary of the classes removed.

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/MediaTypes.scala
```diff
@@ -4,10 +4,10 @@ import com.mesosphere.cosmos.http.MediaType
...
 object MediaTypes {
 
-  private[this] val dcos = MediaType.vndJson(List("dcos")) _
-  private[this] val dcosPackage = MediaType.vndJson(List("dcos", "package")) _
-  private[this] val dcosService = MediaType.vndJson(List("dcos", "service")) _
-  private[this] def vnd(kind: String, version: Int = 1) = dcosPackage(kind, version)
+  val dcos = MediaType.vndJson(List("dcos")) _
+  val dcosPackage = MediaType.vndJson(List("dcos", "package")) _
+  val dcosService = MediaType.vndJson(List("dcos", "service")) _
+  def vnd(kind: String, version: Int = 1): MediaType = dcosPackage(kind, version)
```
jsancio: Maybe we should get rid of this function and instead have the user use `dcosPackage` which is explicit about the version. `dcosPackage` is also a more descriptive function name compared to `vnd` which is technically and inaccurate name.

docs/service.raml
```diff
@@ -1,5 +1,4 @@
 #%RAML 1.0
 title: DC/OS Service
 /service:
-  /start: !include endpoints/service-start.raml
```
jsancio: Okay. Eventually, I want to move this back. This RPC should equal `/package/install` and we can deprecate `/package/install`.

PR #657, opened 2017-07-06
Update version compatibility matrix and some SBT commands

cruhland:


cruhland:
This is a README update so it does not affect the integration tests. Merging.

PR #658, opened 2017-07-06
Refactor error classes to extend CosmosError

takirala:
Removed `PackageDefinitionRenderError.scala` and all the error classes in it are refactored to go under `cosmos-common/src/main/scala/com/mesosphere/cosmos/error` to extend `CosmosError` (previously they used to extend from `RuntimeException` which is Throwable)


cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -157,16 +161,20 @@ object PackageDefinitionRenderer {
...
     (pkgDef.config, options.nonEmpty) match {
       // Success scenarios
-      case (None, false) => Right(())
-      case (Some(_), false) => Right(())
+      case (None, false) => true
+      case (Some(_), false) => true
       // Failure scenarios
-      case (None, true) => Left(OptionsNotAllowed)
+      case (None, true) => {
+        throw OptionsNotAllowed("No schema available to validate the provided options").exception
```
jsancio: See my other comments but this should just be `throw OptionsNotAllowed().exception`.

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -157,16 +161,20 @@ object PackageDefinitionRenderer {
...
+      }
       case (Some(schema), true) =>
-        JsonSchema.jsonObjectMatchesSchema(options, schema)
-          .leftMap(OptionsValidationFailure)
+        JsonSchema.jsonObjectMatchesSchema(options, schema) match {
+          case Left(validationErrors) => throw OptionsValidationFailure(validationErrors).exception
```
jsancio: See my other comment but I think we want this to be `throw JsonSchemaMismatch(validationError).exception`.

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -199,8 +207,8 @@ object PackageDefinitionRenderer {
     // If marathon ever changes its schema for labels then this code will most likely need a
     // new version with this version left intact for backward compatibility reasons.
     obj.cursor.getOrElse[Map[String, String]]("labels")(Map.empty) match {
-      case Left(err) => Left(InvalidLabelSchema(err))
-      case Right(labels) => Right(Json.fromFields(labels.mapValues(_.asJson)))
+      case Left(err) => throw InvalidLabelSchema(err.getMessage).exception
```
jsancio: `throw CirceError(err).exception`

jsancio: We should just use: https://github.com/dcos/cosmos/blob/master/cosmos-common/src/main/scala/com/mesosphere/cosmos/circe/Decoders.scala#L39-L42. Looks like we may have to make it public.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageRenderHandler.scala
```diff
@@ -1,59 +1,40 @@
...
-import scala.util.Left
-import scala.util.Right
 
 private[cosmos] final class PackageRenderHandler(
   packageCache: PackageCollection
 ) extends EndpointHandler[RenderRequest, RenderResponse] {
 
   override def apply(request: RenderRequest)(implicit
-    session: RequestSession
+                                             session: RequestSession
```
jsancio: We usually comment this as:
```
override def apply(
  request: ...
)(
  implicit session: ....
): ... = {
```

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/render/PackageDefinitionRendererSpec.scala
```diff
@@ -250,8 +253,8 @@ class PackageDefinitionRendererSpec extends FreeSpec with Matchers with TableDri
         )
       ).asObject.get
 
-      val Left(err) = PackageDefinitionRenderer.renderMarathonV2App("http://someplace", pkg, Some(options), None)
-      assertResult(OptionsNotAllowed)(err)
+      val exception = intercept[CosmosError](PackageDefinitionRenderer.renderMarathonV2App("http://someplace", pkg, Some(options), None))
+      assertResult(OptionsNotAllowed)(exception)
```
jsancio: I think you want `assertResult(OptionsNotAllowed)(exception.error)`

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/OptionsNotAllowed.scala
```diff
@@ -0,0 +1,14 @@
+package com.mesosphere.cosmos.error
+
+import io.circe.Encoder
+import io.circe.JsonObject
+import io.circe.generic.semiauto.deriveEncoder
+
+final case class OptionsNotAllowed(msg : String) extends CosmosError {
+  override def data: Option[JsonObject] = CosmosError.deriveData(this)
+  override def message: String = msg
```
takirala: @jsancio That is how it was built initially. But that would be a static error message right? We cannot add any dynamic information in that. Is this what you have in mind or am I missing something?

```
final case class OptionsNotAllowed() extends CosmosError {
  override def data: Option[JsonObject] = None
  override def message: String = "No schema available to validate the provided options"
}

object OptionsNotAllowed {
  implicit val encoder: Encoder[OptionsNotAllowed] = deriveEncoder
}
```

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/InvalidLabelSchema.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import io.circe.DecodingFailure
+import io.circe.Encoder
+import io.circe.JsonObject
+import io.circe.generic.semiauto.deriveEncoder
+
+final case class InvalidLabelSchema(cause: String) extends CosmosError {
+  override def data: Option[JsonObject] = CosmosError.deriveData(this)
+  override def message: String = cause
+}
```
takirala: I thought that might be a part of second PR. But it makes sense so that next PR will be less complicated. May be I misunderstood.

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -46,20 +50,20 @@ object PackageDefinitionRenderer {
...
          * json object to send into the mustache context for rendering. The following seq
          * prepares for the merge of all the options, documents at later indices have higher
          * priority than lower index objects order here is important, DO NOT carelessly re-order.
          */
         val mergedOptions = (
           defaultOptionsAndUserOptions ::
-          pkgDef.resourceJson.map(rj => JsonObject.singleton("resource", rj)).toList
-        ).foldLeft(JsonObject.empty)(JsonUtil.merge)
+            pkgDef.resourceJson.map(rj => JsonObject.singleton("resource", rj)).toList
+          ).foldLeft(JsonObject.empty)(JsonUtil.merge)
```
takirala: Changed it accordingly. Interestingly though, my IDE forced be to have this indentation. I had to manually enforce that. However isn't correct indentation in a _sense_? May be we should standardize this somehow. Just a thought.

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
takirala: I don't think i can use that here as there is some transformation going on in the `Right(..)` part. Am I right? 
I will make sure to use that in future.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/OptionsNotAllowed.scala
jsancio: Yes. Your suggestion looks correct.

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
jsancio: Okay. Let's bring this up during our next standup.

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -199,8 +207,8 @@ object PackageDefinitionRenderer {
     // If marathon ever changes its schema for labels then this code will most likely need a
     // new version with this version left intact for backward compatibility reasons.
     obj.cursor.getOrElse[Map[String, String]]("labels")(Map.empty) match {
-      case Left(err) => Left(InvalidLabelSchema(err))
-      case Right(labels) => Right(Json.fromFields(labels.mapValues(_.asJson)))
+      case Left(err) => throw CirceError(err).exception
+      case Right(labels) => Some(Json.fromFields(labels.mapValues(_.asJson)))
```
jsancio: You should still be able to use `convertToException`. E.g.:
```scala
val labels = convertToException(obj.cursor.getOrElse[Map[String, String]]("labels")(Map.empty))
Some(Json.fromFields(labels.mapValues(_.asJson)))
```

Also, when you make this public I think we should change the name of the method to something more descriptive. E.g. `convertToExceptionOfCirceError`. Or something like that.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageRenderHandler.scala
```diff
@@ -1,59 +1,42 @@
 package com.mesosphere.cosmos.handler
 
-import com.mesosphere.cosmos.error.CirceError
-import com.mesosphere.cosmos.error.JsonSchemaMismatch
-import com.mesosphere.cosmos.error.MarathonTemplateMustBeJsonObject
-import com.mesosphere.cosmos.error.ServiceMarathonTemplateNotFound
+import com.mesosphere.cosmos.error._
```
jsancio: No wild card import please. Unless we are importing implicit `val`s.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/render/PackageDefinitionRendererSpec.scala
```diff
@@ -3,8 +3,11 @@ package com.mesosphere.cosmos.render
 import cats.syntax.either._
 import com.mesosphere.cosmos.bijection.CosmosConversions._
 import com.mesosphere.cosmos.circe.Decoders.decode64
-import com.mesosphere.cosmos.circe.Decoders.parse64
 import com.mesosphere.cosmos.circe.Decoders.parse
+import com.mesosphere.cosmos.circe.Decoders.parse64
+import com.mesosphere.cosmos.error.CosmosError
+import com.mesosphere.cosmos.error.CosmosError
+import com.mesosphere.cosmos.error._
```
jsancio: No wild card imports.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageInstallIntegrationSpec.scala
```diff
@@ -151,7 +151,7 @@ final class PackageInstallIntegrationSpec extends FreeSpec with BeforeAndAfterAl
 
       val errorData = JsonObject.singleton("errors", List(schemaError).asJson)
       val errorResponse =
-        ErrorResponse("JsonSchemaMismatch", "Options JSON failed validation", Some(errorData))
+        ErrorResponse("OptionsValidationFailure", "Options JSON failed validation", Some(errorData))
```
jsancio: The third parameter should be `None`, right? Does this mean that this test is not checking for this?

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/render/PackageDefinitionRendererSpec.scala
```diff
@@ -266,9 +271,8 @@ class PackageDefinitionRendererSpec extends FreeSpec with Matchers with TableDri
         marathon = Some(Marathon(mustacheBytes))
       )
 
-      val Left(RenderedTemplateNotJson(ParsingFailure(_, cause))) =
-        PackageDefinitionRenderer.renderMarathonV2App("http://someplace", pkg, None, None)
-      assert(cause.isInstanceOf[jawn.IncompleteParseException])
+      val err = intercept[CosmosException](PackageDefinitionRenderer.renderMarathonV2App("http://someplace", pkg, None, None))
+      assert(err.error.message.equals("exhausted input"))
```
jsancio: This is Java code :). Scala `==` uses `Object`'s `equals` for `AnyRef` (http://scala-lang.org/api/current/scala/AnyRef.html#==(x$1:Any):Boolean). Having said that in this case you want to use ScalaTest's `assertResult` or better yet `Matchers`'s `shouldBe`. E.g.:

```scala
exception.error.message shouldBe "exhausted input"
```

Notice how I renamed `err` to `exception`. Much easier to read later. Also, let's make sure that we test the concrete type of the `exception.error` variable. E.g.:

```scala
exception.error shouldBe a[...]
```
Checkout this link for a list of some of the more useful matchers: http://www.scalatest.org/user_guide/using_matchers

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/render/PackageDefinitionRendererSpec.scala
```diff
@@ -283,8 +287,8 @@ class PackageDefinitionRendererSpec extends FreeSpec with Matchers with TableDri
         marathon = Some(Marathon(mustacheBytes))
       )
 
-      val Left(err) = PackageDefinitionRenderer.renderMarathonV2App("http://someplace", pkg, None, None)
-      assertResult(RenderedTemplateNotJsonObject)(err)
+      val exception = intercept[CosmosException](PackageDefinitionRenderer.renderMarathonV2App("http://someplace", pkg, None, None))
+      assertResult(MarathonTemplateMustBeJsonObject.message)(exception.getMessage)
```
jsancio: I would change this check to:
```scala
exception.error shouldBe MarathonTemplateMustBeJsonObject
```

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -46,11 +50,11 @@ object PackageDefinitionRenderer {
...
     options: Option[JsonObject],
     marathonAppId: Option[AppId]
-  ): Either[PackageDefinitionRenderError, JsonObject] = {
-    pkgDef.marathon.map { marathon =>
+  ): Option[JsonObject] = {
+    pkgDef.marathon.flatMap { marathon =>
       val defaultOptionsAndUserOptions: JsonObject = mergeDefaultAndUserOptions(pkgDef, options)
 
-      validateOptionsAgainstSchema(pkgDef, defaultOptionsAndUserOptions).flatMap { _ =>
+      if (validateOptionsAgainstSchema(pkgDef, defaultOptionsAndUserOptions)) {
```
cruhland: I noticed that, since you're now throwing exceptions instead of returning an `Either` from `validateOptionsAgainstSchema()`, you made the return type `Boolean`. However, `false` is never returned from that method! So you should be able to change it to return `Unit` instead, and then call it directly here without needing to wrap it in an `if`.

Ultimately, you should be able to change the `flatMap` into a `map` on the first line of the method (more comments about this later):

```scala
pkgDef.marathon.map { marathon =>
  // ...
}
```

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -102,9 +106,9 @@ object PackageDefinitionRenderer {
     }
 
     parse(renderedJsonString).map(_.asObject) match {
-      case Left(pe)           => Left(RenderedTemplateNotJson(pe))
-      case Right(None)        => Left(RenderedTemplateNotJsonObject)
-      case Right(Some(obj))   => Right(obj)
+      case Left(pe)           => throw CirceError(pe).exception
+      case Right(None)        => throw MarathonTemplateMustBeJsonObject.exception
+      case Right(Some(obj))   => Some(obj)
```
cruhland: Here's another place where using exceptions has made using `Either`/`Option` redundant. There's no case here where `None` is returned, so you might as well make the return type of the method a bare `JsonObject`.

In general, `Option`, `Either`, `Try`, and `Future` are alternatives to exceptions for error reporting; if you're switching to using exceptions then usually you won't need to use them.

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
cruhland: Another place where you can change the return type to `Json` instead of `Option[Json]`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageRenderHandler.scala
```diff
@@ -1,59 +1,42 @@
...
 import com.mesosphere.cosmos.finch.EndpointHandler
 import com.mesosphere.cosmos.http.RequestSession
 import com.mesosphere.cosmos.render._
 import com.mesosphere.cosmos.repository.PackageCollection
-import com.mesosphere.cosmos.rpc.v1.model.{RenderRequest, RenderResponse}
+import com.mesosphere.cosmos.rpc.v1.model.RenderRequest
+import com.mesosphere.cosmos.rpc.v1.model.RenderResponse
```
cruhland: We've been gradually trying to fix direct imports of `vX.model` classes. Instead, we import the overall package (`com.mesosphere.cosmos.rpc` in this case) and then refer to the classes in the code with qualified names (e.g. `rpc.v1.model.RenderRequest`). This makes it much easier to keep track of the versions we're using.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/render/PackageDefinitionRendererSpec.scala
```diff
@@ -88,7 +92,7 @@ class PackageDefinitionRendererSpec extends FreeSpec with Matchers with TableDri
         )
         val pkg = packageDefinition(pkgDef)(json.noSpaces)
 
-        val Right(rendered) = PackageDefinitionRenderer.renderMarathonV2App("http://someplace", pkg, None, None)
+        val rendered = PackageDefinitionRenderer.renderMarathonV2App("http://someplace", pkg, None, None).get
```
cruhland: Let's keep the form of the code similar to the original. Instead of using `.get` you can have `val Some(rendered) = ...`

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/render/PackageDefinitionRendererSpec.scala
```diff
@@ -116,12 +120,12 @@ class PackageDefinitionRendererSpec extends FreeSpec with Matchers with TableDri
           config = Some(buildConfig(Json.fromJsonObject(defaultsJson)))
         )
 
-        val Right(marathonJson) = PackageDefinitionRenderer.renderMarathonV2App(
+        val marathonJson = PackageDefinitionRenderer.renderMarathonV2App(
```
cruhland: `val Some(marathonJson) = ...`

PR #659, opened 2017-07-10
Updated Circe Error with more information

takirala:


cosmos-common/src/main/scala/com/mesosphere/cosmos/circe/Decoders.scala
```diff
@@ -36,9 +45,28 @@ object Decoders {
...
+          ("line", Json.fromInt(line)),
+          ("index", Json.fromInt(index)),
+          ("col", Json.fromInt(col)),
+          ("data", Json.fromString(jsonData))
+        )
+        Json.fromFields(fieldList).asObject
+    }else None
+  }
+
+  def convertToExceptionOfCirceDecodingError[T](result: Either[Error, T]): T = result match {
```
takirala: @larioj @jsancio What do you think of this method? The output of this method is the the parameter for `data` in `CirceParsingError`. As the json data may be a "invalid json", I thought to wrap the original json as a string inside a new json object which can have other fields such as line, column etcetera. If this is not acceptable, what alternative do you suggest?

jsancio: @takirala When the user gets one of these errors through HTTP these are the outputs that I would like to see:

```javascript
{
  "type": "JsonParsingError",
  "message": "Unable to parse the string as a JSON value",
  "data": {
    "string": "..."
  }
}
```
and
```javascript
{
  "type": "JsonDecodingError",
  "message": "Unable to decode the JSON value as a $typeName",
  "data": {
    "typeName": "...",
    "json": "..."
  }
}
```

We may be able to do better but let's start with this.

takirala: @jsancio Right now, the error response that I am building is 

```
CirceParsingError,Unable to parse the json at line 3,Some(object[data -> "{\n  \"id\": \"helloworld-invalid\",\n  \"cpus\": foobar,\n  \"mem\": 512,\n  \"instances\": 1,\n  \"cmd\": \"python3 -m http.server 9000\",\n  \"container\": {\n    \"type\": \"DOCKER\",\n    \"docker\": {\n      \"image\": \"python:3\",\n      \"network\": \"HOST\"\n    }\n  }\n}\n",line -> 3,col -> 43,index -> 42]))
```
Which would translate to 

```
{
   "type" : "JsonParsingError",
   "message" : "Unable to parse the json at line 3",
   "data" : {
         "string" : "{\n  \"id\": \"helloworld-invalid\",\n  \"cpus\": foobar,\n  \"mem\": 512,\n  \"instances\": 1,\n  \"cmd\": \"python3 -m http.server 9000\",\n  \"container\": {\n    \"type\": \"DOCKER\",\n    \"docker\": {\n      \"image\": \"python:3\",\n      \"network\": \"HOST\"\n    }\n  }\n}\n",
         "line" : 3,
         "col"  : 43,
         "index" : 42
}
```

jsancio: Not in all cases: https://github.com/dcos/cosmos/pull/659/files#diff-2b5d98c415c37e5d6f11f9bd2f380bdbR12

takirala: In the format available to us, One of {line} or {col} makes sense exclusively (i am inclined towards col) but not both together. As of now I will continue developing with all the information and we can decide later which one to get rid of.

takirala: Ah Okay! I will make sure to keep the `message` consistent

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/CirceParsingError.scala
```diff
@@ -0,0 +1,18 @@
+package com.mesosphere.cosmos.error
+
+import io.circe.Json
+import io.circe.JsonObject
+
+final case class CirceParsingError(
+    circeError: io.circe.Error,
+    jsonData: Option[JsonObject]
+  ) extends CosmosError {
```
jsancio: To match example you gave this should be:
```scala
final case class JsonParsingError(
  string: String,
  line: Int,
  column: Int,
  index : Int
) extends CosmosError {
...
}
```

The difference between column and index is not clear to me.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/CirceDecodingError.scala
```diff
@@ -2,7 +2,7 @@ package com.mesosphere.cosmos.error
 
 import io.circe.JsonObject
 
-final case class CirceError(circeError: io.circe.Error) extends CosmosError {
+final case class CirceDecodingError(circeError: io.circe.Error) extends CosmosError {
```
jsancio: To match the example I gave this should be:

```scala
final case class JsonDecodingError(
  typeName: String,
  json: String
) extends CosmosError {
...
}
```

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/CirceParsingError.scala
takirala: @jsancio I was not able to find any documentation on that but I believe (through trail and error and intuition) that col-index = 1. Index is where exactly the error occurred and col is where the first invalid char was found. In our case (where we have a single line json) having both of them doesn't make much sense but if we are parsing a multiline json, I thought may be their values would differ. We can get rid of one of them.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/CirceDecodingError.scala
takirala: Yes, I will be changing that.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/JsonDecodingError.scala
```diff
@@ -0,0 +1,16 @@
+package com.mesosphere.cosmos.error
+
+import io.circe.Json
+import io.circe.JsonObject
+
+final case class JsonDecodingError(
+  typeName : String = "",
```
jsancio: Remove the default value.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/JsonDecodingError.scala
```diff
@@ -0,0 +1,16 @@
...
+  json : String
+) extends CosmosError {
+  override def data: Option[JsonObject] = Json.fromFields(
+    List(
+      ("typeName",Json.fromString(typeName)),
+      ("string", Json.fromString(json))
+    )).asObject
```
jsancio: Use a Circe encoder. See how we do it in other classes. E.g. https://github.com/dcos/cosmos/blob/master/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/PackageNotFound.scala#L8

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -156,7 +156,7 @@ private[cosmos] final class UninstallHandler(
           ).exception
       }
     } catch {
-      case ex: CosmosException if ex.error.isInstanceOf[CirceError] =>
+      case ex: CosmosException if ex.error.isInstanceOf[JsonParsingError] =>
```
larioj: Can this not fail with a JsonDecodingError?

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/JsonDecodingError.scala
```diff
@@ -0,0 +1,17 @@
+package com.mesosphere.cosmos.error
+
+import io.circe.Encoder
+import io.circe.JsonObject
+import io.circe.generic.semiauto.deriveEncoder
+
+final case class JsonDecodingError(
+  typeName : String,
+  json : String
```
jsancio: Add a parameter `decodingErrorMessage`.

cosmos-common/src/main/scala/com/mesosphere/cosmos/circe/Decoders.scala
```diff
@@ -36,9 +45,34 @@ object Decoders {
...
+  def convertToExceptionOfCirceDecodingError[T](
+    result: Either[Error, T],
+    value: String
+  ): T = result match {
     case Right(result) => result
-    case Left(error) => throw CirceError(error).exception
+    case Left(error) => {
+      val df = error.asInstanceOf[DecodingFailure]
+      throw JsonDecodingError(typeName = df.message, value).exception
+    }
```
jsancio: I think we need to change this method to:
```scala
def convertToExceptionOfACosmsoError[T](
  result: Either[Error, T],
  inputValue: String
): T = result match {
  case Right(value) => value
  case Left(ParsingFailure(message, underlying)) =>
    throw JsonParsingError(inputValue, message, underlying.getClass.getName)
  case Left(DecodeFailure(message, _) =>
    throw JsonDecodingFailure(classOf[T].getName, inputValue, message)
}

File: cosmos-common/src/main/scala/com/mesosphere/cosmos/circe/Decoders.scala [hunk 9]
```diff
@@ -36,9 +45,34 @@ object Decoders {
...
+  }
+
+  def convertToExceptionOfCirceDecodingError[T](
+    result: Either[Error, T],
+    value: String
+  ): T = result match {
     case Right(result) => result
-    case Left(error) => throw CirceError(error).exception
+    case Left(error) => {
+      val df = error.asInstanceOf[DecodingFailure]
```
larioj: How do we know that this is always of type `DecodingFailure`?

cosmos-common/src/main/scala/com/mesosphere/cosmos/circe/Encoders.scala
```diff
@@ -18,6 +18,7 @@ import org.jboss.netty.handler.codec.http.HttpMethod
 
 object Encoders {
 
+
```
larioj: delete

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -103,7 +102,10 @@ object PackageDefinitionRenderer {
     }
 
     parse(renderedJsonString).map(_.asObject) match {
-      case Left(pe)           => throw CirceError(pe).exception
+      case Left(pe)           =>
+        throw JsonParsingError(pe.underlying.getClass.getName,
+          pe.message,
+          renderedJsonString).exception
       case Right(None)        => throw MarathonTemplateMustBeJsonObject.exception
       case Right(Some(obj))   => obj
```
jsancio: Please refactor this to use our `parse` function.

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -203,7 +206,8 @@ object PackageDefinitionRenderer {
     // the user know here where we can craft a more informational error message.
     // If marathon ever changes its schema for labels then this code will most likely need a
     // new version with this version left intact for backward compatibility reasons.
-    val labels = convertToExceptionOfCirceError(obj.cursor.getOrElse[Map[String, String]]("labels")(Map.empty))
+    val labels = convertToExceptionOfACosmosError(
+      obj.cursor.getOrElse[Map[String, String]]("labels")(Map.empty), obj.toString())
```
jsancio: Change to:
```scala
val labels = convertToExceptionOfACosmosError(
  obj.cursor.getOrElse[Map[String, String]]("labels")(Map.empty),
  obj.noSpaces
) 
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/ServiceClient.scala
```diff
@@ -86,7 +87,8 @@ abstract class ServiceClient(baseUri: Uri) {
     }
   }
 
-  protected def decodeTo[A](method: HttpMethod, uri: Uri, response: Response)(implicit d: Decoder[A]): Future[A] = {
+  protected def decodeTo[A: Decoder : ClassTag](method: HttpMethod, uri: Uri, response: Response):
+  Future[A] = {
```
jsancio: This doesn't match our standard formatting:
```scala
protected def decodeTo[A: Decoder: ClassTag](
  method: HttpMethod,
  uri: Uri,
  response: Response
): Future[A]
```

cosmos-common/src/main/scala/com/mesosphere/cosmos/circe/Decoders.scala
```diff
@@ -1,48 +1,58 @@
...
   }
 
   def parse64(value: String): Json = {
     parse(base64DecodeString(value))
   }
 
-  def convertToExceptionOfCirceError[T](result: Either[Error, T]): T = result match {
-    case Right(result) => result
-    case Left(error) => throw CirceError(error).exception
+  def convertToExceptionOfACosmosError[T: ClassTag](
```
jsancio: change to `convertToCosmosException`

cosmos-server/src/main/scala/com/mesosphere/cosmos/ServiceClient.scala
```diff
@@ -87,8 +87,11 @@ abstract class ServiceClient(baseUri: Uri) {
     }
   }
 
-  protected def decodeTo[A: Decoder : ClassTag](method: HttpMethod, uri: Uri, response: Response):
-  Future[A] = {
+  protected def decodeTo[A: Decoder : ClassTag](
```
jsancio: Extra space: `...[A: Decoder: ClassTag](`

jsancio:
@takirala Looks like you are getting an error when we try to generate our documentation. Let's take a look at it tomorrow morning.

PR #660, opened 2017-07-14
Update error message while adding a repo with an out of bound index

takirala:
- Updated error message while adding a repository with an index that would result in out of bound exception
- Added a test case to check the same

### DIFF
--- a/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/RepositoryAddIndexOutOfBounds.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/RepositoryAddIndexOutOfBounds.scala
@@ -6,7 +6,7 @@ import io.circe.generic.semiauto.deriveEncoder
 
 final case class RepositoryAddIndexOutOfBounds(attempted: Int, max: Int) extends CosmosError {
   override def data: Option[JsonObject] = CosmosError.deriveData(this)
-  override def message: String = s"Index out of range: $attempted"
+  override def message: String = s"Index ($attempted) is out of range. Index value must be between 0 and $max inclusive."
 }
 
 object RepositoryAddIndexOutOfBounds {
--- a/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/RepositoryAddIndexOutOfBounds.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/RepositoryAddIndexOutOfBounds.scala
@@ -6,7 +6,10 @@ import io.circe.generic.semiauto.deriveEncoder
 
 final case class RepositoryAddIndexOutOfBounds(attempted: Int, max: Int) extends CosmosError {
   override def data: Option[JsonObject] = CosmosError.deriveData(this)
-  override def message: String = s"Index ($attempted) is out of range. Index value must be between 0 and $max inclusive."
+  override def message: String = s"Index ($attempted) is out of range. Index value " + s"${
+      if (max == 0) "can be at most 0"
+      else s"must be between 0 and $max inclusive."
+  }"
 }
 
 object RepositoryAddIndexOutOfBounds {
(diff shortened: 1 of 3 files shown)


cosmos-common/src/main/scala/com/mesosphere/cosmos/error/RepositoryAddIndexOutOfBounds.scala
```diff
@@ -6,7 +6,7 @@ import io.circe.generic.semiauto.deriveEncoder
 
 final case class RepositoryAddIndexOutOfBounds(attempted: Int, max: Int) extends CosmosError {
   override def data: Option[JsonObject] = CosmosError.deriveData(this)
-  override def message: String = s"Index out of range: $attempted"
+  override def message: String = s"Index ($attempted) is out of range. Index value must be between 0 and $max inclusive."
```
takirala: @larioj We were talking yesterday what if the the `max = 0` but that case never happens due to the check here - https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/ZkRepositoryList.scala#L180

larioj: What will the message be when list.size = 0 and I attempt to add at index 20?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositorySpec.scala
```diff
@@ -89,6 +89,18 @@ class PackageRepositorySpec
       assertResult(Status.BadRequest)(status)
       info(expected.message)
     }
+    scenario("the user should be able to add a repository at the end of the list") {
```
larioj: "the user should be able to add a repository at the end of the list using an index"

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/RepositoryAddIndexOutOfBounds.scala
```diff
@@ -6,7 +6,10 @@ import io.circe.generic.semiauto.deriveEncoder
 
 final case class RepositoryAddIndexOutOfBounds(attempted: Int, max: Int) extends CosmosError {
   override def data: Option[JsonObject] = CosmosError.deriveData(this)
-  override def message: String = s"Index out of range: $attempted"
+  override def message: String = s"Index ($attempted) is out of range. Index value " + s"${
+      if (max == 0) "can be at most 0"
```
larioj: What do you think of 'must be 0'?

jsancio: Too fancy in my opinion:

```scala
override def message: String = {
  val suggestion = if (max == 0) "can be at most 0" else s"must be between 0 and $max inclusive."
  s"Index ($attempted) is out of range. Index value " + suggestion
}
```

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/PackageRepositorySpec.scala
```diff
@@ -89,6 +89,18 @@ class PackageRepositorySpec
...
+      val name = "bounds"
+      val index = defaultRepositories.size
+      val repository = rpc.v1.model.PackageRepository(name, uri)
+      val (status, actual) = ItUtil.addRepositoryEither(repository, Some(index))
+      val expected = rpc.v1.model.PackageRepositoryAddResponse(
+        originalRepositories :+ repository
+      )
+      assertResult(Right(expected))(actual)
+      assertResult(Status.Ok)(status)
+    }
```
jsancio: We should add newline before and after the `scenario`. I see that the rest of the code doesn't do this. You can leave it as is or fix the rest of the code.

PR #661, opened 2017-07-15
Enable finagle-stats

jsancio:


jsancio:
This is ready for review.

project/Deps.scala
```diff
@@ -110,8 +110,9 @@ object Deps {
   )
 
   // APLv2.0
-  val twitterUtilCore = Seq(
-    "com.twitter" %% "util-core" % V.twitterUtilCore
+  val twitterUtil = Seq(
+    "com.twitter" %% "util-core" % V.twitterUtil,
+    "com.twitter" %% "finagle-stats" % "6.40.0"
```
cruhland: Why is the version number different from `V.twitterUtil`? Should we update that to `6.40.0` also, or does @takirala's PR take care of it?

cruhland:
@jsancio is there any configuration needed to get the stats to show up? Also, documenting how to find and interpret the stats in the README would be useful.

project/Deps.scala
takirala: @jsancio FYI, my PR uses the twitterUtil as `6.45.0`

jsancio:
Since this PR is not based on top of our dependency update. I am going to update it to match those version and I will update the README.

jsancio:
Hey @cruhland @takirala @larioj, I updated the PR. I won't cherrypick this to `0.4.x` and instead will create a different commit.

README.md
```diff
@@ -91,6 +91,15 @@ java -jar cosmos-server/target/scala-2.11/cosmos-server_2.11-<version>-SNAPSHOT-
      -com.mesosphere.cosmos.dcosUri <dcos-host-url>
 ```
 
+### Debug a running Cosmos
```
takirala: @cruhland and I had a discussion before and I think we expose a port for remote debugging of cosmos by default right? If that is correct, then the name for this section as `Debug ....` may be misleading.

cruhland: Yeah maybe "View metrics" is more clear?

jsancio: Done. Changed to "Cosmos Admin Portal" since it contains more than just metrics.

cruhland:
:ship: it!

PR #662, opened 2017-07-17
Removed io.github.benwhitehead flags from cosmos repo

takirala:
Removed the package.scala file in io.github.benwhitehead package folder
Updated package.scala in com.mesosphere.cosmos folder


cosmos-server/src/main/scala/com/mesosphere/cosmos/package.scala
```diff
@@ -15,27 +15,19 @@ package object cosmos {
   implicit val globalTimer: Timer = new ScheduledThreadPoolTimer()
 
   def getHttpInterface: Option[InetSocketAddress] = {
-    httpInterface().orElse(
-      _root_.io.github.benwhitehead.finch.httpInterface()
-    )
+    httpInterface()
```
jsancio: Let's get rid of all of this helper methods. In general we should avoid one line methods.

takirala: @jsancio So should I get rid of 
```scala
package object cosmos {
.
...helper methods here
.
}
```
entirely and call the methods directly from wherever these methods are called?

jsancio: Yes. Remove `getHttpInterface`, `getHttpsInterface`, `getCertificatePath` and `getKeyPath`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -353,13 +353,13 @@ object CosmosApp {
...
-    getHttpsInterface.map { interface =>
+    httpsInterface.getWithDefault.map { interface =>
       Http
         .server
         .configured(Label("https"))
         .withTransport.tls(
-          getCertificatePath.get.toString,
-          getKeyPath.get.toString,
+          certificatePath.get.toString,
+          keyPath.get.toString,
```
cruhland: Shouldn't these also use `getWithDefault`?

takirala: @cruhland Yes my bad! However as we are not handling the None case, can I change it `cerificationPath.toString` and `keyPath.toString` ?

cruhland: @takirala using `.toString` directly on a `GlobalFlag` instance will return [the value rendered as a command-line flag](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/Flag.scala#L250-L259); that isn't what we want.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -353,13 +353,13 @@ object CosmosApp {
...
-    getHttpsInterface.map { interface =>
+    httpsInterface.getWithDefault.map { interface =>
       Http
         .server
         .configured(Label("https"))
         .withTransport.tls(
-          getCertificatePath.get.toString,
-          getKeyPath.get.toString,
+          certificatePath.getWithDefault.get.toString,
+          keyPath.getWithDefault.get.toString,
```
cruhland: Actually, since both of these flags have type `GlobalFlag[Path]` (instead of `GlobalFlag[Option[Path]]`), and neither has a default value, we can use their `apply()` methods directly:

```scala
certificatePath().toString,
keyPath().toString
```

PR #666, opened 2017-07-27
Reimplement MultiRepository

takirala:
- Removed PackageCollection trait and its implementations - CosmosRepository and MultiRepository
- Added new stand alone class called PackageCollection that uses RepositoryCache for purposes of caching
- Classes Removed/Added : 
```
    - CosmosRepository.scala
    - MultiRepository.scala
    - PackageCollection.scala (was a trait)
    - MultiRepositorySpec.scala
    - CosmosRepositorySpec.scala
    - PackageSearchHandlerSpec.scala
    + PackageCollection.scala (is a standalone class)
    + PackageCollectionSpec.scala (unit test cases with no wiring. tests only pure functions)
    + RepositoryCache.scala
```
- Added new unit test cases.
- Will add more IT/unit test cases if needed.


cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,310 @@
...
-trait PackageCollection {
+final class PackageCollection(repositoryCache: RepositoryCache) {
 
   def getPackagesByPackageName(
     packageName: String
-  )(implicit session: RequestSession): Future[List[universe.v4.model.PackageDefinition]]
+  )(implicit session: RequestSession): Future[List[universe.v4.model.PackageDefinition]] = {
+    val t = repositoryCache.all()
+    t.map { repositories =>
+      val p = PackageCollection.merge(repositories)
```
cruhland: Let's use more descriptive names here than `t` or `p`.

takirala: I looked over this part of code. Should've cleaned up after debugging.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,305 @@
...
+    }
+  }
 
   /**
-   * Return the versions of packages in this collection that the package with the given `name` and
-   * `version` can be upgraded to.
-   */
+    * Return the versions of packages in this collection that the package with the given `name` and
+    * `version` can be upgraded to.
+    */
```
cruhland: Looks like your editor auto-formatted this comment in "Scala style". We prefer using "Java style" formatting for multi-line comments, could you change it back?

cruhland: There should be an IntelliJ setting to use the Java format.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,305 @@
...
     packageDefinition: universe.v4.model.PackageDefinition
-  )(implicit session: RequestSession): Future[List[universe.v3.model.Version]]
-
+  )(implicit session: RequestSession): Future[List[universe.v3.model.Version]] = {
+    repositoryCache.all().map { repositories =>
+      PackageCollection.downgradesTo(
+        packageDefinition,
+        PackageCollection.merge(repositories)
+      )
+    }
```
cruhland: All of the above methods call `repositoryCache.all()`, followed by `PackageCollection.merge()` or `PackageCollection.mergeWithURI()`. What if it was `RepositoryCache`'s responsibility to perform the merge? You could have a method `repositoryCache.merged()` that returns a cached `List[(universe.v4.model.PackageDefinition, Uri)]`, to avoid recomputing it in each of the above methods.

You'd have to update the methods in `object PackageCollection` to take the full `List[(universe.v4.model.PackageDefinition, Uri)]`, but it's simple to do a `map()` in each of them as a pre-processing step. Or I suppose `RepositoryCache` could have both a `merged()` and a `mergedWithUri()` method.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,305 @@
...
+    packageName: String,
+    packageVersion: Option[universe.v3.model.Version],
+    packageDefinitions: List[(universe.v4.model.PackageDefinition, Uri)]
+  ): (universe.v4.model.PackageDefinition, Uri) = {
+
+    val ns = packageDefinitions.filter { case (packageDefinition, _) =>
+      packageDefinition.name == packageName
+    }
+
+    val vs = packageVersion match {
```
cruhland: More descriptive names than `ns` and `vs`, please?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
takirala: @cruhland @jsancio The original idea behind this was to keep the caching logic separate from the sorting/merging logic. We are using a trivial `Map` with a manual eviction policy currently. With the idea that this can be changed if needed in future to use some advanced caching mechanism, I wanted to keep it separate from the nuances of merging/sorting. The unit test cases also test the pure function merge, but I suppose that can be moved to RepositoryCache object. I agree this is a code duplication but it comes at the cost of keeping the caching logic in `RepositoryCache` and the sorting logic in `PackageCollection` separated. What do you think ?

cruhland: I agree that it's a good idea to keep the merge logic separate from the caching logic. Consider my original comment to be a suggestion; it's not necessary to implement in this PR but could be something we do later as a cleanup/optimization.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/SelectedPackageSpec.scala
```diff
@@ -65,7 +68,7 @@ final class SelectedPackageSpec extends FreeSpec with MockitoSugar {
 
       assertResult(SearchResponse(resultsAfterSorting)) {
         Await.result(handler(SearchRequest(None)))
-      }
+      }*/
```
cruhland: Dead code?

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,526 @@
-package com.mesosphere.cosmos.repository
+
```
larioj: Random newline

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,306 @@
...
+      }
+      .toList
+      .sortBy(p => (!p.selected.getOrElse(false), p.name))
+  }
+
   def upgradesTo(
     name: String,
     version: universe.v3.model.Version,
-    possibleUpgrades: List[universe.v4.model.PackageDefinition]
+    packageDefinitions: List[universe.v4.model.PackageDefinition]
```
larioj: Why the name change?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,306 @@
...
+        if name == packageDefinition.name && packageDefinition.canUpgradeFrom(version) =>
+        packageDefinition.version
     }
   }
 
   def downgradesTo(
-    packageDefinition: universe.v4.model.PackageDefinition,
-    possibleDowngrades: List[universe.v4.model.PackageDefinition]
+    pkgDefinition: universe.v4.model.PackageDefinition,
+    packageDefinitions: List[universe.v4.model.PackageDefinition]
```
larioj: Why the name change?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
takirala: `PackageCollection` was upgraded (or rather downgraded?) from a trait to a standalone class. Previously, only `upgradesTo` and `downgradesTo` were the only two concrete methods implemented. Now, this class has all the methods implemented and all pure functions in the companion object have the first parameter as `packageDefinitions: List[PackageDefintion]`

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,287 @@
...
+      val orderName = pkgDef1.name.compare(pkgDef2.name)
+      val orderIndex = index1.compare(index2)
+
+      if (orderName != 0) {
+        orderName
+      } else if (orderIndex != 0) {
+        orderIndex
+      } else {
+        pkgDef2.releaseVersion.value.compare(pkgDef1.releaseVersion.value)
+      }
```
jsancio: This doesn't look correct. I think you want to negate this comparison.

Index A should be first in the list if `A < B` while `releaseVersion` C should be first in the list if `C > D`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
```diff
@@ -0,0 +1,79 @@
...
+    implicit session: RequestSession
+  ): Future[List[(universe.v4.model.Repository, Uri)]] = {
+    packageRepositoryStorage.readCache().flatMap { packageRepositories =>
+      val oldCachedRepos = cachedRepos
+      update(oldCachedRepos, packageRepositories).onSuccess { newRepositories =>
+        cachedRepos = newRepositories
+      } map { newRepositories =>
+        newRepositories.map { case (uri, (repo, _)) =>
+          (repo, uri)
+        }.toList
```
jsancio: We need to revisit this. Order is lost. We need to preserve the order of `packageRepositoryStorage.readCache()`. See my comment in `update`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
```diff
@@ -0,0 +1,79 @@
...
+      }
+    }
+  }
+
+  private[this] def update(
+    oldMap: Map[Uri, (universe.v4.model.Repository, Long)],
+    packageRepositories: List[rpc.v1.model.PackageRepository]
+  )(
+    implicit session: RequestSession
+  ): Future[Map[Uri, (universe.v4.model.Repository, Long)]] = {
```
jsancio: This type is not rich enough. It doesn't encode the information we need. We need to know the order of `universe.v4.model.Repository`. The order of `universe.v4.model.Repository` has to be the same as `packageRepositories: List[rpc.v1.model.PackageRepository]` in the argument.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
```diff
@@ -0,0 +1,79 @@
...
+    Future.traverseSequentially(packageRepositories) { packageRepository =>
+      oldMap.get(packageRepository.uri) match {
+        case Some((repo, timeStamp)) =>
+          fetch(packageRepository, timeStamp, repo).map(value => (packageRepository.uri, value))
+        case None =>
+          fetch(packageRepository).map(value => (packageRepository.uri, value))
+      }
+    } map(_.toMap)
```
jsancio: I think at this point we have the right order. `_.toMap` destroys that order.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
```diff
@@ -0,0 +1,79 @@
...
+  )(
+    implicit session: RequestSession
+  ): Future[(universe.v4.model.Repository, Long)] = {
+    val now = TimeUnit.MILLISECONDS.toSeconds(clock.nowMillis())
+    val lastSec = TimeUnit.MILLISECONDS.toSeconds(lastTimeStamp)
+    val refetchAt = lastSec + TimeUnit.MINUTES.toSeconds(1)
+    if(refetchAt < now || lastSec > now) {
+      universeClient(packageRepository).map { newRepository =>
+        (newRepository, clock.nowMillis())
+      }
```
jsancio: Let's just call `fetch(packageRepository)`. I think that is equivalent, right?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,287 @@
...
+    pkgDefinition: universe.v4.model.PackageDefinition
   ): List[universe.v3.model.Version] = {
-    possibleDowngrades.collect {
-      case possibleDowngrade
-        if packageDefinition.name == possibleDowngrade.name &&
-          packageDefinition.canDowngradeTo(possibleDowngrade.version) => possibleDowngrade.version
+    packageDefinitions.collect {
+      case packageDefinition
+        if pkgDefinition.name == packageDefinition.name &&
+        pkgDefinition.canDowngradeTo(packageDefinition.version) =>
```
larioj: This method is a lot less readable with the name change. From the reader's perspective there is no difference between `packageDefinition.canDowngradeTo(pkgDefinition)` and  `pkgDefinition.canDowngradeTo(packageDefinition)`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,287 @@
...
+  }
 
   /**
-   * Return the versions of packages in this collection that the package with the given `name` and
-   * `version` can be upgraded to.
-   */
+  * Return the versions of packages in this collection that the package with the given `name` and
+  * `version` can be upgraded to.
+  */
+
```
cruhland: Something's still off with the formatting of this comment. It shouldn't appear in the diff.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,287 @@
...
+    }
+  }
 
   /**
-    * Return the versions of packages in this collection that the package
-    * `packageDefinition` can be downgrade to.
-    */
+  * Return the versions of packages in this collection that the package
+  * `packageDefinition` can be downgrade to.
+  */
```
cruhland: Ah, this comment was done in Scala style but it looks like two spaces were deleted from each line instead of just one. The `*`s should line up under the first `*` in the top line.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,287 @@
...
+    repositories: List[(universe.v4.model.Repository, Uri)]
+  ) : List[universe.v4.model.PackageDefinition] = {
+    mergeWithURI(repositories).map { case (packageDefinition, _) =>
+      packageDefinition
     }
   }
 
+  def createRegex(query: String): Regex = {
+    s"""^${safePattern(query)}$$""".r
+  }
```
cruhland: Move this to be directly above the `getPredicate()` method, where it's used.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,287 @@
...
+            pkg.tags,
+            pkg.selected,
+            pkg.images
+          ))
+        val releaseVersion = searchResult.versions.get(pkg.version).map { releaseVersion =>
+          import universe.v3.model.ReleaseVersion
+          implicitly[Ordering[ReleaseVersion]].max(releaseVersion, pkg.releaseVersion)
+        } getOrElse {
+          pkg.releaseVersion
+        }
```
cruhland: I know the above code was just copied from `CosmosRepository`, but here's a suggestion to make it more readable:

```scala
import scala.math.Ordering.Implicits.infixOrderingOps  // At top of file

val releaseVersion = searchResult.versions
  .getOrElse(pkg.version, pkg.releaseVersion)
  .max(pkg.releaseVersion)
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,287 @@
...
+  def search(
+    packageDefinitions: List[universe.v4.model.PackageDefinition],
+    query: Option[String]
+  ): List[rpc.v1.model.SearchResult] = {
+    val predicate = getPredicate(query)
+
+    val searchResults = packageDefinitions.foldLeft(Map.empty[String, rpc.v1.model.SearchResult]) {
+      (state, pkg) =>
+        val searchResult = state
+          .getOrElse(pkg.name, rpc.v1.model.SearchResult(
```
cruhland: This multi-line initialization of `SearchResult` is somewhat distracting. What if we factored it out into its own method? Then this line could be something like:

```scala
val searchResult = state.getOrElse(pkg.name, singleResult(pkg))
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageSearchHandler.scala
```diff
@@ -10,13 +10,11 @@ private[cosmos] final class PackageSearchHandler(
...
-  ): Future[rpc.v1.model.SearchResponse] = {
-    packageCache.search(request.query) map { packages =>
-      val sortedPackages = packages.sortBy(p => (!p.selected.getOrElse(false), p.name.toLowerCase))
-      rpc.v1.model.SearchResponse(sortedPackages)
-    }
-
+  override def apply(
+    request: rpc.v1.model.SearchRequest
+  )(implicit session: RequestSession): Future[rpc.v1.model.SearchResponse] = {
+    packageCache
```
larioj: Can you change the name of this variable to packageCollection? It does not need to leak the fact that it is a cache.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  private[this] def update(
+    oldMap: Map[Uri, (universe.v4.model.Repository, Long)],
+    packageRepositories: List[rpc.v1.model.PackageRepository]
+  )(
+    implicit session: RequestSession
+  ): Future[Map[Uri, (universe.v4.model.Repository, Long)]] = {
+    Future.traverseSequentially(packageRepositories) { packageRepository =>
```
larioj: Why do we do this sequentially vs. in parallel with Future.collect?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  private[this] def update(
+    oldMap: Map[Uri, (universe.v4.model.Repository, Long)],
+    packageRepositories: List[rpc.v1.model.PackageRepository]
+  )(
+    implicit session: RequestSession
+  ): Future[Seq[(Uri, (universe.v4.model.Repository, Long))]] = {
+    Future.traverseSequentially(packageRepositories) { packageRepository =>
```
larioj: Why do we do this sequentially vs. in parallel with Future.collect?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
```diff
@@ -0,0 +1,79 @@
...
+import scala.collection.breakOut
+
+final class RepositoryCache(
+  packageRepositoryStorage: PackageSourcesStorage,
+  universeClient: UniverseClient,
+  clock: Clock = Clock.SYSTEM_CLOCK
+){
+
+  @volatile private[this] var cachedRepos =
+    Map.empty[Uri, (universe.v4.model.Repository, Long)]
```
larioj: This could just be an association list. In should not be very long.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
```diff
@@ -0,0 +1,79 @@
...
+    packageRepository: rpc.v1.model.PackageRepository,
+    lastTimeStamp: Long,
+    oldRepository: universe.v4.model.Repository
+  )(
+    implicit session: RequestSession
+  ): Future[(universe.v4.model.Repository, Long)] = {
+    val now = TimeUnit.MILLISECONDS.toSeconds(clock.nowMillis())
+    val lastSec = TimeUnit.MILLISECONDS.toSeconds(lastTimeStamp)
+    val refetchAt = lastSec + TimeUnit.MINUTES.toSeconds(1)
+    if(refetchAt < now || lastSec > now) {
```
larioj: Why the `lastSec > now`? Shouldn't lastSec alway be less than now?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  def all()(
+    implicit session: RequestSession
+  ): Future[List[(universe.v4.model.Repository, Uri)]] = {
+    packageRepositoryStorage.readCache().flatMap { packageRepositories =>
+      val oldCachedRepos = cachedRepos
+      update(oldCachedRepos, packageRepositories).onSuccess { newRepositories =>
+        cachedRepos = newRepositories.toMap
+      } map { newRepositories =>
+        val result: List[(universe.v4.model.Repository, Uri)] = newRepositories.map {
```
larioj: What is the difference between this and .toList?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
takirala: I agree. It was like that before, so I just kept it the same way. May be some clock inconsistency in remote machines ? @jsancio @cruhland should I get rid of this check? There is no harm in doing this anyway.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
takirala: Yes It can be, but I still have to convert to map / check the entire list in the update method. So I used a map as it can be more intuitive as a cache.

cruhland: @larioj did you have a specific reason for wanting to use an association list? If it's purely for performance reasons I'd say that's a micro-optimization for which we don't have any supporting data.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
cruhland: Sure, let's get rid of it.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
larioj: You can do something like

```scala
    Future.collect(
      packageRepositories.map { packageRepository =>
        oldMap.get(packageRepository.uri) match {
          case Some((repo, timeStamp)) =>
            fetch(packageRepository, timeStamp, repo).map(value => (packageRepository.uri, value))
          case None =>
            fetch(packageRepository).map(value => (packageRepository.uri, value))
        }
      }
    )
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -1,77 +1,287 @@
...
+          pkg.releaseVersion
+        }
+        val newSearchResult = searchResult.copy(
+          versions=searchResult.versions + ((pkg.version, releaseVersion))
+        )
+        state + ((pkg.name, newSearchResult))
+    }
+      .toList
+      .map { case (_, searchResult) => searchResult }
+      .filter(predicate)
```
cruhland: This can be done with the `.values` method:

```scala
val searchResults = packageDefinitions.foldLeft(...) { (state, pkg) =>
    // ...
}
.values
.filter(predicate)
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
larioj: @cruhland @jsancio made a point that having this be a Map loses the ordering of the repositories. So the all() method does not return a list in the correct priority order. In essence we get the priority ordered list of repos from zk, turn it into a Map (losing ordering), and then turn it back into a list for the return type of the all() method.

larioj: I think we want the all() method to return Repos in the correct order. No?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
takirala: Using `toList` will initially build the Sequence and then converts it to list. breakOut gives out a `CanBuildForm` implicit which helps build the result in form of List instead of creating and then transforming. @jsancio feel free to correct/add details :)

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
jsancio: That is correct @larioj. @takirala has a solution to this and will update it in the next commit. Let's continue the conversation after that commit.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
jsancio: That is correct. It is usually code smell if you have code that does `aCollection.map(...).to<Type>`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -36,42 +55,216 @@ trait PackageCollection {
...
+          (
+            uniquePackageCoordinate + packageCoordinate,
+            uniquePackages :+ current
+          )
+        }
+      }
+
+    val packageOrdering = Ordering.Tuple2(
+      implicitly(Ordering[(String, Int)]),
+      implicitly(Ordering[universe.v3.model.ReleaseVersion].reverse)
```
cruhland: Oh, I didn't catch this earlier...`Ordering.apply` works similarly to `implicitly`. What I had intended for you to write here was (notice the use of square brackets instead of parentheses):

```scala
implicitly[Ordering[(String, Int)]],
implicitly[Ordering[universe.v3.model.ReleaseVersion]].reverse
```

But the following still works and is even shorter:

```scala
Ordering[(String, Int)],
Ordering[universe.v3.model.ReleaseVersion].reverse
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -36,42 +55,216 @@ trait PackageCollection {
...
+          state
+        } else {
+          (
+            uniquePackageCoordinate + packageCoordinate,
+            uniquePackages :+ current
+          )
+        }
+      }
+
+    val packageOrdering = Ordering.Tuple2(
```
takirala: Changed the sorting comparator - credits to Charles (implicitly conveyed to Martin Odersky)

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -36,42 +53,216 @@ trait PackageCollection {
...
+    repositories: List[(universe.v4.model.Repository, Uri)]
+  ) : List[(universe.v4.model.PackageDefinition, Uri)] = {
+    val (_, uniquePackageDefinitions) = repositories
+      .zipWithIndex
+      .flatMap { case ((repository, uri), index) =>
+        repository.packages.map { packageDefinition =>
+          ((packageDefinition, uri), index)
+        }
+      }
+      .foldLeft(
```
cruhland: Fold operations are very powerful, and they can be hard to reason about, because they are similar to arbitrary loops in imperative languages. What if we replaced this `foldLeft` with the following?

```scala
.groupBy { case (packageDefinition, _), _) => packageDefinition.packageCoordinate }
.values
.map { similarPackages =>
  similarPackages.minBy { case (_, index) => index }
}
```

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,563 @@
...
+import com.twitter.util.Throw
+import com.twitter.util.Try
 import org.scalacheck.Gen
 import org.scalatest.FreeSpec
+import org.scalatest.Matchers
 import org.scalatest.prop.PropertyChecks
+import org.scalatest.prop.TableDrivenPropertyChecks
 
-final class PackageCollectionSpec extends FreeSpec with PropertyChecks {
+final class PackageCollectionSpec extends FreeSpec
```
cruhland: @takirala could you point out which tests are new in this file? And for tests that previously existed in other files, could you describe what you changed about them (if anything)?

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/CosmosRepositorySpec.scala
```diff
@@ -1,318 +0,0 @@
...
-    "by tag" in {
-      val u = Uri.parse("/uri")
-      val rep = C.rpc.v1.model.PackageRepository("test", u)
-      val l = List(TestingPackages.MaximalV3ModelV3PackageDefinition)
-      val c = CosmosRepository(rep, client(l))
-      assertResult("MAXIMAL")(Try(Await.result(c.search(Some("all")))).get.head.name)
-      assertResult("MAXIMAL")(Try(Await.result(c.search(Some("thing*")))).get.head.name)
-    }
-  }
-  "timeout repository" in {
```
takirala: @cruhland This was the test case that I got rid of. FYI.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
takirala: As discussed offline, i built up the scenarios by an amalgamation of tests from MultiRepoSpec CosmosRepoSpec and PackageCollectionSpec. There was a lot of wiring involved and also test cases were dispersed all over. That being said, may be focusing on the `merge` and `search` would be good starting point as these methods have been refactored heavily. Also, scenarios like `"upgradesTo()"` and `createRegex` were testing pure functions before as well, so there is nothing much that changed in these two cases.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -36,42 +53,202 @@ trait PackageCollection {
...
+    val uniquePackageDefinitions = repositories
+      .zipWithIndex
+      .flatMap { case ((repository, uri), index) =>
+        repository.packages.map { packageDefinition =>
+          ((packageDefinition, uri), index)
+        }
+      }
+      .groupBy { case ((packageDefinition, _), _) => packageDefinition.packageCoordinate }
+      .values
+      .map { similarPackages => similarPackages.minBy { case (_, index) => index } }
```
jsancio: Two packages can have the same name, version and index. At this point in the algorithm (name, index, releaseVersion) is unique.

jsancio: In other words if two packages have the name `packageCoordinate` pick the smallest of (index, -releaseVersion).

cruhland: And make sure to add a test for this case!

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,563 @@
...
+            TestingPackages.MinimalV4ModelPackageDefinitionV4
+          ), Uri.parse("/minimal"))
+        )
+        assertResult(List(
+          TestingPackages.MaximalV3ModelV2PackageDefinition,
+          TestingPackages.MaximalV4ModelPackageDefinitionV4,
+          TestingPackages.HelloWorldV3Package,
+          TestingPackages.MinimalV3ModelV2PackageDefinition,
+          TestingPackages.MinimalV4ModelPackageDefinitionV4
+        ))(PackageCollection.merge(repositories))
```
jsancio: I think we need more test cases. These are some of the scenarios I want us to test. Let's write them in a way that it is trivial to see what they are testing. For example, in the example above I have to look at the details in `MinimalV3..`, `MinimalV4...`, `MaximalV3...`, etc.

Input:
repo0: (name1, version1, releaseVersion2), (name1, version1, releaseVersion1)
Output:
(name1, version1, releaseVersion2)

Input:
repo0: (name1, version1, releaseVersion1)
repo1: (name1, version1, releaseVersion2)
Output:
(name1, version1, releaseVersion1)

Input:
repo0: (name1, version1, releaseVersion2)
repo1: (name1, version1, releaseVersion1)
Output:
(name1, version1, releaseVersion2)

Input:
repo0: (name1, version1, releaseVersion1)
repo1: (name2, version1, releaseVersion1)
Output:
(name1, version1, releaseVersion1), (name2, version1, releaseVersion1)

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,563 @@
...
+          PackageCollection.getPackagesByPackageName(
+            List.empty[universe.v4.model.PackageDefinition],
+            "test"
+          )
+        ) shouldBe Throw(
+          PackageNotFound("test").exception
+        )
+      }
+
+      "invalid package name should throw" in {
```
jsancio: This text doesn't match what the test is doing. Maybe "name not found in non-empt collection".

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,563 @@
...
+      }
+
+      "found minimal" in {
+        val cls = List(TestingPackages.MinimalV3ModelV2PackageDefinition)
+        //val ver = TestingPackages.MinimalV3ModelV2PackageDefinition.version
+
+        Try(PackageCollection.getPackagesByPackageName(cls, "minimal")) shouldBe Return(cls)
+        Try(PackageCollection.getPackagesByPackageName(cls, "MAXIMAL")) shouldBe Throw(
+          PackageNotFound("MAXIMAL").exception
+        )
```
jsancio: Remove this check. I think you are doing this in the test above, no?

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,563 @@
...
+        assertResult(
+          Return(List(TestingPackages.MaximalV3ModelV3PackageDefinition))
+        )(
+          Try(PackageCollection.getPackagesByPackageName(packages, "MAXIMAL"))
+        )
+
+        Try(PackageCollection.getPackagesByPackageName(packages, "test")) shouldBe(
+          Throw(PackageNotFound("test").exception)
+        )
+      }
```
jsancio: I am not sure what this is testing that we are not already testing.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,563 @@
...
+        )(Try(PackageCollection.search(clientdata, Some("minimal")).map(_.name)))
+
+        assertResult(
+          Return(List(Set(minver, min2ver)))
+        )(Try(PackageCollection.search(clientdata, Some("minimal")).map(_.versions.keys)))
+
+        assertResult(
+          Return(List(Set(maxver, max2ver)))
+        )(Try(PackageCollection.search(clientdata, Some("MAXIMAL")).map(_.versions.keys)))
+      }
```
jsancio: @takirala These tests are difficult to read because of these "minimal" and "maximal" packages. Excluding `search` the important information for all of this algorithms is name, version, repo index and release version. Let's create methods that create packages (`PackageDefinition`) based on that information.

From there we can focus on three sets of tests. 1) tests that verify our merging logic specifically `mergeWithUri`. 2) tests that assume that merge is correct and verify the rest of the functions. 3) test the search functionality; here we can use the minimal and maximal packages or create different packages that are better for testing search.

What do you think?

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,547 @@
...
+        )(PackageCollection.merge(List(
+          getRepository(List(a_88_1, a_99_3, b_66_1, b_77_2, c_44_1, c_55_2))
+        )))
+
+        assertResult(
+          List(a_99_2)
+        )(PackageCollection.merge(List(
+          getRepository(List(a_99_2)),
+          getRepository(List(a_99_3))
+        )))
```
jsancio: I like the test but it doesn't match the description above. This is more like "should sort by the smallest index before release"

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,547 @@
...
+
+        assertResult(
+          List(a_99_3)
+        )(PackageCollection.merge(List(
+          getRepository(List(a_99_3)),
+          getRepository(List(a_99_2))
+        )))
```
jsancio: I think you have a more complicated version of this test in "should sort packages by repository index" above.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,547 @@
...
+        ) shouldBe Throw(
+          PackageNotFound("test").exception
+        )
 
-        forAll (genParameters) { case (name, version, packages) =>
-          assert(PackageCollection.upgradesTo(name, version, packages).isEmpty)
+        Try(
+          PackageCollection.getPackagesByPackageVersion(
+            List((a_99_2, Uri.parse("/test")), (a_99_3, Uri.parse("/test"))),
+            a_99_2.name,
```
jsancio: Making this `a_88_1.name` may make it more obvious what you are trying to do.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,547 @@
...
+      "works for packages with same name" in {
+        val sameNamePackages = List(
+          (b_66_1, t0),
+          (b_77_2, t2),
+          (b_99_2, t1)
+        )
+
+        Try(PackageCollection.getPackagesByPackageVersion(
+          sameNamePackages,
+          b_99_2.name,
```
jsancio: Make this `b_77_2.name` to make the test more obvious.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
takirala: @jsancio I have rewrote the test cases to be more readable. I have not changed the search test case. This is ready for your review.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,547 @@
...
+        val all = List(TestingPackages.MaximalV3ModelV3PackageDefinition,
+          TestingPackages.MinimalV3ModelV2PackageDefinition)
+
+        Try(PackageCollection.search(all, None).map(_.name)) shouldBe
+          Return(List("MAXIMAL","minimal"))
+
+        Try(PackageCollection.search(all, Some("minimal")).map(_.name)) shouldBe
+          Return(List("minimal"))
+
+        assertResult(Return(2))(Try(PackageCollection.search(all, None).length))
```
jsancio: You can delete this. You have a better test above in line 400.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -36,42 +53,204 @@ trait PackageCollection {
...
+    }
+  }
+
+  /**
+   *
+   * The merge should remove all the packages that has same name+version value (with lowest index
+   * value staying) and then should sort the rest initially by name and then index and then releaseVersion
+   *
+   * The sorting should use foldLeft as we need to iterate from left to right to preserve the order.
+   * Elements are appended at the end of Sequence and thus Vector is a better choice than List.
```
cruhland: This part of the comment is out of date!

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
cruhland: @takirala is this getting updated?

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/CosmosRepositorySpec.scala
```diff
@@ -1,318 +0,0 @@
...
-      def waitFor(x:Long): Unit = sys.error("unexpected")
-    }
-    val c = CosmosRepository(rep, client(b, a), clock)
-    assertResult("minimal")(Try(Await.result(c.search(Some("minimal")))).get.head.name)
-    assertResult(Return(Nil))(Try(Await.result(c.search(Some("MAXIMAL")))))
-
-    millis = millis + 60*1000*1000
-    assertResult("MAXIMAL")(Try(Await.result(c.search(Some("MAXIMAL")))).get.head.name)
-
-    //timer wrap-around
```
cruhland: @takirala I think this is why there was that extra condition in the `fetch()` method of `RepositoryCache`.

cruhland: However, considering that the timer value is storing seconds in a `long` variable, we'll never need to worry about wrap-around (how many of our customers will be around in 292 billion years? :smile:).

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/CosmosRepositorySpec.scala
cruhland: What specifically made you delete this test? It looks like you could adapt it to use the new `PackageCollection`.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,547 @@
...
 
-final class PackageCollectionSpec extends FreeSpec with PropertyChecks {
+final class PackageCollectionSpec extends FreeSpec
+  with Matchers
+  with PropertyChecks
+  with TableDrivenPropertyChecks {
 
-  import PackageCollectionSpec._
+  def getRepository(
+    packageDefinitions : List[universe.v4.model.PackageDefinition] = List.empty[universe.v4.model.PackageDefinition],
```
cruhland: The default argument is only used in one place. That's not really enough to justify having it -- it would be more clear to always pass the argument explicitly.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,547 @@
...
+          a_99_2,
+          b_66_1
+        )
+        assertResult(List(a_99_2, b_66_1))(PackageCollection.merge(List(getRepository(packages))))
+      }
+
+      "should remove packages with same coordinates" in {
+        val repositories = List(
+          getRepository(List(a_88_1, a_99_2)),
+          getRepository(List(a_88_1, b_66_1))
```
cruhland: Can you change this to use two packages with the same coordinates but different releaseVersions (and indices)? If you did that then you ensure `merge` *must* only use the coordinate to de-duplicate the packages.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,547 @@
...
+
+        Try(PackageCollection.getPackagesByPackageVersion(
+          sameNamePackages,
+          b_99_2.name,
+          Some(b_77_2.version)
+        )) shouldBe Return((b_77_2, t2))
+
+        Try(PackageCollection.getPackagesByPackageVersion(
+          sameNamePackages,
+          b_99_2.name,
```
cruhland: Same as above, use `b_66_1.name` here.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/repository/PackageCollectionSpec.scala
```diff
@@ -1,85 +1,547 @@
...
+    }
+
+    "search" - {
+
+      "not found" in {
+        assertResult(
+          Return(Nil)
+        )(
+          Try(
+            PackageCollection.search(List.empty[universe.v4.model.PackageDefinition], Some("test"))
```
cruhland: Scala should be able to infer the type parameter on `List.empty`, have you tried removing it?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/RepositoryCache.scala
takirala: @cruhland As discussed offline, we are happy with Map as we are not returning the map (converted to list) instead the map is only for storing &searching purposes.

PR #674, opened 2017-08-08
Always complete the promise

takirala:
- Promise is used only in `ZkRepositoryList`
- A constant field is created in zookeeper class for the duration of wait.
- Helper method throws or returns the completed promise.

### DIFF
--- a/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/TimeoutError.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/TimeoutError.scala
@@ -1,17 +1,20 @@
 package com.mesosphere.cosmos.error
 
 
+import com.twitter.util.Duration
 import io.circe.Encoder
 import io.circe.JsonObject
 import io.circe.generic.semiauto.deriveEncoder
+import com.mesosphere.universe.common.circe.Encoders._
 
 final case class TimeoutError(
   operation: String,
   destination: String,
-  timeout: String
+  timeout: Duration
 ) extends CosmosError {
   override def data: Option[JsonObject] = CosmosError.deriveData(this)
-  override def message: String = s"$operation timed out on $destination after $timeout"
+  override def message: String = s"$operation timed out on $destination" +
+    s" after ${timeout.inSeconds} seconds"
 }
 
 object TimeoutError {
--- a/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/TimeoutError.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/TimeoutError.scala
@@ -1,0 +1,19 @@
+package com.mesosphere.cosmos.error
+
+
+import io.circe.Encoder
+import io.circe.JsonObject
+import io.circe.generic.semiauto.deriveEncoder
+
+final case class TimeoutError(
+  operation: String,
+  destination: String,
+  timeout: String
+) extends CosmosError {
+  override def data: Option[JsonObject] = CosmosError.deriveData(this)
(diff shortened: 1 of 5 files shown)


takirala:
This the process I followed to test this : 
 - Spun up a local zookeeper on 2181 port. When cosmos boots up, it connects to this zookeeper.
 - Spun up a cluster in ccm. 
 - Run cosmos in debug mode locally with above dcos cluster. 
 - use dcos cli locally to search for a package
    - Dcos cli gets response for the search result.
    - Shutdown zookeeper and after one minute (when the cache invalidates itself) the search result is an error describing the timeout.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/TimeoutError.scala
```diff
@@ -0,0 +1,19 @@
...
+
+
+import io.circe.Encoder
+import io.circe.JsonObject
+import io.circe.generic.semiauto.deriveEncoder
+
+final case class TimeoutError(
+  operation: String,
+  destination: String,
+  timeout: String
```
jsancio: Make the type for `timeout` a `Duration` and apply the formatting in `def message`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/ZkRepositoryList.scala
```diff
@@ -186,20 +188,30 @@ final class ZkRepositoryList private (
...
-    implicit statsReceiver: StatsReceiver
+    implicit statsReceiver: StatsReceiver, timer: Timer
   ): ZkRepositoryList = {
     val repoList = new ZkRepositoryList(zkClient)
     repoList.start()
     repoList
   }
 
-  private val PackageRepositoriesPath: String = "/package/repositories"
+  private val PackageRepositoriesPath: String = "/cosmos/package/repositories"
```
jsancio: The root path is not always `/cosmos`. Instead it is configured through the `ZooKeeperUri` `GlobalFlag`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/zookeeper/Clients.scala
```diff
@@ -56,20 +55,21 @@ object Clients {
...
   private[this] def updateAcls(
+    zkUri: ZooKeeperUri,
     zkClient: CuratorFramework,
     aclProvider: ACLProvider
   ): Unit = {
     updateAcls(
       zkClient,
       aclProvider,
-      zkClient.getChildren.forPath("/").asScala.toList.map("/" + _)
+      zkClient.getChildren.forPath(zkUri.path).asScala.toList.map(zkUri.path + "/" + _ )
```
jsancio: :+1: 

See my comment in `ZkRepositoryList` but we need to also pass this URI to that class.

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/ZkRepositoryList.scala
```diff
@@ -158,7 +162,7 @@ final class ZkRepositoryList private (
     zkClient.getData().inBackground(
       new ReadHandler(promise)
     ).forPath(
-      ZkRepositoryList.PackageRepositoriesPath
+      zkUri.path + ZkRepositoryList.PackageRepositoriesPath
```
jsancio: Let's make `PackageRepositoriesPath` `private[this]` and add a method that does this computation.

takirala:
@jsancio Updated. is this what you had in mind?

PR #676, opened 2017-08-08
Package resource proxying endpoint

cruhland:
Addresses [DCOS-17584](https://jira.mesosphere.com/browse/DCOS-17584).


cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala
```diff
@@ -146,13 +147,17 @@ with Logging {
       packageRepositoryAdd = standardEndpoint(pkg :: repo :: "add", packageRepositoryAdd),
       packageRepositoryDelete = standardEndpoint(pkg :: repo :: "delete", packageRepositoryDelete),
       packageRepositoryList = standardEndpoint(pkg :: repo :: "list", packageRepositoryList),
+      // TODO proxy Extract URI validation into RequestValidators
```
cruhland: Optional code cleanup: make a function similar to `standardEndpoint` but specifically for this use case -- a query parameter in the endpoint route.

cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
```diff
@@ -0,0 +1,144 @@
...
+    uri: Uri,
+    conn: HttpURLConnection
+  )(implicit
+    sr: StatsReceiver
+  ): (MediaType, Option[String]) = {
+    conn.getResponseCode match {
+      case HttpURLConnection.HTTP_OK =>
+        sr.scope("status").counter("200").incr()
+        // TODO proxy Handle error cases
+        val contentType = MediaTypeParser.parseUnsafe(conn.getHeaderField(Fields.ContentType))
```
cruhland: We should replace `MediaTypeParser.parseUnsafe` with something that generates a useful error if the content type can't be parsed.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandler.scala
```diff
@@ -0,0 +1,81 @@
...
+  contentLengthLimit: StorageUnit,
+  statsReceiver: StatsReceiver
+) {
+
+  import ResourceProxyHandler._
+
+  def apply(uri: Uri): Future[Output[Response]] = {
+    httpClient
+      .fetch(uri, statsReceiver) { responseData =>
+        // TODO proxy May want to factor out a method that can be tested separately
```
cruhland: It might simplify the tests in `ResourceProxyHandlerSpec` if they tested a method factored out from this code block. There would be fewer moving parts.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandler.scala
```diff
@@ -0,0 +1,81 @@
...
+        val response = Response()
+        if (responseData.contentLength.exists(_ >= contentLengthLimit.bytes)) {
+          throw ResourceTooLarge(responseData.contentLength, contentLengthLimit.bytes).exception
+        }
+
+        // TODO proxy Fail if data is too large
+        // Allocate array of min(ContentLength size if defined, limit size - 1)
+        // Buffer InputStream into array; if the end is reached but there's more data, fail
+        // If the data runs out before the end of the array, truncate the array
```
cruhland: There's a bit more logic that needs to be written here; see TODO in `ResourceProxyHandlerSpec` for the expected behavior.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandler.scala
```diff
@@ -0,0 +1,81 @@
...
+          throw ResourceTooLarge(contentLength = None, contentLengthLimit.bytes).exception
+        }
+
+        response.content = Buf.ByteArray.Owned(contentBytes)
+        response.contentType = responseData.contentType.show
+        response
+      }
+      .map { result =>
+        // TODO proxy Handle errors
+        Output.payload(result.right.get)
```
cruhland: `result` is an `Either`. We need to throw `CosmosException`s for each of the errors that could be in `Left`. See `UniverseClient` for an example.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandler.scala
```diff
@@ -0,0 +1,81 @@
...
+      }
+  }
+
+}
+
+object ResourceProxyHandler {
+
+  // TODO proxy Determine what the actual limit should be; maybe use a flag?
+  val DefaultContentLengthLimit: StorageUnit = 5.megabytes
```
cruhland: We need to look at resource files in Universe to determine what the resource size limit should be. It could be hardcoded here, or maybe we should make it a command-line flag for Cosmos.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/HttpClientSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  }
+
+}
+
+object HttpClientSpec {
+
+  val stats: StatsReceiver = NullStatsReceiver
+
+  // TODO proxy Use ItObjects.V4TestUniverse once Jose's PR is merged
```
cruhland: The URL below can be referenced with `ItObjects.V4TestUniverse` now.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/Requests.scala
```diff
@@ -208,6 +209,11 @@ object Requests {
     }
   }
 
+  def packageResource(resourceUri: Uri): Response = {
+    val request = CosmosRequests.packageResource(resourceUri)
+    CosmosClient.submit(request)
+  }
```
jsancio: This should be in a different layer of the integration tests. For now, let's move it to `ResourceProxyHandlerIntegrationSpec` it is the only place where it is used.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandlerIntegrationSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util.Await
+import io.circe.jawn.decode
+import org.scalatest.Matchers
+import org.scalatest.Assertion
+import org.scalatest.FreeSpec
+import org.scalatest.prop.TableDrivenPropertyChecks
+import org.scalatest.prop.TableFor1
+
+final class ResourceProxyHandlerIntegrationSpec extends FreeSpec
```
jsancio: I thought you were going to rename this to `ResourceProxyHandlerSpec` to match the rest of the integration specs.

cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
```diff
@@ -0,0 +1,184 @@
...
+    uri: Uri,
+    statsReceiver: StatsReceiver,
+    headers: (String, String)*
+  )(
+    processResponse: ResponseData => A
+  ): Future[Either[HttpClient.Error, A]]
+
+}
+
+object HttpClient extends HttpClient {
```
jsancio: Can we instead make this a `final class` and get rid of the `trait HttpClient`?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandlerIntegrationSpec.scala
takirala: We already have a `ResourceProxyHandlerSpec` in the tests. Should we still rename this?

cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
jsancio: We should change this https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/BuildProperties.scala#L24 to `extend AnyVal`. E.g.

```scala
implicit class BuildProperties(val ...) extends AnyVal
...
```

cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  def fetch[A](
+    uri: Uri,
+    statsReceiver: StatsReceiver,
+    headers: (String, String)*
+  )(
+    processResponse: ResponseData => A
+  ): Future[Either[Error, A]] = {
```
jsancio: Move `statsReceiver: ...` to here and change to implicit.

takirala: This method is called from universe client and it has two implicit StatsReceiver objects. I guess that was the reason for structuring it this way. FWIW, the stats receiver is declared implicit at the start of this method body and is passed on to helper methods as an implicit.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/ResourceTooLarge.scala
```diff
@@ -0,0 +1,25 @@
...
+  override def message: String = {
+    val lengthText = contentLength.map(_ + " bytes").getOrElse("unknown")
+    s"Resource of $lengthText length has reached or exceeded the limit of $limit bytes"
+  }
+
+  override def data: Option[JsonObject] = CosmosError.deriveData(this)
+
+  override def exception: CosmosException = {
+    CosmosException(this, Status.Forbidden, headers = Map.empty, causedBy = None)
+  }
```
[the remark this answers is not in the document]
takirala: I initially had `5xx` but then since there is nothing wrong with the upstream server, i changed it to [[403](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.4)-The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated. ]

cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
jsancio: Why does UniverseClient have two implicit StatsReceiver? That sounds like a bug to me.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandlerIntegrationSpec.scala
jsancio: Okay.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/ResourceTooLarge.scala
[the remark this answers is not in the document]
jsancio: Okay.

cosmos-test-common/src/main/scala/com/mesosphere/cosmos/http/HttpRequest.scala
```diff
@@ -46,28 +46,35 @@ object HttpRequest {
...
   }
 
   def post(
     path: RpcPath,
     body: Buf,
     contentType: MediaType,
-    accept: MediaType
+    accept: MediaType,
+    host: String,
+    urlScheme: String
```
jsancio: How does the client know what to set for host and urlScheme. I quick grepped through the code and I couldn't find where we are using this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
takirala: I think i misspoke. There is one statsReceiver with [two scopes](https://github.com/dcos/cosmos/pull/676/files#diff-61fe139ee68175f2c567a9cbefc9f859R79) I modified the method signature to accept an implicit. But I still need to pass the variable manually (from the universe client) as there are two implicits (one is the statReceiver passed to universe client and the other is the scope derived from that stat receiver)

cosmos-test-common/src/main/scala/com/mesosphere/cosmos/http/HttpRequest.scala
takirala: This is used [here](https://github.com/dcos/cosmos/blob/master/cosmos-test-common/src/test/scala/com/mesosphere/cosmos/finch/VersionedResponsesSpec.scala#L71)

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandler.scala
```diff
@@ -0,0 +1,161 @@
...
+
+    val length = responseData.contentLength match {
+      case Some(len) => len.toInt
+      case None => contentLengthLimit.bytes.toInt - 1
+    }
+    val contentBytes = Array.ofDim[Byte](length)
+    val bytesRead = bufferFully(responseData.contentStream, contentBytes)
+
+    if (responseData.contentLength.isDefined && bytesRead < contentBytes.length) {
+      throw GenericHttpError(uri = uri, clientStatus = Status.InternalServerError).exception
```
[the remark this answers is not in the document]
takirala: @jsancio Assuming we have to return 5xx, i am changing this to [503](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.5.3)(The server, while acting as a gateway or proxy, received an invalid response from the upstream server it accessed in attempting to fulfill the request) as a response without header is an invalid response.

cosmos-test-common/src/main/scala/com/mesosphere/Generators.scala
```diff
@@ -138,6 +141,57 @@ object Generators {
     }
   }
 
+  def genV3ResourceTestData() : Gen[(collection.immutable.Set[String],
+    universe.v3.model.Assets,
+    universe.v3.model.Images,
+    universe.v3.model.Cli)] = {
```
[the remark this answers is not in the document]
takirala: @jsancio These methods return an `Gen[Set[String],....` where the first parameter is an expected set of urls. I could not extend PackageDefinition if I wanted that. I am moving these methods to `PackageCollectionSpec.scala` to retain that parameter. The `expectedUrls` set is helpful as this method is used multiple times in `PackageCollectionSpec`

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/PackageCollection.scala
```diff
@@ -166,6 +173,23 @@ object PackageCollection {
     }
   }
 
+  def allUrls(repositories: List[(universe.v4.model.Repository, Uri)]): Set[String] = {
+    repositories.flatMap { case (repo, _) =>
+      repo.packages.flatMap {
+        _ match {
+          case v2: universe.v3.model.V2Package =>
+            v2.resource.map(r => r.assets.map(assetsSet) ++ r.images.map(imagesSet))
```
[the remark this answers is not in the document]
takirala: `assetsSet` return an `Set[String]`, how can I use flatMap if it doesn't return an `Option[?]` ?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandler.scala
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    bufferFully(responseData.contentStream, EofDetector) match {
+      case 0 if responseData.contentLength.isEmpty =>
+        contentBytes.dropRight(contentBytes.length - bytesRead)
+      case x if x > 0 =>
+        throw convertToCosmosError(uri, responseData, contentLengthLimit).exception
+      case _ => contentBytes
+    }
```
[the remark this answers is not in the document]
takirala: I ran a test script to check this. All the urls do have a content length header.

PR #677, opened 2017-08-08
Suppress stack trace

takirala:
- Removed full stack trace where an exception is expected.
- Removed duplication of dependency specification.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -60,8 +60,7 @@ lazy val testCommon = project.in(file("cosmos-test-common"))
       Deps.curator ++
       Deps.mockito ++
       Deps.scalaCheck ++
-      Deps.scalaCheckShapeless ++
-      Deps.scalaTest
+      Deps.scalaCheckShapeless
   )
   .dependsOn(
     server
@@ -84,7 +83,7 @@ lazy val integrationTests = project.in(file("cosmos-integration-tests"))
       // No additional properties needed for these tests
       additionalProperties = Nil,
       streamsValue = (streams in runMain).value
-    ),
+    ) ++ Seq(Tests.Argument(TestFrameworks.ScalaTest, "-oS")),
     // Uses (compile in Compile) in addition to (compile in IntegrationTest), the default
     definedTests in IntegrationTest := {
       val frameworkMap = (loadedTestFrameworks in IntegrationTest).value
--- a/build.sbt
+++ b/build.sbt
@@ -38,7 +38,6 @@ lazy val server = project.in(file("cosmos-server"))
     name := baseDirectory.value.name,
     libraryDependencies ++=
       Deps.bijectionUtil ++
-      Deps.circe ++
       Deps.curator ++
       Deps.logback ++
       Deps.mustache ++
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -59,9 +58,7 @@ lazy val testCommon = project.in(file("cosmos-test-common"))
     libraryDependencies ++=
       Deps.curator ++
       Deps.mockito ++
-      Deps.scalaCheck ++
-      Deps.scalaCheckShapeless ++
-      Deps.scalaTest
+      Deps.scalaCheckShapeless
```
takirala: @jsancio These dependencies are already present in the cosmos-common which is used by cosmos-server which is used by cosmos-test-common. So I thought we can get rid of this. Is that correct or do we need to list out all the libraries for each module that it depends on?

jsancio: Sounds good. Dependencies are transitive.

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
```diff
@@ -36,7 +36,9 @@ final class ServiceUninstaller(
     work.rescue {
       case ex if retries > 0  =>
         if (!ex.isInstanceOf[PredicateDoesNotObtain]) {
-          logger.info(s"Uninstall attempt for $appId didn't finish. $retries retries left.")
+          logger.info(s"Uninstall attempt for $appId didn't finish. $retries retries left.",
+            ex.getClass.getSimpleName,
+            ex.getMessage)
```
jsancio: What is this doing? Don't you want to do?
```
logger.info(s"Uninstall attempt for $appId didn't finish. $retries retries left. Type name: ${ex.getClass.getSimpleName}; Message: ${ex.getMessage}")

Comment (takirala) [inline cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala]: Oops! I forgot to add in the parameters. My intention was to do 

```
logger.info(s"Uninstall attempt for $appId didn't finish. $retries retries left. Type name: {}; Message: {}",ex.getClass.getSimpleName, ex.getMessage)
```
So that the arguments won't be evaluated if loglevel is > `INFO`

Now, I think I can do this instead : 
```
logger.info(s"Uninstall attempt for $appId didn't finish. $retries retries left." +
            s" Type name: ${ex.getClass.getSimpleName}; Message: ${ex.getMessage}",
            appId,
            retries,
            ex.getClass.getSimpleName,
            ex.getMessage)
```
This will skip the string concatenation if the loglevel is `INFO`. Shall i do this or just do 
```
logger.info(s"Uninstall attempt for $appId didn't finish. $retries retries left." +
            s" Type name: ${ex.getClass.getSimpleName}; Message: ${ex.getMessage}",
            appId,
            retries,
            ex.getClass.getSimpleName,
            ex.getMessage)
```
thus preferring readability over a minute optimization?

jsancio: I would use Scala's `s""`. We can optimize this stuff later.

PR #681, opened 2017-08-16
Stop uninstall background task if Marathon app vanishes

cruhland:
Partially addresses [DCOS-17237](https://jira.mesosphere.com/browse/DCOS-17237).

Apologies for the delay on this, I thought I might not be able to have a test case for it but I eventually found a scenario that reproduced the issue consistently.

### DIFF
--- a/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandlerSpec.scala
+++ b/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandlerSpec.scala
@@ -22,6 +22,7 @@ import io.circe.Json
 import io.circe.JsonObject
 import io.circe.jawn._
 import java.util.UUID
+import org.scalatest.Assertion
 import org.scalatest.FreeSpec
 import org.scalatest.concurrent.Eventually
 import org.scalatest.time.SpanSugar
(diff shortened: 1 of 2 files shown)


cruhland:
@larioj @takirala I updated this with some test changes that should make it easier to diagnose failures. I also deleted a redundant test.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandlerSpec.scala
```diff
@@ -104,73 +106,96 @@ final class UninstallHandlerSpec extends FreeSpec with Eventually with SpanSugar
...
+  }
+
 }
 
 object UninstallHandlerSpec {
 
+  val HelloWorldPackageName: String = "hello-world"
+  val HelloWorldAppId: AppId = AppId(HelloWorldPackageName)
+
+  // Ensure only one "world" pod is launched, to avoid resource starvation
```
takirala: For my own understanding, what does `world pod is launched, to avoid resource starvation` mean?

cruhland: By default the package launches two pods, but our test clusters don't have enough resources to successfully run both of them -- the install hangs. I added the comment to explain why we have to use the custom config; it wasn't obvious from the code why it was needed. The `"world"` quote just refers to the name of the pod.

PR #685, opened 2017-08-21
Return HTTP 503 instead of 400 for service unavaiblity

takirala:
Change http status code from bad request to service unavailable whenever marathon service is unavailable.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
@@ -113,7 +113,7 @@ final class DefaultUniverseClient(
     Stat.timeFuture(fetchScope.stat("histogram")) {
       Future { repository.uri.toURI.toURL.openConnection() } handle {
         case t @ (_: IllegalArgumentException | _: MalformedURLException | _: URISyntaxException) =>
-          throw CosmosException.apply(RepositoryUriSyntax(repository, t.getMessage), t)
+          throw CosmosException(RepositoryUriSyntax(repository, t.getMessage), t)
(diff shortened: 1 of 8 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -113,7 +113,7 @@ final class DefaultUniverseClient(
     Stat.timeFuture(fetchScope.stat("histogram")) {
       Future { repository.uri.toURI.toURL.openConnection() } handle {
         case t @ (_: IllegalArgumentException | _: MalformedURLException | _: URISyntaxException) =>
-          throw CosmosException(RepositoryUriSyntax(repository, t.getMessage), t)
+          throw CosmosException.apply(RepositoryUriSyntax(repository, t.getMessage), t)
```
takirala: @larioj We can say this is a 400 (Same for L160 I guess?) right?

cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/ZkRepositoryList.scala
```diff
@@ -250,7 +250,7 @@ private final class WriteHandler(
       } else {
         val exception = if (code == KeeperException.Code.BADVERSION) {
           // BADVERSION is expected so let's display a friendlier error
-          CosmosException(ConcurrentAccess(), KeeperException.create(code, event.getPath))
+          CosmosException.apply(ConcurrentAccess(), KeeperException.create(code, event.getPath))
```
takirala: @larioj this is a 400 as well (also for L280) ?

takirala:
@larioj the existing test case had an unreachable statement. I updated the test case.

PR #686, opened 2017-08-24
Update documentation to reflect the actual RPC of ListVersionsRequest

takirala:


docs/schemas/cosmos/rpc/v1/model/ListVersionsRequest-schema.json
```diff
@@ -6,12 +6,12 @@
       "type": "string"
     },
     "includePackageVersions": {
-      "type": "boolean",
-      "default": false
```
larioj: 👍

PR #687, opened 2017-08-25
Avoid uninstalling the app if its Mesos framework ID has changed

cruhland:
Intended to address the remainder of [DCOS-17237](https://jira.mesosphere.com/browse/DCOS-17237).

This implements bullet 2 from that issue; bullet 3 hasn't been done yet.

### DIFF
--- a/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/SDKUninstallVarNotFound.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/cosmos/error/SDKUninstallVarNotFound.scala
@@ -5,11 +5,11 @@ import io.circe.Encoder
 import io.circe.JsonObject
 import io.circe.generic.semiauto.deriveEncoder
 
-final case class UninstallAlreadyQueued(appId: AppId) extends CosmosError {
+final case class SDKUninstallVarNotFound(appId: AppId) extends CosmosError {
   override def data: Option[JsonObject] = CosmosError.deriveData(this)
   override def message: String = s"AppId [$appId] is already queued for uninstall"
 }
 
-object UninstallAlreadyQueued {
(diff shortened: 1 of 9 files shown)


cosmos-common/src/main/scala/com/mesosphere/cosmos/error/UninstallAlreadyQueued.scala
```diff
@@ -0,0 +1,15 @@
+package com.mesosphere.cosmos.error
+
+import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
+import io.circe.Encoder
+import io.circe.JsonObject
+import io.circe.generic.semiauto.deriveEncoder
+
+final case class UninstallAlreadyQueued(appId: AppId) extends CosmosError {
```
takirala: @larioj Should I rename this class?

larioj: Yes.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/SDKUninstallVarNotFound.scala
```diff
@@ -0,0 +1,15 @@
+package com.mesosphere.cosmos.error
+
+import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
+import io.circe.Encoder
+import io.circe.JsonObject
+import io.circe.generic.semiauto.deriveEncoder
+
+final case class SDKUninstallVarNotFound(appId: AppId) extends CosmosError {
```
larioj: Should be SdkUninstallVarNotFound. The scala convention is that abbreviations are treated as words e.g. it would be HttpResponse, rather than HTTPResponse.

PR #689, opened 2017-09-05
Remove encoders and decoders and move the code to relevant classes

takirala:
Addresses https://jira.mesosphere.com/browse/DCOS-11045


cosmos-common/src/main/scala/com/mesosphere/cosmos/circe/Decoders.scala
```diff
@@ -1,21 +1,43 @@
...
 import io.circe.ParsingFailure
+import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
 import java.util.Base64
 import scala.reflect.ClassTag
 import scala.reflect.classTag
 
 object Decoders {
 
+  implicit val decodeUri: Decoder[Uri] = Decoder.decodeString.map(Uri.parse)
```
larioj: Why were these guys moved here?

takirala: they were in `com.mesosphere.universe.common.circe` but were being used extensively in `com.mesosphere.cosmos.*` (like rpc models etc..) so i moved them here and got rid of the `Encoders.scala` and `Decoders.scala` in `com.mesosphere.universe.common.circe`

larioj: 👍

PR #691, opened 2017-09-07
[DCOS_OSS-1603] Fix broken testOnly command

takirala:
`testOnly` was not working previously. This fixes that by re-adding the `definedTests` parameter which makes sure all the tests are discovered. This makes sure the `filter` (provided as *TestName*) on the `testOnly` will filter on list of all tests instead of an empty list.

This PR also bring few other changes:
- Update cosmos binary in the `cosmos-package.json` app to latest version and add a health check.
- Add a sleep and retry in `HttpClient` (This is for [DCOS-45802](https://jira.mesosphere.com/browse/DCOS-45802))
- Perform some clean up (Such as in `UninstallHandlerSpec`)


jsancio:
@takirala and @larioj, one way around this is to run http://doc.scalatest.org/3.0.0/index.html#org.scalatest.run$ directly. For example:
```
scala> org.scalatest.run(new ListVersionsSpec with IntegrationBeforeAndAfterAll)
```

Can you try and see if this solve your issue. If so let's document it.

takirala:
@jsancio how do I get the `repl` to execute that command (none of the `console`, `consoleProject` or `consoleQuick` have all the classes in class path) ?

takirala:
Thanks for the review @alenkacz and @kvish 
I will fix the failing CI test via mesosphere/cosmos-enterprise#65

PR #692, opened 2017-09-26
Move common injection to their corresponding class

jsancio:


jsancio:
@larioj We need this to be part of the common library.

cosmos-common/src/main/scala/com/mesosphere/universe/v3/syntax/PackageDefinitionOps.scala
```diff
@@ -1,188 +0,0 @@
-package com.mesosphere.universe.v3.syntax
-
-import com.mesosphere.cosmos.rpc.v1.model.PackageCoordinate
-import com.mesosphere.universe
-import io.circe.Json
-import io.circe.JsonObject
-import io.circe.syntax.EncoderOps
-
-final class PackageDefinitionOps(val pkgDef: universe.v4.model.PackageDefinition) extends AnyVal {
```
takirala: why not retain the class instead of declaring it as `implicit` in PackageDefintion ?

jsancio: I moved the class as an implicit anyval in `PackageDefinition` now we can use the ops without having to import implicit conversions.

jsancio: Keep it as a separate class wasn't very helpful because `PackageDefinition.scala` itself was using the ops class.

jsancio:
Thanks for the review @takirala

PR #693, opened 2017-09-27
Use base32 for package coordinate

jsancio:


jsancio:
@larioj Can I get a reivew?

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PackageCoordinate.scala
```diff
@@ -19,16 +19,18 @@ object PackageCoordinate {
...
         coordinate.asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
-      )
+      ).toLowerCase.replace("=", "")
     }
 
     def rev(str: String): Try[PackageCoordinate] = {
       Try {
         val coordinate  = new String(
-          Base64.getUrlDecoder.decode(str),
+          base32.decode(str.toUpperCase),
```
takirala: what was the idea for this being an `toUpperCase` now (and not before)?

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PackageCoordinate.scala
```diff
@@ -19,16 +19,18 @@ object PackageCoordinate {
...
 
   implicit val stringInjection: Injection[PackageCoordinate, String] = {
+    val base32 = new Base32()
+
     def fwd(coordinate: PackageCoordinate): String = {
-      Base64.getUrlEncoder.encodeToString(
+      base32.encodeAsString(
         coordinate.asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
-      )
+      ).toLowerCase.replace("=", "")
```
larioj: This works on the reverse direction?

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PackageCoordinate.scala
```diff
@@ -19,16 +19,18 @@ object PackageCoordinate {
   implicit val decodePackageCoordinate: Decoder[PackageCoordinate] = deriveDecoder[PackageCoordinate]
 
   implicit val stringInjection: Injection[PackageCoordinate, String] = {
+    val base32 = new Base32()
+
     def fwd(coordinate: PackageCoordinate): String = {
-      Base64.getUrlEncoder.encodeToString(
+      base32.encodeAsString(
         coordinate.asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
```
larioj: Does this sort the Json? It would be nice if we could do `coordinate.asJson.sortedNoSpaces.getBytes(StandardCharsets.UTF_8)` that way the encoding of the coordinate would always be unique.

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PackageCoordinate.scala
jsancio: Yes: https://github.com/apache/commons-codec/blob/trunk/src/main/java/org/apache/commons/codec/binary/Base32.java#L376-L378

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PackageCoordinate.scala
jsancio: By default base32 uses uppercase letters. I want us to instead use lowercases. This is mainly because the Docker API expects lowercases in their API. We are going to use PackageCoordinates for the repo in docker registry land.

That means that during encoding we need to lowercase it and during decoding we need to uppercase it.

cosmos-common/src/main/scala/com/mesosphere/cosmos/rpc/v1/model/PackageCoordinate.scala
jsancio: I think this does what you want: https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Printer.scala#L220-L224. I think Circe's `Json` class preserves the order and I think `coordinate.asJson` is deterministic. I.e. it always generates the same order.

I the future we may not want to use Json at all of this. I will create an issue to address it and mark it as a blocker.

PR #694, opened 2017-10-03
Basic refactor for MediaType

jsancio:
Finish this tasks before merging:

- [x] Analyze the dependency report and make sure that we are using the same version and there are any major evictions.


cosmos-server/src/main/scala/com/mesosphere/cosmos/repository/UniverseClient.scala
```diff
@@ -383,16 +379,6 @@ object DefaultUniverseClient {
...
     List(MediaTypes.UniverseV4Repository, MediaTypes.UniverseV3Repository, MediaTypes.UniverseV2Repository)
-
-  def parseContentType(header: Option[String]): TwitterTry[MediaType] = {
-    TwitterTry(header.getOrElse(throw UnsupportedContentType(SupportedMediaTypes).exception))
-      .flatMap(MediaTypeParser.parse)
-      .handle {
-        case MediaTypeParseError(_, _) =>
-          throw UnsupportedContentType(SupportedMediaTypes).exception
-      }
-  }
```
jsancio: This is not used anywhere. Is this a bug or it is okay to delete this? cc @takirala

cosmos-common/src/main/scala/com/mesosphere/http/QualityValue.scala
```diff
@@ -19,15 +21,15 @@ object QualityValue {
   def parse(value: String): Try[QualityValue] = {
```
jsancio: This is only used by tests. Should we just remove this? cc @takirala @larioj

jsancio:
@larioj This is ready for review!

cosmos-common/src/main/scala/com/mesosphere/http/QualityValue.scala
larioj: That's fine by me

jsancio:
Thanks for the review @larioj !

PR #696, opened 2017-10-20
HttpClient error class code cleanup

takirala:
* Remove sealed trait in HttpClient
* Remove Either (of HttpClient fetch) to throw/handle the error
* Change unit/it tests accordingly.

Addresses [DCOS_OSS-1669](https://jira.mesosphere.com/browse/DCOS_OSS-1669) and [DCOS_OSS-1739](https://jira.mesosphere.com/browse/DCOS_OSS-1739)


cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
```diff
@@ -100,7 +108,7 @@ object HttpClient {
         throw UnsupportedRedirect(List(uri.scheme.get), loc).exception
       case status =>
         sr.scope("status").counter(status.toString).incr()
-        throw UnexpectedStatus(status)
+        throw GenericHttpError(HttpMethod.GET, uri, Status.fromCode(status)).exception
```
jsancio: What if `Status.fromCode(status)` fails? It can fail if the status code return by the client is one that finagle/netty doesn't know about.

takirala: This is the definition of `fromCode` : 
```
def fromCode(code: Int): Status =
    statuses.getOrElse(code, Status(code))
```

While `statuses` is a map for all standard error codes, If it is not present, it will instantiate a new one. This is the `unapply` : 

```
object Status {
    object Unknown {
        def unapply(status: Status): Option[Status] =
          Some(status).filter(s => s.code < 100 || s.code >= 600)
     }
  .
  .
}
```
Unless finagle changes this, we should be good.

jsancio: Sounds good. Thanks for the update!

PR #697, opened 2017-10-20
Refactor package coordinate

jsancio:


cosmos-common/src/main/scala/com/mesosphere/universe/v1/model/PackageCoordinate.scala
```diff
@@ -12,32 +12,45 @@ import java.nio.charset.StandardCharsets
...
       Try {
         val coordinate  = new String(
-          base32.decode(str.toUpperCase),
+          base32.decode(string.toUpperCase),
           StandardCharsets.UTF_8
         )
 
         decode[PackageCoordinate](coordinate)
-      }
+      } fold (
```
takirala: How does this work? I thought it should be like
```
} fold (
    pc => PassWith(pc)
    _ => Fail, 
   )
```

jsancio: For `Try` the first parameter is of type `Throwable => R` while the second parameter is `T => R`. https://www.scala-lang.org/api/current/scala/util/Try.html#fold[U](fa:Throwable=>U,fb:T=>U):U

PR #698, opened 2017-10-25
Change the test URLs

jsancio:


takirala:
@jose Any ideas why github didn't work but google did ?

jsancio:
Yes. Github.com actually cares about the accept header while google.com just ignores. Github.com was return 406.

PR #702, opened 2017-11-20
add dcos-log v2 capability

mnaboka:
with DC/OS 1.11 we are introducing new logging API. Also known as dcos-log v2.
A new capability is required to handle the logs properly from dcos-cli.

dcos-log v2 [design doc](https://docs.google.com/document/d/1Z2gWi1H7lc3nNGqLUU9Hecvm9EyPLI-6t3uK1ANlHyA/edit#heading=h.apvnjfls8tkq)

Related PRs:
[dcos-log](https://github.com/dcos/dcos-log/pull/43)
[dcos-cli](https://github.com/dcos/dcos-cli/pull/1105)
[dcos](https://github.com/dcos/dcos/pull/1935)


klueska:
@jsancio Do we have an idea of how we want to change the current schema to be more generic for use cases like this? I'm inclined to allow the change as is for now, with the understanding that we design a proper v2 CAPABILITIES schema going forward.

jsancio:
> Do we have an idea of how we want to change the current schema to be more generic for use cases like this? I'm inclined to allow the change as is for now, with the understanding that we design a proper v2 CAPABILITIES schema going forward.

I don't @klueska as I don't really understand the problem we are trying to solve. We should look at prior art and try to describe the problem we are trying to solve with this endpoint.

PR #706, opened 2017-12-08
Clean up cosmos error

takirala:
I will create one for cosmos-ee if needed. So far, everything seems to compile & pass.


cosmos-common/src/main/scala/com/mesosphere/cosmos/error/CosmosException.scala
```diff
@@ -10,15 +10,10 @@ final case class CosmosException(
...
 object CosmosException {
-  def apply(error: CosmosError): CosmosException = {
-    CosmosException(error, Status.BadRequest, Map.empty, None)
-  }
-
   def apply(
     error: CosmosError,
-    causedBy: Throwable,
-    status: Status = Status.BadRequest
+    causedBy: Option[Throwable] = None
```
takirala: I changed this to be Option to get rid of the other `apply()` method. Cuz of this, I had to change the way `causedBy` is passed to `Some(causedBy)` from just `causedBy`. The number of changes were less, so I did this. Let me know if this is not ideal and I can revert.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/IncompleteUninstall.scala
```diff
@@ -11,6 +12,8 @@ final case class IncompleteUninstall(
   override def message: String = {
     s"Incomplete uninstall of package [$packageName] due to Mesos unavailability"
   }
+
+  override def status: Status = Status.ServiceUnavailable
```
takirala: This was being passed from outside. However as there is only one value, I moved it here. If needed, we can always move this out.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/GenericHttpError.scala
```diff
@@ -18,9 +18,8 @@ final case class GenericHttpError(
     s"Unexpected upstream http error: ${method.getName} ${uri.toString} ${clientStatus.code}"
   }
 
-  def exception(status: Status): CosmosException = {
-    CosmosException(this, status, Map.empty, None)
-  }
+  override def status: Status = clientStatus
```
jsancio: This is not semantically the same. In this case you want:
```scala
final case class GenericHttpError(
   status: Status,
   method: HttpMethod = HttpMethod.GET,
   uri: Uri,
   clientStatus: Status
 ) extends CosmosError {
  ...

  // It should but double check that "def status" is overridden by the "val status" above.
 }

File: cosmos-common/src/main/scala/com/mesosphere/cosmos/error/UniverseClientHttpError.scala [hunk 4]
```diff
@@ -11,7 +11,8 @@ import org.jboss.netty.handler.codec.http.HttpMethod
 final case class UniverseClientHttpError(
   packageRepository: rpc.v1.model.PackageRepository,
   method: HttpMethod,
-  clientStatus: Status
+  clientStatus: Status,
+  cosmosStatus: Status
```
jsancio: rename `cosmosStatus` to `status` and remove the `override def status`. It should but let's check that Scala overrides `def status` when you do this.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Services.scala
```diff
@@ -77,22 +77,16 @@ object Services {
       service(request)
         .rescue {
           case ce: ChannelException =>
-            Future.exception(CosmosException.apply(
-              ServiceUnavailable(serviceName),
-              ce,
-              Status.ServiceUnavailable)
+            Future.exception(
+              CosmosException.apply(ServiceUnavailable(serviceName), Some(ce))
```
jsancio: You can omit the `apply` symbol and use this syntax instead `CosmosException(ServiceUnavailable...)`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Services.scala
```diff
@@ -77,22 +77,16 @@ object Services {
...
             )
           case e: NoBrokersAvailableException =>
-            Future.exception(CosmosException.apply(
-              ServiceUnavailable(serviceName),
-              e,
-              Status.ServiceUnavailable)
+            Future.exception(
+              CosmosException.apply(ServiceUnavailable(serviceName), Some(e))
```
jsancio: You can omit the `apply` symbol and use this syntax instead `CosmosException(ServiceUnavailable...)`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/Services.scala
```diff
@@ -77,22 +77,16 @@ object Services {
...
             )
           case e: RequestException =>
-            Future.exception(CosmosException.apply(
-              ServiceUnavailable(serviceName),
-              e,
-              Status.ServiceUnavailable)
+            Future.exception(
+              CosmosException.apply(ServiceUnavailable(serviceName), Some(e))
```
jsancio: You can omit the `apply` symbol and use this syntax instead `CosmosException(ServiceUnavailable...)`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -111,12 +111,8 @@ private[cosmos] final class UninstallHandler(
...
               throw MultipleFrameworkIds(op.packageName, op.packageVersion, fwName, all).exception
           } handle {
-            case su @ CosmosException(ServiceUnavailable(_), _, _, _) =>
-              throw CosmosException.apply(
-                IncompleteUninstall(op.packageName),
-                su,
-                Status.ServiceUnavailable
-              )
+            case su @ CosmosException(ServiceUnavailable(_), _, _) =>
+              throw CosmosException.apply(IncompleteUninstall(op.packageName), Some(su))
```
jsancio: You can omit the `apply` symbol and use this syntax instead `CosmosException(ServiceUnavailable...)`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageRepositoryAddHandler.scala
```diff
@@ -46,7 +47,17 @@ private[cosmos] final class PackageRepositoryAddHandler(
      */
     universeClient(repository).unit.handle {
       case ce: CosmosException =>
-        throw ce.copy(status = Status.BadRequest)
+        ce.error match {
```
takirala: @jsancio Is this acceptable? 

It is a partial match but it may break when we add more errors in future. Previously we used to copy the object with overwriting status. Now, cosmos exception does not have status and cosmos error being a trait, i found this to be the only easy work around.

cosmos-common/src/main/scala/com/mesosphere/cosmos/error/CosmosError.scala
```diff
@@ -2,10 +2,12 @@ package com.mesosphere.cosmos.error
 
 import io.circe.Encoder
 import io.circe.JsonObject
+import com.twitter.finagle.http.Status
```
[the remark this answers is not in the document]
takirala: @jsancio This would result in removing the `com.twitter.finagle.http.Status` type in all the four cosmos modules. This is what you mean right?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageRepositoryAddHandler.scala
takirala: @jsancio ^^

jsancio: I think this is okay for now. I think the long term solution is to add the following to `CosmosError`:
```scala
trait CosmosError[A <: CosmosError] {
  ...
  def withStatus(status: Status): A
}
```

PR #710, opened 2018-01-18
Set release version

jsancio:


takirala:
@jsancio I know we agreed https://jira.mesosphere.com/browse/DCOS_OSS-2087 is not a priority. But as the cosmos bump is not blocking us right away, how about fixing ^ and merging it before we merge this PR?

PR #713, opened 2018-01-23
Fetch postuninstallnotes from marathon labels instead of a repository like universe

takirala:
Address [DCOS_OSS-2087](https://jira.mesosphere.com/browse/DCOS_OSS-2087)

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
@@ -15,7 +15,7 @@ import com.mesosphere.cosmos.error.UninstallNonExistentAppForPackage
 import com.mesosphere.cosmos.finch.EndpointHandler
 import com.mesosphere.cosmos.handler.UninstallHandler._
 import com.mesosphere.cosmos.http.RequestSession
-import com.mesosphere.cosmos.label
+import com.mesosphere.cosmos.repository.PackageCollection
 import com.mesosphere.cosmos.rpc
 import com.mesosphere.cosmos.service.ServiceUninstaller
 import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
@@ -31,6 +31,7 @@ import org.slf4j.Logger
 
 private[cosmos] final class UninstallHandler(
   adminRouter: AdminRouter,
+  packageCollection: PackageCollection,
   uninstaller: ServiceUninstaller
 ) extends EndpointHandler[rpc.v1.model.UninstallRequest, rpc.v1.model.UninstallResponse] {
   lazy val logger: Logger = org.slf4j.LoggerFactory.getLogger(getClass)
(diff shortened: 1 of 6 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -54,23 +53,16 @@ private[cosmos] final class UninstallHandler(
...
-      .map { detailsAndNotes =>
-        val results = detailsAndNotes.map { case (detail, postUninstallNotes) =>
+      .map { details =>
+        val results = details.map { case detail =>
           rpc.v1.model.UninstallResult(
             detail.packageName,
             detail.appId,
             detail.packageVersion,
-            postUninstallNotes
+            detail.packageMetadata match {
```
larioj: I thought we would do best effort meaning that we would attempt one and if that was not available we would attempt the other. It seems like we're doing a lot of corroboration between marathon and the package registries. We are doing this elsewhere too (in the upgrade endpoint). We should try to abstract this away. Maybe we can create an class called `InstalledPackages` or something better that looks at both marathon and the packageCollection to try to get information about a particular package.

takirala: If I am able to get the metadata, thus `postUninstallNotes` from the marathon label, there is no need to fall back to package collection right? If I understand correctly, best effort is when the first one (in this case marathon label) fails then we fall back to second one (in this case package collection) but if the marathon label doesn't provide the `postUninstallNotes` there is no reason to look at the package collection (as we can be sure it would give the same data - unless I am missing something?). 

As for the later comment, I would say lets keep this PR small to prioritize the effort and we can revisit when this comes to be a repetitive pattern, I can add a TODO note mentioning the same, what do you say?

jsancio: I agree with @larioj. We are doing something similar at least in one other place. We should standardize how we get the PackageDefinition from a running service MarathonApp and use that algorithm everywhere.

larioj: @takirala detail.packageMetadata returns an optional because that particular label may have not been stored if the service was started with an old cosmos. In those cases we could possibly retrieve the uninstall notes from the registry rather than just returning None.

As for whether to do the refactor now or later. I'd say try to look at https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceUpdateHandler.scala#L120 
It's basically doing what you need to do here. You do not need to create a whole class but at least try to see if you can use the function. Let me know if you have questions.

takirala:
I have moved the code to package object which is now being used by ServiceUpdate Handler and Uninstall Handler. If this is good enough, I will need to make the changes in cosmos-ee repo as well. 
@larioj @jsancio what do you think?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/package.scala
```diff
@@ -0,0 +1,84 @@
...
+    packageCollection: PackageCollection,
+    marathonApp: MarathonApp
+  )(
+    implicit session: RequestSession
+  ): Future[PackageWithSource] = {
+    orElse(Future.value(getStoredPackageWithSource(marathonApp)))(
+      lookupPackageWithSource(packageCollection, marathonApp)
+    ).map(_.getOrElse {
+      throw new IllegalStateException(
+        "Unable to retrieve the old package definition"
```
larioj: This is throwing an error. In the uninstall case we are okay with `None`. Maybe change the signature of this function to return an optional and then do the throw elsewhere.

larioj:
@takirala The organization of the code looks good to me. Left you a comment.

PR #719, opened 2018-02-23
Fix service uninstall when the env var is null in marathon json

takirala:
Address [DCOS-21374](https://jira.mesosphere.com/browse/DCOS-21374)

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
@@ -28,6 +28,7 @@ import com.twitter.finagle.http.Status
 import com.twitter.util.Future
 import io.circe.Json
 import io.circe.JsonObject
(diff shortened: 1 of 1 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -178,15 +179,13 @@ private[cosmos] final class UninstallHandler(
...
-    // Those are removed here to ensure the Marathon update succeeds.
-    // Presently, those fields are:
-    // - uris
-    // - version
     appJson.add(
       envJsonField,
       Json.fromJsonObject(
-        appJson(envJsonField).get.asObject.get.add(SdkUninstallEnvvar, Json.fromString("true"))
+        appJson(envJsonField)
+          .map(_.asObject.get)
```
jsancio: `.flatMap(_.asObject)`. We should feel uncomfortable when we see a `.get` in the code. :)

PR #727, opened 2018-03-21
Update all of the version the latest versions

jsancio:


jsancio:
@takirala Ready for review. I won't merge it until I get the other depending projects working.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/finch/RequestValidatorsSpec.scala
```diff
@@ -166,9 +167,10 @@ final class RequestValidatorsSpec extends FreeSpec with Matchers with PropertyCh
     request: HttpRequest
   ): Assertion = {
     whenever (MediaType.parse(request.headers(Fields.ContentType)).isFailure) {
+      // Work around for DCOS_OSS-2289
```
takirala: Can you comment on [DCOS_OSS-2289](https://jira.mesosphere.com/browse/DCOS_OSS-2289) and close if applicable?

project/Deps.scala
```diff
@@ -84,11 +84,6 @@ object Deps {
     "com.github.spullara.mustache.java" % "compiler" % V.mustache
   )
 
-  // APLv2.0
-  val netty = Seq(
-    "io.netty" % "netty-common" % V.netty
```
takirala: considering netty is one of the crucial libraries we use, Can you add a comment on who brings in `netty` for us? We can refer in future if we are to upgrade netty or just to refer which version of netty we are using.

cosmos-test-common/src/test/scala/com/mesosphere/cosmos/finch/RequestValidatorsSpec.scala
takirala: nvm misread the diff.

PR #732, opened 2018-05-30
Release patch v0.4.4

takirala:
Address [DCOS-37674](https://jira.mesosphere.com/browse/DCOS-37674)

We are using a very old version of node js (used for generating docs) which resulted in [this error](https://teamcity.mesosphere.io/viewLog.html?buildId=1079083&buildTypeId=DcosIo_Cosmos_PublishRelease&tab=buildLog&_focus=13619). This is not needed for cosmos 0.5.x and above as we have moved the documentation to dcos docs site.

### DIFF
--- a/docker/ramltools/Dockerfile
+++ b/docker/ramltools/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:4.8
+FROM node:9.1
 RUN npm install -g raml2html && \
     npm install -g https://github.com/stoplightio/api-spec-converter
 COPY raml2swagger.js /usr/local/bin/raml2swagger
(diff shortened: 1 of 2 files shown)


takirala:
@benclarkwood we haven't released v0.4.x in a while and when i released 0.4.3 recently, the very old version of node js that we use resulted in a CI failure (linked in the description). we can discard 0.4.3 and can release v0.4.4 of cosmos to DC/OS.

docker/ramltools/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:4.8
```
takirala: Tested that this works locally.

docker/ramltools/Dockerfile
```diff
@@ -1,4 +1,4 @@
...
+FROM node:9.1
```
takirala: Tested that this works locally.

PR #734, opened 2018-08-01
Improve logging formatting and upgrade scala version

takirala:
Address [DCOS_OSS-3690](https://jira.mesosphere.com/browse/DCOS_OSS-3690)

TODO : Updated to latest scala version. Need to do the same in ee version as well.

### DIFF
--- a/cosmos-common/src/main/scala/com/mesosphere/cosmos/finch/EndpointHandler.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/cosmos/finch/EndpointHandler.scala
@@ -16,7 +16,8 @@ abstract class EndpointHandler[Request, Response](successStatus: Status = Status
   final def apply(context: EndpointContext[Request, Response]): Future[Output[Json]] = {
     apply(context.requestBody)(context.session)
       .respond {
-        case Throw(e) => logger.warn(s"Processing [${context.requestBody}] resulted in : ${e.getMessage}")
+        case Throw(e) =>
+          logger.warn(s"Processing [${context.requestBody}] resulted in : ${e.getMessage}")
         case _ => ()
       }
       .map { response =>
--- a/cosmos-common/src/main/scala/com/mesosphere/cosmos/finch/EndpointHandler.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/cosmos/finch/EndpointHandler.scala
@@ -4,19 +4,26 @@ import com.mesosphere.cosmos.http.RequestSession
 import com.twitter.finagle.http.Fields
 import com.twitter.finagle.http.Status
 import com.twitter.util.Future
+import com.twitter.util.Throw
 import io.circe.Json
 import io.circe.syntax._
 import io.finch._
 
 abstract class EndpointHandler[Request, Response](successStatus: Status = Status.Ok) {
 
-  final def apply(context: EndpointContext[Request, Response]): Future[Output[Json]] = {
-    apply(context.requestBody)(context.session).map { response =>
-      val encodedResponse = response.asJson(context.responseEncoder.encoder)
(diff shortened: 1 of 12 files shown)


takirala:
Ping @kvish and @jeid64 please review

cosmos-common/src/main/scala/com/mesosphere/cosmos/finch/EndpointHandler.scala
```diff
@@ -4,19 +4,26 @@ import com.mesosphere.cosmos.http.RequestSession
...
-      val encodedResponse = response.asJson(context.responseEncoder.encoder)
+  private[this] val logger = org.slf4j.LoggerFactory.getLogger(getClass)
 
-      Output.payload(encodedResponse, successStatus)
-        .withHeader(Fields.ContentType -> context.responseEncoder.mediaType(response).show)
-    }
+  final def apply(context: EndpointContext[Request, Response]): Future[Output[Json]] = {
+    apply(context.requestBody)(context.session)
+      .respond {
+        case Throw(e) => logger.warn(s"Processing [${context.requestBody}] resulted in : ${e.getMessage}")
```
nickbp: Is there a risk that `requestBody` is very big? If so it may make sense to truncate it to the first 1024 bytes or so

cosmos-server/src/main/scala/com/mesosphere/cosmos/MarathonPackageRunner.scala
```diff
@@ -31,13 +34,15 @@ final class MarathonPackageRunner(adminRouter: AdminRouter) {
           case Status.Conflict =>
             throw ServiceAlreadyStarted().exception
           case status if (400 until 500).contains(status.code) =>
+            logger.warn(s"Marathon returned [${status.code}]: ${response.contentString}")
```
nickbp: Similar to above, except for `contentString`. Here and below.

cosmos-common/src/main/scala/com/mesosphere/cosmos/finch/EndpointHandler.scala
takirala: I referred to cosmos [docs](https://docs.mesosphere.com/1.11/deploying-services/package-api/) when I was implementing this. Almost all the RPCs are limited by the number of fields they can have. Only for the `/package/install` we may have unbounded number of json fields in the `options` field. I made the above judgement based on : 
 * We only print the body when there is an error. 
 * If it got to this point, that means the request resolved to one of the RPCs definitely thus making sure that we are not printing some arbitrary content posted to an arbitrary endpoint.
 * If we truncate the body, we may end up losing some interesting data when trying to debug an error. And of course, we can always come back to addressing this if it turns out to be a problem.

If you still think truncating is a good option.

cosmos-common/src/main/scala/com/mesosphere/cosmos/finch/EndpointHandler.scala
```diff
@@ -4,19 +4,27 @@ import com.mesosphere.cosmos.http.RequestSession
...
+  private[this] val logger = org.slf4j.LoggerFactory.getLogger(getClass)
 
-      Output.payload(encodedResponse, successStatus)
-        .withHeader(Fields.ContentType -> context.responseEncoder.mediaType(response).show)
-    }
+  final def apply(context: EndpointContext[Request, Response]): Future[Output[Json]] = {
+    apply(context.requestBody)(context.session)
+      .respond {
+        case Throw(e) =>
+          logger.warn(s"Processing [${context.requestBody}] resulted in : ${e.getMessage}")
```
nickbp: Could use `trimContentForPrinting()` here?

PR #735, opened 2018-08-02
Change ResourceProxyHandler to stream data instead of buffering

takirala:


takirala:
Ping @kvish @jeid64 please review

PR #741, opened 2018-08-22
Fix a race condition while trying to create zk nodes

takirala:
Address [DCOS-40633](https://jira.mesosphere.com/browse/DCOS-40633)

**Problem** :
When multiple requests (e.g.: Package describe requests) try to use `PackageCollection`, cosmos is translating those multiple requests in to multiple non-synchronized calls to ZK to `createOrSet` and this is problematic as only the first create call succeeds and all the rest of the calls throw an error.

**Solution** :
Move the ZK read logic to be inside the caffeine future `LoadingCache` which auto manages multiple requests that result in futures.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
@@ -144,5 +144,5 @@ object ServiceUninstaller {
 
   val RetryInterval: Duration = 10.seconds
 
-  private val DefaultRetries: Int = 50
+  private val DefaultRetries: Int = 10000
 }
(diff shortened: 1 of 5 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
```diff
@@ -144,5 +144,5 @@ object ServiceUninstaller {
 
   val RetryInterval: Duration = 10.seconds
 
-  private val DefaultRetries: Int = 50
+  private val DefaultRetries: Int = 10000
```
takirala: @nickbp previously we were waiting for 10 * 50 ~ 10 minutes.
i bumped this to 10 * 10000 ~ 27 hours ~ 1 day.

PR #746, opened 2018-08-25
adding v5 package tests

kvish:


takirala:
- Do we have integration tests in for all the four handlers that were modified in the recent custom package manager change?
- Is it possible to get the package definition that kubernetes uses to create a v5 package and install the manager from it and then use the manager for actual routing?
- If the above is too much work, we can at least replace the cosmos in resource section of https://infinity-artifacts.s3.amazonaws.com/cosmos-integration-test/package-with-manager/stub-universe-hello-world.json with a cosmos that we already published at https://teamcity.mesosphere.io/viewLog.html?buildId=1176018&buildTypeId=DcosIo_Cosmos_PublishRelease&tab=artifacts (v0.6.0 ) ?

kvish:
1. We have tests for all 4 RPCs
2. dont think it is in our scope to use their package manager to test, this testing should be done by k8s. We should just verify that the RPCs are routed correctly
3. updated the cosmos in our marathon app

takirala:
Where are the integration tests for Service Update handler - I looked at https://github.com/dcos/cosmos/blob/master/cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/ServiceUpdateSpec.scala ?

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -19,6 +19,8 @@ object ItObjects {
     "ae6a07ac0b53924154add2cd61403c5233272d93/repo/repo-up-to-1.10.json"
   val V4TestUniverseConverterURI = "https://universe-converter.mesosphere.com/transform?url=" + V4TestUniverse
 
+  val V5TestPackage = "http://downloads.mesosphere.io/dcos-enterprise/cosmos/v0.5/stub-universe-hello-world.json"
```
takirala: What is `v0.5` ? You mean `v5.0` ?

kvish: ah shoot good catch, let me update tools infra to fix that path

kvish:
bad merge with my branch to reenable tests, good catch with service update test

takirala:
Also do we comply with the error messages in the Desing Doc.
https://docs.google.com/document/d/1hJTZImY5JBeIBkQVgPUjVu0TwMb6bkzu71q4SbDDIfo/edit#heading=h.diwc7x48wjuw ?
For example, where are we testing for `Package Manager [foo] for package [bar] was not found. Install [foo] or specify managerId` ?

I am not sure if that is intentionally excluded. If so, can you file tickets for them? cc @nickbp

nickbp:
Given that we aren't including a user-facing `managerId` field, that error should probably be tweaked to omit mentioning that option.

kvish:
we will currently error with marathon app not found when checking if the app is running before forwarding the request. We have a debug log that will log the error (manager xyz wasnt running), if the custom manager is not running, this should be sufficient for todays purposes since we can easily check the log and see what the error in talking to the manager was. This should be sufficient considering k8s will be the only user

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/ItObjects.scala
```diff
@@ -19,6 +19,8 @@ object ItObjects {
     "ae6a07ac0b53924154add2cd61403c5233272d93/repo/repo-up-to-1.10.json"
   val V4TestUniverseConverterURI = "https://universe-converter.mesosphere.com/transform?url=" + V4TestUniverse
 
+  val V5TestPackage = "http://downloads.mesosphere.io/dcos-enterprise/cosmos/v5.0/stub-universe-hello-world.json"
```
nickbp: Is this file in version control? Should have a local copy that gets uploaded with the build or something.

nickbp: Could have a comment here specifying that the file is manually updated, and the source content is located at SOME_REPO_PATH.

kvish: RPC requires a uri is passed, we can keep a local copy as well? not sure if that is valuable since we assume this file will rarely change (part of the reason it is in downloads.mesosphere.io and not infinity artifacts)

kvish: as discussed this is added under resources in cosmos integration tests

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/ServiceUpdateSpec.scala
```diff
@@ -1,19 +1,24 @@
...
+import com.mesosphere.cosmos.ItOps._
+import com.mesosphere.cosmos.http.CosmosRequests
 import com.mesosphere.cosmos.http.HttpRequest
 import com.mesosphere.cosmos.http.ServiceRpcPath
 import com.mesosphere.cosmos.http.TestContext
-import com.mesosphere.cosmos.rpc
 import com.mesosphere.cosmos.rpc.MediaTypes
 import com.mesosphere.cosmos.test.CosmosIntegrationTestClient
 import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
+import com.mesosphere.cosmos.rpc.v1.model.ServiceUpdateRequest
```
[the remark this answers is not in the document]
takirala: The reason for this is it gets highly unreadable when you have multiple versions.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/ServiceUpdateSpec.scala
```diff
@@ -146,6 +170,14 @@ object ServiceUpdateSpec {
     )(timeout)
   }
 
+  def submitServiceUpdateRequest(
+    request: ServiceUpdateRequest
+  )(
+    implicit testContext: TestContext
+  ): Response = {
+    CosmosClient.submit(CosmosRequests.serviceUpdate(request))
```
takirala: Since this is being used in only one place and the method body length is one line, you can remove this method.
Or extend the `serviceUpdate` below to do what you need to do.

cosmos-integration-tests/src/main/resources/package-with-custom-manager.json
```diff
@@ -0,0 +1,354 @@
+// NOTE:This file is a reflection of http://downloads.mesosphere.io/dcos-enterprise/cosmos/v5.0/stub-universe-hello-world.json.
```
takirala: Move these notes to a read me (easy to copy paste the file as a valid json). Add info on the whom to ping (tools-infra) to update data.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/ServiceUpdateSpec.scala
kvish: resolved calling the method inline

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/ServiceUpdateSpec.scala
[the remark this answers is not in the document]
kvish: resolved

PR #747, opened 2018-08-28
fix silent null ref in custom package manager and pass through all arguments

kvish:


kvish:
good catch hector, we werent passing them through because we were running pre custom manager cosmos as our manager for tests which dont understand these fields
this is updated now

PR #748, opened 2018-08-29
package-manager-router: get managerId using the package name

hectorj2f:
Uninstall operation isn't able to obtain the `managerId` so the package manager isn't called.


hectorj2f:
Pinging @kvish for a review

kvish:
closing since my PR will include this with a few other changes

PR #749, opened 2018-08-29
[DCOS-41118] pass on all information derived from appId

kvish:
if we get a command containing packageName and appId it will be routed correctly this addresses that issue


cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -53,7 +53,7 @@ private[cosmos] final class UninstallHandler(
       req.managerId,
       Option(req.packageName),
       req.packageVersion,
-      req.appId
+      appId = if(req.appId.isDefined) req.appId else Option(AppId(req.packageName))
```
takirala: You can do `Some(appId.getOrElse(req.packageName)`

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -53,7 +53,7 @@ private[cosmos] final class UninstallHandler(
       req.managerId,
       Option(req.packageName),
       req.packageVersion,
-      req.appId
+      appId = Some(req.appId.getOrElse(AppId(req.packageName)))
```
hectorj2f: To clarify this can be None if the appId isn't the same of our package name. Then the package manager should code something similar to what it's done in https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala#L62  and https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala#L256. Is that correct ? @kvish @nickbp

kvish: yes this is correct, the appId does not need to be the same as the packageName, will need to find the appId first before making this determination

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -49,61 +49,63 @@ private[cosmos] final class UninstallHandler(
...
+              case (app, uninstallOp) =>
+                customPackageManagerRouter.getCustomPackageManagerId(
+                  req.managerId,
+                  Option(req.packageName),
+                  req.packageVersion,
+                  Some(req.appId.getOrElse(uninstallOp.appId))
+                ).flatMap {
+                  case Some(managerId) if !managerId.isEmpty =>
+                    logger.debug(s"Request [$req] requires custom manager: [$managerId]")
+                    return customPackageManagerRouter.callCustomPackageUninstall(req, managerId) //scalastyle:ignore return
```
hectorj2f: @kvish Why don't we pass the `uninstallOp.appId` value to the package manager if it's `None`? I made a PR to your PR https://github.com/dcos/cosmos/pull/750

kvish:
PR now addresses the following issues:
- if request contains name + appId, get information via appId when no version supplied
- Decorate outbound requests to the custom manager with packageName/packageVersion and in the case of ServiceUpdate add currentPackageVersion
- support uninstall from just packageName

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -49,61 +49,63 @@ private[cosmos] final class UninstallHandler(
...
+              case (app, uninstallOp) =>
+                customPackageManagerRouter.getCustomPackageManagerId(
+                  req.managerId,
+                  Option(uninstallOp.packageName),
+                  uninstallOp.packageVersion.as[Option[universe.v3.model.Version]],
+                  Some(req.appId.getOrElse(uninstallOp.appId))
+                ).flatMap {
+                  case Some((Some(managerId), Some(pkgName), Some(pkgVersion))) if !managerId.isEmpty =>
+                    logger.debug(s"Request [$req] requires custom manager: [$managerId]")
+                    return customPackageManagerRouter.callCustomPackageUninstall(req, managerId, pkgName, pkgVersion) //scalastyle:ignore return
```
hectorj2f: `appId` is missing here. Please use `uninstallOp.appId` as I did in my PR.

kvish:
@hectorj2f resolved your comment, we will take the appId from uninstall op if it was not provided in the request

kvish:
PR addresses: DCOS-41118, supports uninstall from just a package name. We are also passing on the packageName and packageVersion we derive from marathon during the look up process.

hectorj2f:
I am testing it now

PR #750, opened 2018-08-30
uninstall-handler: fill the request to the package manager

hectorj2f:
@kvish I am wondering, why don't we send the `uninstallOp` collected properties  to the package manager in the request ?


kvish:
I am going to make a more general fix to pass args onwards for all RPCs

kvish:
i have a PR to resolve this

PR #752, opened 2018-09-04
pass origin information to custom manager

kvish:


cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
```diff
@@ -89,6 +89,9 @@ class AdminRouterClient(
   )(implicit session: RequestSession): Future[Response] = {
     val uri = "service" / managerId.toUri / "service" / "describe"
     val p = post(uri, body.asJson)
+    p.headerMap.set("X-Forwarded-Proto", session.originInfo.urlScheme.toString)
```
takirala: Reuse constant strings. For e.g. see https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/Cosmos.scala#L163

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
```diff
@@ -89,6 +89,9 @@ class AdminRouterClient(
...
+    p.headerMap.set("X-Forwarded-Host", session.originInfo.host)
+    p.headerMap.set("X-Forwarded-Port", session.originInfo.port.getOrElse(""))
```
takirala: Extract these lines into a separate method like : 
```
def withOrigin(request: Request)(implicit session: RequestSession): Request = {
    request.headerMap.set(UrlSchemeHeader, session.originInfo.urlScheme.toString)
    .... 
    ...
}
```

and then you can make the same change in all the methods that call custom manager.

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
```diff
@@ -88,7 +89,7 @@ class AdminRouterClient(
     body: rpc.v1.model.ServiceDescribeRequest
   )(implicit session: RequestSession): Future[Response] = {
     val uri = "service" / managerId.toUri / "service" / "describe"
-    val p = post(uri, body.asJson)
+    val p = addOriginHeaders(post(uri, body.asJson))
```
takirala: We can do the same for `postCustomServiceUpdate` and `postCustomPackageUninstall` for the sake of consistency ? (Or am I missing something?)

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
```diff
@@ -108,4 +109,12 @@ class AdminRouterClient(
 
 object AdminRouterClient {
   @volatile private var cachedDcosVersion: Option[DcosVersion] = None
+
+  def addOriginHeaders(request: Request)(implicit session: RequestSession): Request = {
+    request.headerMap.add(util.ForwardedProtoHeader, session.originInfo.urlScheme.toString)
+    request.headerMap.add(util.ForwardedHostHeader, session.originInfo.host)
+    request.headerMap.add(util.ForwardedPortHeader, session.originInfo.port.getOrElse(""))
```
takirala: I think it's better if we do this : 
```
session.originInfo.port match {
  case Some(..) => // then set the header.
  case None(..) => // then skip the header.
}
```
with the current logic we are setting `ForwardedPortHeader` to be an empty string which does not make sense for the downstream.

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
```diff
@@ -4,6 +4,7 @@ import com.mesosphere.cosmos.http.RequestSession
 import com.mesosphere.cosmos.rpc.MediaTypes
 import com.mesosphere.cosmos.thirdparty.adminrouter.model.DcosVersion
 import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
+import com.mesosphere.util
```
takirala: (nit) you can `import com.mesosphere.util.UrlSchemeHeader` and other `val`s

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
kvish: sure no harm in appending the headers

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
kvish: since we are using all the under the import makes sense to keep it generic like this imo

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
kvish: mainly for simplicity to construct host + port, mke can concat these without checking if the port header was sent.

takirala: i am assuming they would do something like this : 
"<hostString>" + ":" + "<port>"
if the <port> is empty the above doesn't make any sense. of course they can check if the header is blank or not, but having `""` as a valid value for the Port header does not make any sense and we should skip the header if it is not present.

kvish: sure ok I see the point will make the change

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
```diff
@@ -108,4 +109,17 @@ class AdminRouterClient(
 
 object AdminRouterClient {
   @volatile private var cachedDcosVersion: Option[DcosVersion] = None
+
+  def addOriginHeaders(request: Request)(implicit session: RequestSession): Request = {
+    session.originInfo.port match {
```
takirala: this is duplicating proto and host headers.

how about : 
```
request.headerMap.add(util.ForwardedProtoHeader, session.originInfo.urlScheme.toString)
request.headerMap.add(util.ForwardedHostHeader, session.originInfo.host)

session.originInfo.port match {
  case Some(port) => request.headerMap.add(util.ForwardedPortHeader, port)
  case None => ()
}
request
```

kvish: sure makes more sense

cosmos-server/src/main/scala/com/mesosphere/cosmos/AdminRouterClient.scala
```diff
@@ -108,4 +109,17 @@ class AdminRouterClient(
...
+
+  def addOriginHeaders(request: Request)(implicit session: RequestSession): Request = {
+    request.headerMap.add(util.ForwardedProtoHeader, session.originInfo.urlScheme.toString)
+    request.headerMap.add(util.ForwardedHostHeader, session.originInfo.host)
+
+    session.originInfo.port match {
+      case Some(port) =>
+        request.headerMap.add(util.ForwardedPortHeader, port)
+      case None =>
+        request.headerMap.add(util.ForwardedProtoHeader, session.originInfo.urlScheme.toString)
```
jimmidyson: Why are you adding the proto header again if there is no port? This is being done above already isn't it?

PR #753, opened 2018-09-04
Update Readme to link to docs tooling

takirala:


takirala:
@kvish lets update the schema rpcs in above json once all the requests from k8 are shipped.

PR #755, opened 2018-09-20
[DCOS-42020] Check for package manager before doing any uninstall work

kvish:
uninstall will not proceed if the package manager is not installed

as a note: tested this from cluster UI and cli to double check errors, cluster UI hangs with no error when cosmos reject uninstall request.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
@@ -62,49 +62,42 @@ private[cosmos] final class UninstallHandler(
               customPackageManagerRouter.callCustomPackageUninstall(
                 req, managerId, pkgName, pkgVersion, req.appId.getOrElse(apps.head.id))
             case _ => {
-              getMarathonApps(req.packageName, req.appId)
-                .map(apps => createUninstallOperations(req.packageName, apps))
-                .map { uninstallOps =>
-                  val all = req.all.contains(true)
-                  if (all || uninstallOps.size <= 1) {
-                    uninstallOps
-                  } else {
-                    throw AmbiguousAppId(req.packageName, uninstallOps.map(_._2.appId)).exception
-                  }
-                }
-                .flatMap { uninstallOps =>
-                  Future.collect(
-                    uninstallOps
-                      .map { case (app, uninstallOp) => (app, runUninstall(uninstallOp)) }
-                      .map { case (app, f) => f.map(app -> _) }
-                  )
-                }
-                .flatMap { uninstallDetails =>
-                  Future.collect(
-                    uninstallDetails.map { case (app, detail) =>
-                      getPackageWithSource(packageCollection, app).map { res =>
-                        (
-                          detail,
-                          res match {
-                            case Some((pkg, _)) => pkg.postUninstallNotes
-                            case None => None
-                          }
-                        )
-                      }
+              val uninstallOps = createUninstallOperations(req.packageName, apps)
+              val all = req.all.contains(true)
+              if (!(all || uninstallOps.size <= 1)) {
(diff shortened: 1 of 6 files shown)


kvish:
sequence of uninstalls now gives the following messages

Mesospheres-MacBook-Pro-44:~ karanvishwanathan$ dcos package uninstall beta-mke
WARNING: This action cannot be undone. This will uninstall [beta-mke] and delete all of its persistent data (logs, configurations, database artifacts, everything).
Please type the name of the service to confirm: yes
Expected 'beta-mke'. You supplied 'yes'.
Please type the name of the service to confirm: beta-mke
Uninstalled package [beta-mke] version [2.0.0-1.12.0-beta]
The Mesosphere Kubernetes Engine service is being uninstalled.
Mesospheres-MacBook-Pro-44:~ karanvishwanathan$ dcos package uninstall beta-kubernetes
WARNING: This action cannot be undone. This will uninstall [beta-kubernetes] and delete all of its persistent data (logs, configurations, database artifacts, everything).
Please type the name of the service to confirm: beta-kubernetes
the custom manager: '/beta-mke', is not installed for this package'

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -50,68 +50,58 @@ private[cosmos] final class UninstallHandler(
...
-                    )
-                }
-                (app, runUninstall(uninstallOp))
+      .flatMap { apps =>
+        customPackageManagerRouter.getCustomPackageManagerId(
+          req.managerId,
+          Option(req.packageName),
+          apps.head.packageVersion,
+          Option(req.appId.getOrElse(apps.head.id))
+        ).flatMap {
```
takirala: So I think there are few flaws in the uninstall logic. 
- The `UninstallHandler` does not handle the `all` parameter correctly. For e.g. we should NOT do `apps.head.packageVersion`. It is completely possible for the `head.pacakgeVersion` to be different from the `packageVersion` of the next element in the Sequence. 
- What if the user passes the `all` flag and some of them need to go via custom manager and some of them need to be done by cosmos itself. This would fail. Although, we may have no case of that happening, it does not harm us to handle this scenario correctly.
- One way to solve the above two is to deprecate `all` for good. However, it may be too late to make this decision for 1.12 now.
- The `method.length` are `cyclomatic.complexity` kind of important because the logic not generic and someone reading this code in the future may have some difficulty understand this. 

To address all of the above, i have created https://github.com/dcos/cosmos/pull/756 and https://github.com/mesosphere/cosmos-enterprise/pull/60

Also please see https://github.com/dcos/cosmos/pull/756#issue-216837584

kvish: 1. passing package version here is just for managerId look up from the package def. based on assumption 2 
2. 2nd assumption here is that a request is either fully handled by a custom manager or fully handled by cosmos, therefore we forward the all bit to the custom manager and let it handle the case

will review your PR

takirala:
- I have cleanup a method in Custom Manager Router so that we can remove the ignore of cyclomatic check. See https://github.com/dcos/cosmos/compare/uninstall-fixes-2...cleanup-uninstall-handler
- Can you paste a well formatted command line output when the CLI tries to `install` (I see `uninstall` above, but I am hoping the error message has been updated with proper formatting) a package that requires a custom manager but none is installed. 
- Just as an FYI, after this is merged, we have to merge https://github.com/mesosphere/cosmos-enterprise/pull/60 as well so that we can make a `0.6.1` release.

kvish:
install output with no mke

`Mesospheres-MacBook-Pro-44:~ karanvishwanathan$ dcos package install beta-kubernetes --yes
the custom manager: '/beta-mke', is not installed for this package`

takirala:
@kvish You can merge after someone from k8 team reviews the error messages.

kvish:
got an offline ack on the error messages, merging this

PR #756, opened 2018-09-20
Cleanup uninstall handler logic

takirala:
Sidenote : Please cleanup all the branches when you are done.

- Removed the cyclomatic complexity. The code was too complicated to be edited by some one in the future.
- We should think about deprecating the `--all` flag in CLI. If not one is using it, removing it makes sense.

- We should have augmented `EndpointHanlder` to contain a generic routing logic for custom package manager. With this in mind, I have created `CustomEndpointHanlder` to enable us to move the routing logic to `EndpointHanlder` in future. When we had design doc for CustomPackageManager, we had a superficial header/payload based proxying to custom manager. We significantly deviated from this (especially in `UninstallHanlder`) and this PR fixes some of this behavior. 
- Also, by doing the above, it makes is trivial to get rid of the custom package manager logic if we ever need to.

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageInstallHandler.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageInstallHandler.scala
@@ -4,7 +4,6 @@ import com.mesosphere.cosmos.MarathonPackageRunner
 import com.mesosphere.cosmos.error.CosmosException
 import com.mesosphere.cosmos.error.PackageAlreadyInstalled
 import com.mesosphere.cosmos.error.ServiceAlreadyStarted
-import com.mesosphere.cosmos.finch.EndpointHandler
 import com.mesosphere.cosmos.http.RequestSession
 import com.mesosphere.cosmos.render.PackageDefinitionRenderer
 import com.mesosphere.cosmos.repository.PackageCollection
@@ -20,16 +19,65 @@ private[cosmos] final class PackageInstallHandler(
   packageCollection: PackageCollection,
   packageRunner: MarathonPackageRunner,
   customPackageManagerRouter: CustomPackageManagerRouter
-) extends EndpointHandler[rpc.v1.model.InstallRequest, rpc.v2.model.InstallResponse] {
+) extends CustomEndpointHandler[rpc.v1.model.InstallRequest, rpc.v2.model.InstallResponse] {
 
   private[this] lazy val logger = org.slf4j.LoggerFactory.getLogger(getClass)
 
-  // scalastyle:off method.length
   override def apply(
     request: rpc.v1.model.InstallRequest
   )(
     implicit session: RequestSession
   ): Future[rpc.v2.model.InstallResponse] = {
+    tryCustomPackageManager(request).flatMap {
+      case Some(customResponse) => Future(customResponse)
+      case None =>
+        packageCollection
+          .getPackageByPackageVersion(
+            request.packageName,
+            request.packageVersion.as[Option[universe.v3.model.Version]]
+          )
+          .flatMap { case (pkg, sourceUri) =>
+            PackageDefinitionRenderer.renderMarathonV2App(
+              sourceUri,
+              pkg,
+              request.options,
+              request.appId
+            ) match {
+              case Some(renderedMarathonJson) =>
+                packageRunner.launch(renderedMarathonJson)
+                  .map { runnerResponse =>
+                    rpc.v2.model.InstallResponse(
+                      packageName = pkg.name,
+                      packageVersion = pkg.version,
+                      appId = Some(runnerResponse.id),
+                      postInstallNotes = pkg.postInstallNotes,
+                      cli = pkg.rewrite(rewriteUrlWithProxyInfo(session.originInfo), identity).cli
+                    )
+                  }
+                  .handle {
+                    case CosmosException(ServiceAlreadyStarted(_), _, _) =>
+                      throw PackageAlreadyInstalled().exception
+                  }
+              case None =>
+                Future {
+                  rpc.v2.model.InstallResponse(
+                    packageName = pkg.name,
+                    packageVersion = pkg.version,
+                    appId = None,
+                    postInstallNotes = pkg.postInstallNotes,
+                    cli = pkg.rewrite(rewriteUrlWithProxyInfo(session.originInfo), identity).cli
+                  )
+                }
+            }
+          }
+    }
+  }
+
+  override def tryCustomPackageManager(
+    request: rpc.v1.model.InstallRequest
+  )(
+    implicit session: RequestSession
+  ): Future[Option[rpc.v2.model.InstallResponse]] = {
     customPackageManagerRouter.getCustomPackageManagerId(
       request.managerId,
       Option(request.packageName),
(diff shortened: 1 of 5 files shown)


kvish:
also have a PR #755 that works as well, im coupling this with removal of the marathon app check since we can just catch the 404 and avoid 2 requests. lets merge one of these tmmrw

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageInstallHandler.scala
```diff
@@ -20,67 +19,75 @@ private[cosmos] final class PackageInstallHandler(
...
                       packageVersion = pkg.version,
-                      appId = None,
+                      appId = Some(runnerResponse.id),
                       postInstallNotes = pkg.postInstallNotes,
                       cli = pkg.rewrite(rewriteUrlWithProxyInfo(session.originInfo), identity).cli
                     )
                   }
-              }
+                  .handle {
+                    case CosmosException(ServiceAlreadyStarted(_), _, _) =>
```
kvish: would any doc need to be updated since we are changing the user facing error here

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -36,74 +35,78 @@ private[cosmos] final class UninstallHandler(
...
-                      res match {
-                        case Some((pkg, _)) => pkg.postUninstallNotes
-                        case None => None
-                      }
-                    )
-                  }
-                }
+        Future.collect {
+          apps.map { case (app, op) =>
+            customPackageManagerRouter
```
kvish: would it make more sense to move the call to route to custom manager before construction uninstall operations? since we are creating uninstall operations that will potentially not be used when all we need is the app list returned on line #49 

I propose something like 
` getMarathonApps(req.packageName, req.appId)
      .flatMap { apps =>
        customPackageManagerRouter.getCustomPackageManagerId(
          req.managerId,
          Option(req.packageName),
          apps.head.packageVersion,
          Option(req.appId.getOrElse(apps.head.id))
        ).flatMap {
          case Some((Some(managerId), Some(pkgName), Some(pkgVersion))) if !managerId.isEmpty =>
            logger.debug(s"Request [$req] requires a custom manager: [$managerId]")
            customPackageManagerRouter.callCustomPackageUninstall(
              req, managerId, pkgName, pkgVersion, req.appId.getOrElse(apps.head.id))
          case _ => {
            val uninstallOps = createUninstallOperations(req.packageName, apps)`

kvish: I supposed if we take the CustomEndpointHandler change then this is not relevant?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
```diff
@@ -270,6 +273,12 @@ private[cosmos] final class UninstallHandler(
     }
     uninstallOperations
   }
+
+  override def tryCustomPackageManager(
+    req: rpc.v1.model.UninstallRequest
+  )(
+    implicit session: RequestSession
```
kvish: is this WIP?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceUpdateHandler.scala
```diff
@@ -70,14 +61,36 @@ final class ServiceUpdateHandler(
...
+      Option(request.appId)
+    ).flatMap {
+      case Some((Some(managerId), Some(pkgName), Some(pkgVersion))) if !managerId.isEmpty =>
+        logger.debug(s"Request [$request] requires a custom manager: [$managerId]")
+        customPackageManagerRouter.callCustomServiceUpdate(
+          request,
+          managerId,
+          pkgName,
+          pkgVersion
+        ).map(Some(_))
```
kvish: is the .`map(Some(_))` required here?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
takirala: >  call to route to custom manager before construction uninstall operations

Assume that `getMarathonApps(req.packageName, req.appId)` returns a list of 10 elements and it could be possible that the first few elements need to be processed by cosmos as is and then few by different custom managers and the few more by cosmos as is again. We need to ensure this case is covered. However, it may never happen in reality (as k8 is the only consumer and _i think_ we cannot run multiple k8 clusters of different versions - but if we do one day, we should handle that scenario).

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/ServiceUpdateHandler.scala
takirala: Yes.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
takirala: ideally, as per design doc of Custom manager we should have to look at the headers and payload to decide to route to custom manager. `UninstallHandler` does something that is out of band with the design doc. This is a placeholder to cleanup in future (may be when we deprecate / remove `all` flag ?!)

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
takirala: Also, we should throw the `AmbiguousAppId` error before invoking the CustomPackageManager logic. If we decide not to do this, then the CustomPackageManager SHOULD NOT rely on the appId (instead it should strictly rely only on the incoming payload / headers).

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/PackageInstallHandler.scala
takirala: @kvish i am not sure I understand, what is the change you are referring to ?

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
kvish: hm if we assume that an uninstall operation could be handled by multiple package managers then we should never forward the --all flag, this could cause different managers to collide with each other. As of now our contract with mke is to forward the request as is, so we send the --all flag to mke, so we can assume a single mke will either handle the entire request or cosmos will.

takirala: I agree that `--all` may result in some discrepancies. But at the same time, relying on `apps.head.<somethinghere>` would lead to unpredictable behavior.

Best solution : deprecate `--all` :)

kvish: uninstall must rely on the appId, since we must maintain back-compat with cli the uninstall would be triggered via `dcos package uninstall beta-kubernetes` so we we only have packageName in our payload.

cosmos-server/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandler.scala
kvish: maybe add a TODO marker here as well?

PR #760, opened 2018-11-04
[DCOS-44443] Remove the deprecated DCOS_PACKAGE_METADATA label

takirala:
Address [DCOS-44443](https://jira.mesosphere.com/browse/DCOS-44443)

We have deprecated the DCOS_PACKAGE_METADATA label from 1.11 onwards. DC/OS Frontend was the only component that was using this label. They switched to use `/service/describe` instead of this label (And that was needed for `package-registry` to work anyway). Since this label is deprecated, we can remove the code to encode/decode this label.

This should not require any changes to `cosmos-enterprise` (`CI` passes for fwiw and I have tested manually in an enterprise cluster that few basic flows work as expected).


takirala:
FYI @nLight
This is happening as discussed [in this jira comment](https://jira.mesosphere.com/browse/DCOS-44443?focusedCommentId=187552&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-187552 )

PR #763, opened 2019-01-10
Refactor uninstall to use async/await. Add more logging

alenkacz:


cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
```diff
@@ -23,52 +23,87 @@ final class ServiceUninstaller(
...
-    appId: AppId,
-    frameworkIds: Set[String],
-    deploymentId: String,
-    retries: Int = ServiceUninstaller.DefaultRetries
-  )(
+                 appId: AppId,
+                 frameworkIds: Set[String],
+                 deploymentId: String,
+                 retries: Int = ServiceUninstaller.DefaultRetries
+               )(implicit session: RequestSession): TwitterFuture[Unit] = async {
```
takirala: Using `asyncAwait` is certainly one correct way to do things. But:

1. cosmos is used as a dependency in other projects such as `universe-checker`, `package-registry`, and `cosmos-enterprise`.
2. introducing a new library dependency would increase the (already steep) learning curve for new contributors to `cosmos` (especially because the code is edited relatively rarely)
3. we do not have scala futures anywhere in the codebase and I think it would be nice to retain the same behavior. It would only be confusing if we decide to use scala futures and twitter futures intermittently. Also, I think there are ways to achieve a similar behavior with existing constructs in the code without adding new dependencies. This is also important because of (1) above. More importantly, we have monix's `Task` which is a very similar and an efficient alternative to `asyncAwait`. Because of this, having `asyncAwait` as a dependency _may_ lead to some confusion.

I also feel that we are not simplifying/optimizing the complicated uninstall flow by introducing `async` construct. The existing `for(...)yield(...)` can certainly be refactored to improve the code readability while retaining the same performance. We usually have a [Result](https://github.com/dcos/cosmos/blob/master/cosmos-common/src/main/scala/com/mesosphere/error/package.scala#L6) class which can be used in a `for..yield..` construct. [StorageEnvelope.scala](https://github.com/dcos/cosmos/blob/master/cosmos-common/src/main/scala/com/mesosphere/cosmos/model/StorageEnvelope.scala#L33-L76) is an example. Error classes you have introduced can be used in a `Result[T]` with the `Left` being one of `CosmosError`. I think, this with the combination of `for..yield..` would be a good alternative for usage of `nested ifs` and `async` and `await` (and scala futures).

These are my two cents. Happy to work/discuss on this more.

alenkacz:
@takirala thanks for the feedback. I guess I understand with the async/await part. I would probably fight that in a normal occasion because we probably should lean toward some common stack of libraries. At the same time we did not define that stack as a team and at the same time using async/await did not lead to a neat and easy to follow logic as I was hoping for.

At the same time I don't necessarily agree with the scala future thing. I think that everyone should know scala futures and using them in an isolated module since it's very explicit (none of the usages is ambiguous between 'is this scala or is this twitter future'). I find that part fairly ok...

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
alenkacz: @takirala thanks for the feedback. I guess I understand with the async/await part. I would probably fight that in a normal occasion because we probably should lean toward some common stack of libraries. At the same time we did not define that stack as a team and at the same time using async/await did not lead to a neat and easy to follow logic as I was hoping for.

At the same time I don't necessarily agree with the scala future thing. I think that everyone should know scala futures and using them in an isolated module since it's very explicit (none of the usages is ambiguous between 'is this scala or is this twitter future'). I find that part fairly ok...

alenkacz:
I'll reiterate with more functional solution

PR #764, opened 2019-01-11
Add logging to the uninstall function, fix looking for SDK_UNINSTALL env var

alenkacz:
- More verbose logging to be able to say at which step of uninstall you are or the process failed
- SDK_UNINSTALL env was not being looked at properly
- reduced number of marathon requests (we pulled app two times for unknown reason)
- refactored the overall flow to a pipeline of steps

JIRA issues: DCOS_OSS-4662

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
@@ -1,18 +1,13 @@
 package com.mesosphere.cosmos.service
 
 import com.mesosphere.cosmos.AdminRouter
-import com.mesosphere.cosmos.error.CosmosException
-import com.mesosphere.cosmos.error.MarathonAppNotFound
-import com.mesosphere.cosmos.error.SdkUninstallVarNotFound
+import com.mesosphere.cosmos.error.{CosmosException, MarathonAppNotFound, UninstallFailed}
 import com.mesosphere.cosmos.handler.UninstallHandler
 import com.mesosphere.cosmos.http.RequestSession
-import com.mesosphere.cosmos.thirdparty.marathon.model.AppId
+import com.mesosphere.cosmos.thirdparty.marathon.model.{AppId, MarathonAppResponse}
 import com.twitter.conversions.time._
 import com.twitter.finagle.http.Status
-import com.twitter.util.Try.PredicateDoesNotObtain
-import com.twitter.util.Duration
-import com.twitter.util.Future
-import com.twitter.util.Timer
+import com.twitter.util.{Duration, Future, Timer}
 
 final class ServiceUninstaller(
   adminRouter: AdminRouter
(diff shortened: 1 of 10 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
```diff
@@ -145,4 +164,18 @@ object ServiceUninstaller {
   val RetryInterval: Duration = 10.seconds
 
   private val DefaultRetries: Int = 10000
+
+  sealed trait StepResult
+  case object Success extends StepResult
+  case object Retry extends StepResult
+  case object Finish extends StepResult
```
Fabs: What is the semantic difference of Retry and Finish?

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
```diff
@@ -145,4 +164,18 @@ object ServiceUninstaller {
...
+
+  implicit class RichFuture(val currentStep: Future[StepResult]) extends AnyVal {
```
Fabs: This looks a lot to me like a `Monad Transformer`[1] :). What do you think about naming `next`, `flatMap` :)? If you think about it, your RichFuture is like `Future[Result[T]]` where `Result[T]` is StepResult, and does not carry the value T (is it because of you only interested in the effect, but not the response?). Do you think if Instead of a `3 types` you would be using Either, RichFuture could have been `EitherT`[2] from Cats?


[1] - https://www.youtube.com/watch?v=jd5e71nFEZM 
[2] - https://typelevel.org/cats/datatypes/eithert.html

Fabs: In that cae also, the `stepFunction` could receive what is inside the `currentStep`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
```diff
@@ -32,62 +27,86 @@ final class ServiceUninstaller(
...
-        Future.Done
-      case ex:CosmosException if ex.error.isInstanceOf[SdkUninstallVarNotFound] =>
-        // The app is already being uninstalled / already uninstalled by some other thread.
-        // This is a fresh re-install and should be ignored
-        Future.Done
-      case ex if retries > 0  =>
-        if (!ex.isInstanceOf[PredicateDoesNotObtain]) {
-          logger.info(s"Uninstall attempt for $appId didn't finish. $retries retries left." +
-            s" Type name: ${ex.getClass.getSimpleName}; Message: ${ex.getMessage}")
+    checkDeploymentStep(deploymentId)
```
Fabs: That looks cool. If you see my comment below, if next would be `flatMap` maybe we could even do something like:

```scala
for {
  (appId, frameworks) <- checkDeploymentStep(deploymentId)
  status <- checkSdkUninstallStatusStep(appId, frameworks)
  deleteResult <- deleteSchedulerStep(appId)
} yield ...
```

And since you made all steps `Future[StepResult]` then you do not even need to lift the functions in the context...

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
alenkacz: hmm good point. Success is just that the step succeeded, Finish is that we should mark uninstall as finished. I will try to rename these

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
alenkacz: what do you see as an advantage of naming it flatmap? I was aiming to have a nice readable almost fluent flow of code, that's why I kind of like the `next`

alenkacz: oh I see, e.g. to enable usage inside for comprehension...

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
jeschkies: I'm torn here. I think you are right, @Fabs, but I'm not to big of a fan of the for-comprehension and `onStepSuccess`. The only advantage is that we are not nesting if-else clauses which we would do with an async/await. However, I find it way easier to reason about business logic with of-else clauses. Eg

```scala
retryOn(...) { async {
    if(await(checkSdkUninstallStatusStep(appId, frameworkIds))) {
      await(deleteSchedulerStep(appId))
    }
  }
}
```
the error handling is shifted to `retryOn` and the async block has just the business logic. Also, flat mapping futures has an overhead to co-routines. bBesides, which method are async? All of them? It is not apparent from the `onStepSuccess`.

I don't min d the current solution but find it not easier to reason and debug the the original code.

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
```diff
@@ -32,62 +27,97 @@ final class ServiceUninstaller(
...
     implicit session: RequestSession
-  ): Future[Unit] = {
-    adminRouter.getAppRawJson(appId).map { json =>
-      json.contains(UninstallHandler.envJsonField) &&
-        json(UninstallHandler.envJsonField).toString.toBoolean match {
-        case true => throw SdkUninstallVarNotFound(appId).exception
-        case false => ()
-      }
+  ): Future[StepResult] = {
+    getApp(appId).flatMap {
```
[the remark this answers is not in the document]
alenkacz: we agreed with Tarun that pulling in another library for async handling is not good idea. I had a PR up using async/await but it also requires maping all twitter futures to scala futures so it kind of breaks the consistency of this project.

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
alenkacz: I actually find the current approach pretty neat - when you really go and break your behavior into nice functions, I find the result easier to understand than nested async await with if and elses... Especially if you e.g. have to handle failed futures and use try/catch

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
```diff
@@ -32,62 +27,97 @@ final class ServiceUninstaller(
...
+            case true => StepSuccess
+            case _ => Retry
+          }
+        } else {
+          logger.warn(s"No SDK_UNINSTALL label present on app $appId: $schedulerApp")
+          Future(Retry)
+        }
+      case None =>
+        // The scheduler is already removed from marathon; stop now to avoid uninstalling a restarted app
+        logger.warn("Marking uninstall as done because scheduler does not exist anymore in Marathon - nothing to uninstall.")
```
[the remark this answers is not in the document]
alenkacz: thanks, added more logging to this and other places of this method

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
Fabs: I think that it also makes easier for other devs in the future recognize the pattern, another possible argument is that if `RichFuture` and `Future` have the same api (`flatMap` in this case), then they are also interchangeable.

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
Fabs: I like the current pattern too, maybe I went too far with the `flatMap madness`.

cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
alenkacz: let's talk about it in the office, I cannot exactly picture it as flatmap since RIchFuture is an implicit class so I cannot introduce another FlatMap. But it's true that I can do Either[StepSuccess, StepEnd] where StepEnd would be either failure or UninstallSuccess. I'll follow up in the next PR. Thanks for the help, @Fabs

alenkacz:
I'll follow up in the next PR

PR #765, opened 2019-01-15
Read SDK_UNINSTALL from env

takirala:
- Also refined the definition of `MarathonApp`. Previously `labels` was mandatory but looking at the actual raml spec in marathon repo i realized that it is Optional. Updated the definition to reflect the same, i don't think this is breaking change however (as it is internal only - there are no Encoders for this class)
- Removed some other unused methods.


alenkacz:
Thanks @takirala

PR #769, opened 2019-02-02
Verify resource url with and without url decoding

takirala:


takirala:
I am merging this but @kvish please review & comment as needed :)

AmrAbdelrazik:
thanks @takirala

PR #773, opened 2019-03-18
Remove zip universe support

takirala:
[DCOS_OSS-5013](https://jira.mesosphere.com/browse/DCOS_OSS-5013)

This is for 1.12 and above. We stopped supporting zip universes starting 1.8 and this PR removes the deadcode. This is needed to removed unnecessary metrics (See #771 ). This also bring couple other changes:
 - Always log exceptions in full. Lack of this makes debugging issues like DCOS-46647 difficult.
 - Do not retry uninstall when the auth token is expired.


cosmos-server/src/main/scala/com/mesosphere/cosmos/service/ServiceUninstaller.scala
```diff
@@ -72,6 +71,11 @@ final class ServiceUninstaller(
       }
     }
   }.rescue {
+    case ce: CosmosException if ce.error.isInstanceOf[Unauthorized] =>
```
takirala: Ideally, cosmos can implement authz and refresh the token on expiry. But until then, we can stop retrying when we know retry will fail. See this slack convo https://mesosphere.slack.com/archives/CAA1Z7NKT/p1553816267233400 on how it affected MWT. cc @timcharper

PR #776, opened 2019-03-22
Do not reimplement json merge

alenkacz:


cosmos-common/src/main/scala/com/mesosphere/universe/common/JsonUtil.scala
```diff
@@ -22,17 +22,6 @@ object JsonUtil {
...
-
-      val mergedValue = (targetValueOpt.flatMap(_.asObject), fragmentValue.asObject) match {
-        case (Some(targetObject), Some(fragmentObject)) =>
-          Json.fromJsonObject(merge(targetObject, fragmentObject))
-        case _ => fragmentValue
-      }
-
-      updatedTarget.add(fragmentKey, mergedValue)
-    }
+    Json.fromJsonObject(target).deepMerge(Json.fromJsonObject(fragment)).asObject.get
```
takirala: one of the reasons i don't fully understand circe, wish there was a way to do this without have `.get` but 👍

alenkacz: well ... merge on generic json object does not have to output an object so I had to do get since I know that it must be objects if I am merging two objects 🤷‍♀️

PR #779, opened 2019-05-22
Ignore exceptions when closing stream

alenkacz:
JIRA issues: DCOS-54077

Co-authored-by: Tarun Gupta Akirala <takirala@users.noreply.github.com>

### DIFF
--- a/cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
+++ b/cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
@@ -45,7 +45,10 @@ object HttpClient {
       .flatMap { case (responseData, conn) =>
         Future(processResponse(responseData))
           // IOException when closing stream should not fail the whole request, that's why we're ignoring exceptions with Try
-          .ensure(Try(responseData.contentStream.close()))
+          .ensure({
+          Try(responseData.contentStream.close())
+          ()
+        })
           .ensure(conn.disconnect())
       }
       .handle { case e: IOException =>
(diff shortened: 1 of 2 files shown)


cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
```diff
@@ -43,7 +44,8 @@ object HttpClient {
     fetchResponse(uri, DEFAULT_RETRIES, headers: _*)
       .flatMap { case (responseData, conn) =>
         Future(processResponse(responseData))
-          .ensure(responseData.contentStream.close())
+          // IOException when closing stream should not fail the whole request, that's why we're ignoring exceptions with Try
+          .ensure(Try(responseData.contentStream.close()))
           .ensure(conn.disconnect())
```
takirala: wdyt of wrapping disconnect call also in a Try ?

takirala: And i think the CI is failing cuz `ensure` is expecting a unit value.

alenkacz: I checked and it does not advertise to be throwing any exceptions... they actually even try catch IO exception inside the disconnect call. So I found it not necessary. But if you want to be super-safe, happy to do it

takirala: ah okay. we don't need Try in that case.

cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/ResourceProxyHandlerIntegrationSpec.scala
```diff
@@ -36,6 +37,34 @@ final class ResourceProxyHandlerIntegrationSpec extends FreeSpec
       assertResult(HttpResponseStatus.FORBIDDEN)(response.error.status)
     }
 
+    "fail when the url is present but is in-accessible" in {
+      /**
```
takirala: @alenkacz i don't like this test very much but this is the best compromise i have found so far. lmk if you have any thoughts.

PR #788, opened 2020-06-17
chore: remove some dead code

pierrebeitz:
it looks like Flaggables and all the implicit values are never used.
let's see what CI says.


cosmos-server/src/main/scala/com/mesosphere/cosmos/HttpClient.scala
```diff
@@ -33,10 +33,8 @@ object HttpClient {
 
   val logger: Logger = org.slf4j.LoggerFactory.getLogger(getClass)
 
-  def fetch[A](
-    uri: Uri,
-    headers: (String, String)*
-  )(
+  def fetch[A](uri: Uri, headers: (String, String)*)(
```
pierrebeitz, line 36: this was not intentional. will fixup.

PR #789, opened 2020-06-24
Bump JSON Schema version and lower cache size.

jeschkies:
Summary:
This will lower the memory usage by ignoring the JSON schema cache.
Since we are not using any URIs to load the JSON schema it is cached
each time a new.

JIRA issues: D2IQ-68617

### DIFF
--- a/cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
+++ b/cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
@@ -25,6 +25,9 @@ import java.io.StringWriter
 import java.io.Writer
 import java.nio.charset.StandardCharsets
 import java.util.Base64
+
+import com.github.fge.jsonschema.cfg.ValidationConfiguration
+
 import collection.JavaConverters._
 
 object PackageDefinitionRenderer {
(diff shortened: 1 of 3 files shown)


jeschkies:
This is a run with this patch
benchmark

compare it to the current master
master

timcharper (approved):

cosmos-common/src/main/scala/com/mesosphere/cosmos/render/PackageDefinitionRenderer.scala
```diff
@@ -42,7 +42,8 @@ object PackageDefinitionRenderer {
   }
 
   private[this] implicit val jsf: JsonSchemaFactory = {
-    val validationConfiguration = ValidationConfiguration.newBuilder().setCacheSize(10).freeze()
+    // Disable cache
```
pierrebeitz, line 45: why?

jeschkies: The cache uses the URI as a key. Since we are not providing any URI the cache has no effect.

pierrebeitz (approved):

pierrebeitz:
the python-error is fixed by now. next step is to fix `master`s CI.

janisz (approved):

kaiwalyajoshi (approved):

PR #791, opened 2020-07-14
D2IQ-69788

pierrebeitz:
fix D2IQ-69788. review commit-by-commit!


cosmos-integration-tests/src/main/scala/com/mesosphere/cosmos/handler/UninstallHandlerSpec.scala
```diff
@@ -206,7 +206,7 @@ final class UninstallHandlerSpec extends FreeSpec with Eventually with SpanSugar
   }
 
   def waitUntilDeleted(appId: AppId, clue: Any): Assertion = {
-    eventually(timeout(2.minutes), interval(10.seconds)) {
+    eventually(timeout(5.minutes), interval(10.seconds)) {
```
pierrebeitz, line 209: from what i've seen the framework never even realizes that it shall uninstall things. it's rather not that it does not have enough time to do so but that it just does not seem to care.

